PR #19, opened 2021-02-10
Enable 'bold' with ctrl+b on flutter web

rish07:


lib/widgets/editor.dart
```diff
@@ -73,20 +73,16 @@ abstract class RenderAbstractEditor {
   void selectPosition(SelectionChangedCause cause);
 }
 
-Widget _defaultEmbedBuilder(BuildContext context, Embed node) {
+Widget _defaultEmbedBuilder(BuildContext context, leaf.Embed node) {
   switch (node.value.type) {
-    case 'divider':
```
singerdmx, line 78: We actually removed the support for divider.
We only support image.
So 'divider' can be removed

lib/widgets/editor.dart
```diff
@@ -73,20 +73,16 @@ abstract class RenderAbstractEditor {
...
+    case 'image':
       final style = QuillStyles.getStyles(context, true);
       return Divider(
         height: style.paragraph.style.fontSize * style.paragraph.style.height,
         thickness: 2,
         color: Colors.grey.shade200,
       );
-    case 'image':
```
singerdmx, line 85: Why?
we need this for mobile

lib/widgets/text_line.dart
```diff
@@ -60,7 +60,7 @@ class TextLine extends StatelessWidget {
         textAlign,
         textDirection,
         1.0,
-        Localizations.localeOf(context, nullOk: true),
```
singerdmx, line 63: Why do you want to remove this?

lib/widgets/toolbar.dart
```diff
@@ -446,7 +446,13 @@ Widget _selectHeadingStyleButtonBuilder(
     height: iconSize * 1.77,
     fillColor: Theme.of(context).canvasColor,
     child: Text(
-      _valueToText[value],
```
singerdmx, line 449: Why are we doing this again?

lib/widgets/text_selection.dart
```diff
@@ -199,14 +199,14 @@ class EditorTextSelectionOverlay {
...
-          clipboardStatus,
-        ),
+            context,
+            editingRegion,
+            baseLineHeight,
+            midpoint,
+            endpoints,
+            selectionDelegate,
+            clipboardStatus,
+            Offset(0, 0)),
```
singerdmx, line 209: Why are we doing this again?

lib/widgets/editor.dart
rish07: okay, will remove this

lib/widgets/editor.dart
rish07: For some reason, the network image wasn't loading on web

singerdmx: Yes, it is expected.
check out https://github.com/memspace/zefyr/issues/478

lib/widgets/text_line.dart
rish07: The function accepts only context in flutter master channel. We need to keep developing on flutter master channel because flutter stable channel doesn't have web support. That is why we change the flutter channel by doing` flutter channel master`, `flutter upgrade` and `flutter config --enable-web`

lib/widgets/editor.dart
singerdmx: Actually search zefyr regarding kIsWeb

image

lib/widgets/text_selection.dart
rish07: The same reason as before, flutter master expects the last argument as offset

lib/widgets/text_line.dart
singerdmx: Got it. Can you update readme for that

lib/widgets/toolbar.dart
rish07: Because _valueToText accepts an Attribute.x value but `value` is just x. Therefore I have written cases for those

lib/widgets/text_line.dart
rish07: will do

lib/widgets/text_selection.dart
singerdmx: OK, got it

lib/widgets/toolbar.dart
singerdmx: I don't understand. This change breaks at least mobile. Are you sure?
How come web and mobile differ on this?

rish07: https://user-images.githubusercontent.com/49035175/107504442-4f67e080-6bc1-11eb-9045-c1caffab86ef.mp4

The previous version works well on phone but gives a text null error until I click on the screen. My suggestion would be checking if the platform is web and put if-else condition.

singerdmx: Yes, let's use kIsWeb

lib/widgets/editor.dart
rish07: fixed

lib/widgets/toolbar.dart
rish07: fixed

lib/widgets/constants.dart
```diff
@@ -0,0 +1 @@
+const bool kIsWeb = identical(0, 0.0);
```
singerdmx, line 1: kIsWeb is not provided by flutter?

singerdmx: I mean we don't need to define this, right?

rish07: yes, you are right, removed it.

lib/widgets/editor.dart
```diff
@@ -73,20 +73,20 @@ abstract class RenderAbstractEditor {
...
-        height: style.paragraph.style.fontSize * style.paragraph.style.height,
-        thickness: 2,
-        color: Colors.grey.shade200,
-      );
     case 'image':
-      String imageUrl = node.value.data;
-      return imageUrl.startsWith('http')
-          ? Image.network(imageUrl)
-          : Image.asset(imageUrl);
+      switch (kIsWeb) {
```
singerdmx, line 79: Can we not use switch here? Just if else should suffice

rish07: fixed

PR #24, opened 2021-02-16
Fix image size issue

rish07:
- Although this fixes the sizing issue on the web since HTML is not working on the mobile version, the mobile build fails. Opening this draft PR just for @singerdmx  to check the sizing issue if fixed.

https://user-images.githubusercontent.com/49035175/108018207-11454380-703d-11eb-8b28-0a50e1e55596.mp4

### DIFF
--- a/lib/widgets/responsive_widget.dart
+++ b/lib/widgets/responsive_widget.dart
@@ -1,0 +1,36 @@
+import 'package:flutter/material.dart';
+
+class ResponsiveWidget extends StatelessWidget {
+  final Widget largeScreen;
+  final Widget mediumScreen;
+  final Widget smallScreen;
+
+  const ResponsiveWidget({Key key, @required this.largeScreen, this.mediumScreen, this.smallScreen}) : super(key: key);
+
+  static bool isSmallScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width < 800;
+  }
+
+  static bool isLargeScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width > 1200;
+  }
+
+  static bool isMediumScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width >= 800 && MediaQuery.of(context).size.width <= 1200;
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
+      builder: (context, constraints) {
+        if (constraints.maxWidth > 1200) {
+          return largeScreen;
+        } else if (constraints.maxWidth <= 1200 && constraints.maxWidth >= 800) {
+          return mediumScreen ?? largeScreen;
+        } else {
+          return smallScreen ?? largeScreen;
+        }
+      },
+    );
+  }
+}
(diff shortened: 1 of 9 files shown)


lib/widgets/responsive_widget.dart
```diff
@@ -0,0 +1,36 @@
...
+  static bool isSmallScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width < 800;
+  }
+
+  static bool isLargeScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width > 1200;
+  }
+
+  static bool isMediumScreen(BuildContext context) {
+    return MediaQuery.of(context).size.width >= 800 && MediaQuery.of(context).size.width <= 1200;
```
singerdmx, line 19: Nit: Do we need to use the screen size and use percentage of full size for this logic?

lib/widgets/responsive_widget.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:flutter/material.dart';
+
+class ResponsiveWidget extends StatelessWidget {
+  final Widget largeScreen;
+  final Widget mediumScreen;
+  final Widget smallScreen;
+
+  const ResponsiveWidget({Key key, @required this.largeScreen, this.mediumScreen, this.smallScreen}) : super(key: key);
+
```
singerdmx, line 9: also make 800, 1200 constants as class variables

lib/widgets/responsive_widget.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:flutter/material.dart';
+
+class ResponsiveWidget extends StatelessWidget {
```
singerdmx, line 3: Is ResponsiveWidget for web only?
Can we leverage it for mobile? Or another class with similar logic?

lib/widgets/responsive_widget.dart
rish07: Yes we need to get if the screen that is opened is between small and large

lib/widgets/responsive_widget.dart
rish07: ResponsiveWidget.isSmallScreen is basically phone version

lib/widgets/responsive_widget.dart
rish07: Okay

lib/widgets/responsive_widget.dart
singerdmx:
```
this.embedBuilder =
          kIsWeb ? _defaultEmbedBuilderWeb : _defaultEmbedBuilder}
```

Here if `kIsWeb` is false, you invoke `_defaultEmbedBuilder` which does not use `ResponsiveWidget`

rish07: Yes because on phone the size wasn't an issue, the small screen works when the screen size is small on the web. For example, people using web apps on browsers.

singerdmx: Well, @ bluesky2091 seems to be complaining 

https://github.com/memspace/zefyr/issues/478#issuecomment-778836590

singerdmx: Oh my bad, it is web

PR #31, opened 2021-02-19
Support for non-styling custom attribute

ritheshSalyan:
This pull request contains the following major changes:

1. Change type of SizeAttribute from String to double.
2. support for the non-styling custom attribute.


app/assets/sample_data.json
```diff
@@ -451,7 +451,7 @@
    },
    {
       "attributes": {
-         "size": "small"
```
singerdmx, line 454: Can you keep those but add three inserts?

singerdmx (commented):
I thought you are supposed to use stable branch.
That is stable channel

app/windows/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+
+# Visual Studio build-related files.
+x64/
+x86/
+
+# Visual Studio cache files
+# files ending in .cache can be ignored
+*.[Cc]ache
+# but keep track of directories ending in .cache
+!*.[Cc]ache/
```
singerdmx, line 17: Should .vscode/launch.json  be ignored?

ritheshSalyan: you can ignore that

example/main.dart
```diff
@@ -1,40 +0,0 @@
-import 'dart:async';
-import 'dart:io';
-
```
singerdmx, line 3: This file is for pub. Why are you removing it?

lib/models/documents/attribute.dart
```diff
@@ -201,12 +202,12 @@ class StrikeThroughAttribute extends Attribute<bool> {
   StrikeThroughAttribute() : super('strike', AttributeScope.INLINE, true);
 }
 
-class FontAttribute extends Attribute<String> {
-  FontAttribute(String val) : super('font', AttributeScope.INLINE, val);
+class FontAttribute extends Attribute<double> {
```
singerdmx, line 205: We want to support both string and double, right?

lib/models/rules/format.dart
```diff
@@ -132,3 +132,41 @@ class ResolveInlineFormatRule extends FormatRule {
...
   }
 }
+
+class ResolveInlineIgnoreFormatRule extends FormatRule {
+  const ResolveInlineIgnoreFormatRule();
+
+  @override
+  Delta applyRule(Delta document, int index,
+      {int len, Object data, Attribute attribute}) {
+    if (attribute.scope != AttributeScope.IGNORE) {
```
singerdmx, line 142: Should this be `==` ?

lib/models/rules/rule.dart
```diff
@@ -70,6 +71,6 @@ class Rules {
         throw e;
       }
     }
-    throw ('Apply rules failed');
+    // throw ('Apply rules failed');
```
singerdmx, line 74: Can we do something smarter?
Apply a rule that does nothing for ignore

lib/models/documents/attribute.dart
ritheshSalyan: yes.

singerdmx (commented):
If your goal is support font size of double, you can still assume string at the beginning.
Parse the string to see if it is double and do logic accordingly.
There seems no need to play with ignore or rule

lib/models/documents/attribute.dart
singerdmx: Then don't change it to double.
We still want to accept string like `small`

lib/models/rules/rule.dart
ritheshSalyan: yes, you can. I just copied the same rule as of inline  `ResolveInlineFormatRule()`

lib/models/documents/attribute.dart
singerdmx: > If your goal is support font size of double, you can still assume string at the beginning.
> Parse the string to see if it is double and do logic accordingly.
> There seems no need to play with ignore or rule

image


You just need to handle when `font.value` is `double`, right?

app/assets/sample_data.json
ritheshSalyan: sure. 
check the latest commits.
I reverted type of SizeAttribute to string

now you can set font size in 2 ways 

```dart
   double fontSize = 50.0;
 _controller.formatSelection(SizeAttribute("$fontSize"));

```

Or
```dart

 _controller.formatSelection(SizeAttribute("small"));

```

singerdmx (commented):
Can we have a much smaller change that is only handling font value of double, please?
And probably against stable?
This change is too big

lib/widgets/text_line.dart
```diff
@@ -148,6 +149,11 @@ class TextLine extends StatelessWidget {
           res = res.merge(defaultStyles.sizeHuge);
           break;
         default:
+          double fontSize = double.tryParse(size.value);
```
singerdmx, line 152: I think you only need this file's change, right?

app/assets/sample_data.json
```diff
@@ -473,6 +473,30 @@
       },
       "insert": "Huge"
    },
+   {
```
singerdmx, line 476: Yes, we need this as well

lib/widgets/toolbar.dart
```diff
@@ -742,6 +742,7 @@ class _HistoryButtonState extends State<HistoryButton> {
   }
 
   void _setIconColor() {
+    if(!mounted) return;
```
singerdmx, line 745: OK, we can keep this

singerdmx (commented):
can you shrink down to three files only?
Make a separate pull request if you want change others

ritheshSalyan:
ok

singerdmx:
@ritheshSalyan  are you planning to send a new MR?

PR #41, opened 2021-02-25
Add base64 support in image import

rish07:


lib/widgets/editor.dart
```diff
@@ -1,5 +1,7 @@
+import 'dart:convert';
 import 'dart:io' as io;
 import 'dart:math' as math;
+import 'dart:convert';
```
singerdmx, line 4: This is imported twice

PR #61, opened 2021-03-06
Fix Import issue

rish07:


lib/widgets/editor.dart
```diff
@@ -110,6 +110,7 @@ Widget _defaultEmbedBuilder(BuildContext context, leaf.Embed node) {
 }
 
 Widget _defaultEmbedBuilderWeb(BuildContext context, leaf.Embed node) {
+  var ui = UniversalUI();
```
singerdmx, line 113: Will this be used somewhere?

singerdmx: Also use `final` instead of `var`

singerdmx: OK,  I see it will be used in `ui.platformViewRegistry.registerViewFactory`

singerdmx: can you move it after `case 'image':`

PR #88, opened 2021-03-15
Update quiver hashcode dependency

cbracken:
Development on the quiver_* subpackages has been discontinued in favour
of the main quiver package at https://github.com/google/quiver-dart. The
code itself is identical, but is more actively maintained.

Updated the dependency on the tuple package to 2.0.0, since it also
depends on quiver 3.0.0. Both packages has been migrated to null-safety
at those versions.

### DIFF
--- a/lib/models/documents/attribute.dart
+++ b/lib/models/documents/attribute.dart
@@ -1,4 +1,4 @@
-import 'package:quiver_hashcode/hashcode.dart';
+import 'package:quiver/core.dart';
 
 enum AttributeScope {
   INLINE, // refer to https://quilljs.com/docs/formats/#inline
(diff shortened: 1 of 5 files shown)


cbracken:
Hey there -- I'm one of the two primary authors of quiver/quiver_hashcode. A long time ago, we had the brilliant idea to attempt to break quiver up into a bunch of packages. In the end, the main quiver package ended up remaining far more popular, so we've decided to focus our efforts there, rather than continue to maintain each library in two places. As noted, the code itself is identical, and Dart's tree-shaking compilers result mean all the bits of quiver that you don't u
se are optimised away, so the end result should be identical. Like the subpackage, quiver doesn't have any particularly complex dependencies.

Apologies for the disruption. If you have any questions I'm happy to answer them.

singerdmx:
thanks

Issue #90, opened 2021-03-17
Custom handler when clicking on an image


friebetill:
Is it possible to use a custom handler when clicking on an image in view mode?

I tried to give my image embed a `GestureDetector` and also tried `AbsorbPointer` and `IgnorePointer`, but that didn't help to disable the default handler and enable mine.

mordivgor:
In the worst case you can always try to implement your own embed builder for handling new type of widget.

PR #94, opened 2021-03-19
Add hitTesting to allow changing click behavior in embeds

friebetill:
The code change allows to override the default behaviour of clicking on
images and thus solves #90. 

Now the only thing left to do is to use a `RawGestureDetector` in the 
`EmbedBuilder` to deactivate the "old" behaviour (by entering the 
GestureArena and win). The `embedBuilder` could look like this:
```dart
Widget embedBuilder(BuildContext context, Embed node, bool readOnly) {
  return FutureBuilder<File>(
    future: File('image_path.png'),
    builder: (context, snapshot) {
      final gestures = {
        ReadDependentGestureRecognizer: GestureRecognizerFactoryWithHandlers<
            ReadDependentGestureRecognizer>(
          () => ReadDependentGestureRecognizer(readOnly: readOnly),
          (instance) => instance
            ..onTapUp = ((_) => print('Test'))
        )
      };

      return RawGestureDetector(
        gestures: gestures,
        child: Image.file(snapshot.data),
      );
    }
  );
}

class ReadDependentGestureRecognizer extends TapGestureRecognizer {
  ReadDependentGestureRecognizer({@required this.readOnly});

  final bool readOnly;

  @override
  bool isPointerAllowed(PointerDownEvent event) {
    if (!readOnly) {
      return false;
    }
    return super.isPointerAllowed(event);
  }
}
```

**Explanation of the code:**
When a `PointerDownEvent` is received by the `GestureBinding` a hit
test is performed to determine which `HitTestTarget` nodes are affected.
For this the (I think) element tree is traversed downwards to find
out which elements has been hit. To find out which elements were
hit, the method `hitTestSelf` and `hitTestChildren` are used. Since
the `hitTestChildren` method of `TextLine` has not been implemented 
yet, `false` was always returned, which is why the hit test
never arrived at the `embedBuilder`. With the code change, the hit is
passed on and can be processed correctly in the embed.

Additionally I removed the class `_TransparentTapGestureRecognizer`, 
because this class always accepted the gesture, even if the recognizer 
lost in the arena.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [1.0.9]
+* Web support for raw editor and keyboard listener.
+
 ## [1.0.8]
 * Support token attribute.
 
--- a/app/pubspec.lock
+++ b/app/pubspec.lock
@@ -152,7 +152,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "1.0.8"
+    version: "1.0.9"
   flutter_test:
     dependency: "direct dev"
     description: flutter
--- a/lib/widgets/text_line.dart
+++ b/lib/widgets/text_line.dart
@@ -785,6 +785,11 @@ class RenderEditableTextLine extends RenderEditableBox {
     );
     _cursorPainter.paint(context.canvas, effectiveOffset, position);
   }
+
+  @override
+  bool hitTestChildren(BoxHitTestResult result, {Offset position}) {
+    return this._children.first.hitTest(result, position: position);
+  }
 }
 
 class _TextLineElement extends RenderObjectElement {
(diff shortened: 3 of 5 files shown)


friebetill:
I'll have to take another look at it, since this solution also triggers the `RawGestureDetector` in edit mode.

friebetill:
Is there a reason for `rejectGesture`?
```dart
class _TransparentTapGestureRecognizer extends TapGestureRecognizer {
  _TransparentTapGestureRecognizer({
    Object debugOwner,
  }) : super(debugOwner: debugOwner);

  @override
  void rejectGesture(int pointer) {
    if (state == GestureRecognizerState.ready) {
      acceptGesture(pointer);
    } else {
      super.rejectGesture(pointer);
    }
  }
}
```
This will execute the callback even if the `embedBuilder` won the arena.

singerdmx:
I think the reason is just what you mentioned above

friebetill:
Okay, I tested the editor and could not see any disadvantage in removing the `rejectGesture` method.
Therefore, I suggest that we remove the behaviour. Unless you know the disadvantage when the callbacks are not always executed, I couldn't find any.

singerdmx:
I think you are collaborator now.
So you can go ahead change it.

PR #87, opened 2021-03-15
Sound null safety migration

MillerAdulu:
This PR attempts to resolve #69.

Some of the issues involved were upgrading some depended packages to null-safety and localizing some packages which are no longer maintained. 

One error that stands out is, it's not usable because of a GlobalKey issue in the widget tree once QuillEditor.basic is added to a page

How to try out the package. In `pubspec.yaml`

```yaml
flutter_quill: 
    git:
      url: https://github.com/singerdmx/flutter-quill.git
      ref: e1f3f9aa95dd32f9490222d14a7be3969696f351
```

PR #95, opened 2021-03-19
Fix basic widget rendering and editor usage

MillerAdulu:
With the early merging of #87, there was still an issue with rendering the editor widget. In addition, because of changes in casting lists and iterables in Flutter 2.0, an issue with casting was causing failure to apply styles to text properly while editing. That issue is now fixed in this PR.


singerdmx:
@MillerAdulu can you migrate app folder, please?

image

singerdmx:
@MillerAdulu Is there a way to address this?

image

PR #97, opened 2021-03-20
fix toggle color doesn't work

theacheng:
### 1. Before:

test2

-----
### 2. After:

test1

-----
```dart
void main() {
  bool? isToggled = null;
  print(isToggled == true); //false
}
```

### DIFF
--- a/lib/widgets/toolbar.dart
+++ b/lib/widgets/toolbar.dart
@@ -353,12 +353,12 @@ Widget defaultToggleStyleButtonBuilder(
   final theme = Theme.of(context);
   final isEnabled = onPressed != null;
   final iconColor = isEnabled
-      ? isToggled != null
+      ? isToggled == true
           ? theme.primaryIconTheme.color
           : theme.iconTheme.color
       : theme.disabledColor;
   final fillColor =
-      isToggled != null ? theme.toggleableActiveColor : theme.canvasColor;
+      isToggled == true ? theme.toggleableActiveColor : theme.canvasColor;
   return QuillIconButton(
     highlightElevation: 0,
     hoverElevation: 0,


lib/widgets/toolbar.dart
```diff
@@ -353,12 +353,12 @@ Widget defaultToggleStyleButtonBuilder(
   final theme = Theme.of(context);
   final isEnabled = onPressed != null;
   final iconColor = isEnabled
-      ? isToggled != null
+      ? isToggled == true
```
friebetill, line 356: Nit: You could omit `true`.
```suggestion
      ? isToggled
```

lib/widgets/toolbar.dart
```diff
@@ -353,12 +353,12 @@ Widget defaultToggleStyleButtonBuilder(
...
           ? theme.primaryIconTheme.color
           : theme.iconTheme.color
       : theme.disabledColor;
   final fillColor =
-      isToggled != null ? theme.toggleableActiveColor : theme.canvasColor;
+      isToggled == true ? theme.toggleableActiveColor : theme.canvasColor;
```
friebetill, line 361: Like above
```suggestion
      isToggled ? theme.toggleableActiveColor : theme.canvasColor;
```

lib/widgets/toolbar.dart
theacheng: We can only do this with this nullable variable
```dart
void main() {
  bool? isToggled = null;
  print(isToggled!); //Script error.
}
```

theacheng: void main() {
  bool? isToggled = null;
  print(isToggled); //null
}

friebetill: Thanks I understand, I haven't had a chance to look at non-nullable yet.

friebetill (approved):
LGTM

Issue #39, opened 2021-02-24
Editor view bounces in 0.3.1


JimTim:
In the new 0.3.1 release, unfortunately, the view always jumps up. Probably it scrolls to the bottom every time a letter is entered. This brings a constant scrolling effect.

https://user-images.githubusercontent.com/6737248/108962235-c056d000-7678-11eb-87cf-fa90b94c9ad4.mp4

I think it has to do with the following commit: https://github.com/singerdmx/flutter-quill/commit/471c1cb7ee542977b777b0fc10cbaa5cfb35ddea

In version 0.3.0 the problem does not occur and therefore I went back to this release. So no stress.

singerdmx:
> the view always jumps up

Do you mean the keyboard jumps up?

singerdmx:
OK, let me look into this

singerdmx:
I cannot repro what you have.
Can you try "app" folder code, please?
Also I am unsure why the placeholder "Add Content" does not go away.

singerdmx:
Is there something special about the "Title"? It seems to have a light blue background.
What exactly is its attribute?
Something is different than what we have.

singerdmx:
![image](https://user-images.githubusercontent.com/122956/108963766-67167e80-762f-11eb-9a42-b6dfd30af4bd.png)

singerdmx:
@JimTim submitted https://github.com/singerdmx/flutter-quill/commit/a5e1194bb34c8436454e70e6b474dbc1c9fe23df

PR #111, opened 2021-03-26
Rebuild editor when keyboard is already open

friebetill:
If the keyboard is already open, but the editor thinks that the
keyboard is not open, the text will not be updated when writing.

This can easily happen if one has a `TabBarView` with two children,
each with an `QuillEditor`, see the video and code for an example:

<details><summary>Video</summary>

https://user-images.githubusercontent.com/10923085/112652803-196b7c80-8e4e-11eb-8596-12ed42d71876.mov

</details>

<details><summary>Example</summary>

```dart
import 'package:flutter_quill/widgets/controller.dart';
import 'package:flutter_quill/widgets/editor.dart';

void main() => runApp(MyApp());

class MyApp extends StatefulWidget {
  @override
  _MyAppState createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  late QuillController _controller1;
  late QuillController _controller2;

  @override
  void initState() {
    _controller1 = QuillController.basic();
    _controller2 = QuillController.basic();
    super.initState();
  }

  @override
    return MaterialApp(
      home: DefaultTabController(
        length: 2,
        child: Scaffold(
          appBar: AppBar(
            title: Text('Flutter Quill tabs demo'),
            bottom: TabBar(
              tabs: [
                Tab(text: 'First'),
                Tab(text: 'Second'),
              ],
            ),
          ),
          body: TabBarView(
            children: [
              QuillEditor.basic(controller: _controller1, readOnly: false),
              QuillEditor.basic(controller: _controller2, readOnly: false),
            ],
          ),
        ),
      ),
    );
  }
}
```
</details>

### DIFF
--- a/lib/widgets/editor.dart
+++ b/lib/widgets/editor.dart
@@ -866,6 +866,11 @@ class RenderEditor extends RenderEditableContainerBox
     );
   }
 
+  /// Returns the y-offset of the editor at which [selection] is visible.
+  ///
+  /// The offset is the distance from the top of the editor and is the minimum
+  /// from the current scroll position until [selection] becomes visible.
+  /// Returns null if [selection] is already visible.
   double? getOffsetToRevealCursor(
       double viewportHeight, double scrollOffset, double offsetInViewport) {
     List<TextSelectionPoint> endpoints = getEndpointsForSelection(selection);
--- a/lib/widgets/raw_editor.dart
+++ b/lib/widgets/raw_editor.dart
@@ -886,15 +886,11 @@ class RawEditorState extends EditorState
   }
 
   void _didChangeTextEditingValue() {
-    if (kIsWeb) {
-      _onChangeTextEditingValue();
-      requestKeyboard();
-      return;
-    }
+    _onChangeTextEditingValue();
 
-    if (_keyboardVisible) {
-      _onChangeTextEditingValue();
-    } else {
+    if (kIsWeb) {
+      requestKeyboard();
+    } else if (!_keyboardVisible) {
       requestKeyboard();
     }
   }
--- a/lib/widgets/raw_editor.dart
+++ b/lib/widgets/raw_editor.dart
@@ -702,6 +702,7 @@ class RawEditorState extends EditorState
       _keyboardVisible = true;
     } else {
       _keyboardVisibilityController = KeyboardVisibilityController();
+      _keyboardVisible = _keyboardVisibilityController!.isVisible;
       _keyboardVisibilitySubscription =
           _keyboardVisibilityController?.onChange.listen((bool visible) {
         _keyboardVisible = visible;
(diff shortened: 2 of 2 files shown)


singerdmx:
Not sure this is right

singerdmx:
This may bring back old issue

singerdmx:
See https://github.com/singerdmx/flutter-quill/commit/08ed4c3c85b2bf38cb9631339f81d5ab68edc114

singerdmx:
I'm afraid this change will break single editor behavior.
You are trying to solve multiple editors. Maybe you can pass additional parameter to handle it specially.
DO NOT DO this for multiple editors

singerdmx:
Old issue is https://github.com/singerdmx/flutter-quill/issues/39

Make sure it does not come back

singerdmx:
When keyboard is on, we don't want to call `_onChangeTextEditingValue`
Because it triggers scrollController to scroll

singerdmx:
@friebetill is it possible to associate keyboard with some globalkey? See https://github.com/singerdmx/flutter-quill/pull/113

So you have two keyboards?

friebetill:
I did not have time yesterday, but I will look at the problem again in detail today! Thanks for the reference to the issue #39.

> So you have two keyboards?

I have exactly the problem seen in the video. I have a TabBarView with two editors and switching between the tabs makes the editor not rebuilding when the keyboard stays open.

friebetill:
Okay I have now solved the error by setting the initial visibility of the keyboard depending on whether the keyboard is visible.

Regarding the commit https://github.com/singerdmx/flutter-quill/commit/08ed4c3c85b2bf38cb9631339f81d5ab68edc114:
You solved the editor jumping with this code:
Screenshot 2021-03-27 at 16 46 48
In the code the offset to the point where the selection is visible is calculated. If the selection is not visible, the editor is scrolled until the selection is visible. This prevents jumping, since now no more jumps are made when the selector is visible. 

I don't quite understand the other changes with the `KeyboardListener`, they don't help solve the problem. As far as I can see, we could remove the changes. But maybe I missed something, do you remember why you added the changes?

---
 
**A approach that I have also tried:**
The approach with keys does not work either. But I found out that the state is always destroyed when navigating between tabs. [The problem can be solved](https://stackoverflow.com/questions/49087703/preserving-state-between-tab-view-pages) by implementing `AutomaticKeepAliveClientMixin`. Now the state is preserved, but there is still the problem that the state of the second tab is not created until one navigate there once. That means, there is still the error, that if you open the keyboard in the first tab and then navigate to the second tab, the KeyboardListener of the second tab did not hear the opening of the keyboard. Therefore, unfortunately, this is not a solution.

singerdmx:
KeyboardListener is to detect if keyboard is already open. If true, don't scroll at all

singerdmx:
Actually, it is for the first time you open keyboard, we want to trigger scroll

singerdmx:
The issue is described here https://github.com/memspace/zefyr/issues/485

singerdmx:
basically when keyboard is from invisible to become visible, we want to scroll

PR #117, opened 2021-03-27
Prefer const constructors in immutables

friebetill:


singerdmx:
Need to resolve conflict

PR #157, opened 2021-04-20
Pass getTextPosition instead of passing the tap text position to be more flexible; optional ignoreFocus on replaceText

hyouuu:
So that client can check text position around tapping area, as well as replaceText programmatically without bringing up the keyboard


singerdmx:
LGTM

PR #183, opened 2021-04-25
WIP: remove file picker

singerdmx:


GyuriMajercsik (approved):
Looking good. The less dependency, the better.

SeasonLeee:
does this means the future version of the package won't have that adding image icon or ability? 

Is there a way to have custom button to the tool bar?

Ps, we have so many incompatible issue after adding the package and mostly due to that file_picker......... just after hours research finally fix them... and this pr is going to remove it.... oh... just... why does this not come to you earlier😅

singerdmx:
@SeasonLeee  you can pick up the work left and make a new PR

SeasonLeee:
@singerdmx  Yeah, I would love to, but I am concerned about the fact that, it seems a huge breaking change to the package.

Right now the feature of adding image by tapping the  image_button of tool_bar is right out of the package.

Now, if we are removing the file_picker, how can we be sure this is not going to wreck the existing user customer of the package? 

I mean, I have done a great deal of work on making my project to be compatible with the flutter_quill, which all issues caused by that file_picker(which they have admit on their repo that due to their incomplete migration to AndroidX that the project have a lot of AndroidX incompatible issues..... - -|||)

So... what's your thoughts?

singerdmx:
Do it and bump version to 1.4.0

meitantenik:
Plz do this quick I am having trouble with file picker if you merge this It'll make my life easier

singerdmx:
@meitantenik it is not finished. You are welcome to finish the rest and create a new PR

singerdmx:
Submitted in https://github.com/singerdmx/flutter-quill/commit/2ba2e7e309e27c51dcaca0d37acd90ecd0117b9b

Issue #158, opened 2021-04-20
Checkbox not work on tapping


theacheng:
I need to click at near text to make checkbox work.

![test1](https://user-images.githubusercontent.com/29684683/115368504-57f81b00-a1f1-11eb-8598-f732be1bd663.gif)

This checkbox update on this method: `bool _onTapping(TapUpDetails details) {...}`
But when I click on checkbox, it call `void onTapDown(TapDownDetails details)` instead. 

Desire result is to update checkbox on tap up.

singerdmx:
@theacheng the problem is how to figure out when the checkbox is clicked - we compromised to use beginning of the line since we were unable to find a way to know when checkbox is hit

kevinDespoulains:
Hi, 

I made this PR : https://github.com/singerdmx/flutter-quill/pull/186, could you please check if it fits your needs ?

theacheng:
That works well for me. Thank you!

![test1](https://user-images.githubusercontent.com/29684683/116063732-82dde580-a6af-11eb-836c-7ef165543bc6.gif)

-----

But it would be better if we are able to hold it and update on tapping up. (I'm not sure if this is good behavior). 
This is when we holding it:

![test1](https://user-images.githubusercontent.com/29684683/116064344-20d1b000-a6b0-11eb-8c5c-5566bc2647e4.gif)

kevinDespoulains:
I added the behavior you asked by wrapping the checkboxes in a GestureDetector.

I also added the usage of a key on the checkboxes widgets because clicking on the first checkbox with some other checkboxes selected under it had a bad side effect (An unwanted animation on the checkboxes was triggered).

Let me know if it is working as you want ;)

theacheng:
It looks good without that animation for the first checkbox (that's fair for other checkboxes :D)
Thank you again!

![test1](https://user-images.githubusercontent.com/29684683/116070395-c4be5a00-a6b6-11eb-9888-fd361fa194dc.gif)

singerdmx:
@kevinDespoulains thanks so much

PR #186, opened 2021-04-26
Updating checkbox to handle tap

kevinDespoulains:
Hi,

Linked to issue [#158](https://github.com/singerdmx/flutter-quill/issues/158), I was looking for a way to make checkbox clickable.

What to you think about this?

### DIFF
--- a/lib/widgets/text_block.dart
+++ b/lib/widgets/text_block.dart
@@ -698,9 +698,7 @@ class _BulletPoint extends StatelessWidget {
   }
 }
 
-/// A Checkbox widget.
 class _Checkbox extends StatelessWidget {
-  /// Creates a [_Checkbox].
   const _Checkbox({
     Key? key,
     this.style,
--- a/lib/widgets/text_block.dart
+++ b/lib/widgets/text_block.dart
@@ -163,6 +163,7 @@ class EditableTextBlock extends StatelessWidget {
 
     if (attrs[Attribute.list.key] == Attribute.checked) {
       return _Checkbox(
+        key: UniqueKey(),
         style: defaultStyles!.leading!.style,
         width: 32,
         isChecked: true,
@@ -173,9 +174,9 @@ class EditableTextBlock extends StatelessWidget {
 
     if (attrs[Attribute.list.key] == Attribute.unchecked) {
       return _Checkbox(
+        key: UniqueKey(),
         style: defaultStyles!.leading!.style,
         width: 32,
-        isChecked: false,
         offset: block.offset + line.offset,
         onTap: onCheckboxTap,
       );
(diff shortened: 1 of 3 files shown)


lib/widgets/text_block.dart
```diff
@@ -161,12 +163,23 @@ class EditableTextBlock extends StatelessWidget {
 
     if (attrs[Attribute.list.key] == Attribute.checked) {
       return _Checkbox(
-          style: defaultStyles!.leading!.style, width: 32, isChecked: true);
+        key: UniqueKey(),
```
matuella, line 166: Any particular reason for adding an `UniqueKey` every time this widget builds?

lib/widgets/text_block.dart
```diff
@@ -685,46 +698,41 @@ class _BulletPoint extends StatelessWidget {
   }
 }
 
-class _Checkbox extends StatefulWidget {
-  const _Checkbox({Key? key, this.style, this.width, this.isChecked})
-      : super(key: key);
-
+/// A Checkbox widget.
+class _Checkbox extends StatelessWidget {
+  /// Creates a [_Checkbox].
```
matuella, line 703: I don't think these docs add any relevant info to the class, personally I would remove them or get a little more in-depth of what they actually do.

matuella (commented):
That looks like a good solution for me, I've tested it and apparently works fine. Ideally, we wouldn't need a gesture detector to solve this problem, as the checkbox itself knows how to handle this, but I'm not sure the root of this problem (maybe conflicting gestures?), so I would say this is overall good. I can't approve this because you've removed some code that I am not sure the impact of it (the `_flipListCheckbox` one).

@singerdmx, do you know if the removed code makes sense?

lib/widgets/text_block.dart
kevinDespoulains: Yes, without using a key you can observe an unwanted animation on checkboxes when first chekbox is tapped.

Tested on version 1.2.1 :
Animation

lib/widgets/text_block.dart
kevinDespoulains: I removed these lines in the latest commit.

kevinDespoulains:
- The GestureDetector is used to handle long press, as suggested in issue https://github.com/singerdmx/flutter-quill/issues/158. I could not find a way to handle this gesture directly from checkbox.
- For the _flipListCheckbox part, I removed it to avoid checking/unchecking a checkbox on a click next to it (like it was before). Let me know if you want me to add it back.

lib/widgets/text_block.dart
GyuriMajercsik: A `Key(index)` would make more sense. Basically, each item should have its own key.

matuella: I agree that a `Key` is necessary here, as they are all of the same class type, but just like @GyuriMajercsik suggested, a key that doesn't change every time this is built would be ideal. While IMO [index is not ideal](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318), it's probably decent for this particular case here.

matuella:
> * For the _flipListCheckbox part, I removed it to avoid checking/unchecking a checkbox on a click next to it (like it was before). Let me know if you want me to add it back.

Apparently that's fine for me, just wanted to make sure that @singerdmx also thinks this as well.

PR #187, opened 2021-04-26
Simple viewer

em6m6e:
Add a QuillSimpleViewer widget with 2 modes:
-if the parameter truncate if false, just a read-only widget, without any user interaction (no SingleChildScrollView, GestureDetector), outbound Container with constraints or Global Key used
-if the parameter truncate is true,  truncate the document (truncateWidth x truncateHeight size) and rescale it (truncateScale parameter). It's useful for preview a bunch of documents as thumbnails.


singerdmx:
image

Issue #229, opened 2021-05-20
[Mobile] Can I adjust the toolbar to multi-line display?


zhaoce:
I don’t want to scroll the toolbar in mobile. 
Maybe it’s more intuitive to display with multiple lines. 
Can I control how to display it?

like this：
<img width="489" alt="image" src="https://user-images.githubusercontent.com/16445113/118970017-595b6580-b9a0-11eb-9d63-30df577bab35.png">

zhaoce:
Add this feature in pr: #229

PR #230, opened 2021-05-20
toolbar multi-line display #229

zhaoce:
Add a parameter to Toolbar, it supports multi-line display


zhaoce:
This modification will not have any impact on other users

PR #236, opened 2021-05-22
Fix paste

### LINKED ISSUE #235: [Mobile]  paste not work
paste does not work nothing appears after clicking paste  and when running the cut does not disappear

friebetill:
closes #235.

### DIFF
--- a/lib/src/widgets/raw_editor.dart
+++ b/lib/src/widgets/raw_editor.dart
@@ -680,7 +680,7 @@ class RawEditorState extends EditorState
   @override
   void userUpdateTextEditingValue(
       TextEditingValue value, SelectionChangedCause cause) {
-    // TODO: implement userUpdateTextEditingValue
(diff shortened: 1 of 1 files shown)


hyouuu:
@friebetill  Thanks for adding paste support! However I found that after I paste stuff then tap enter or any input, the pasted content would be deleted (sometimes even affect other text surrounding the pasting location) - any thoughts on the cause?

Found some relevant stuff for zefyr - probably needs to do something like replaceText to make the controller handle the text change properly https://github.com/memspace/zefyr/issues/420

friebetill:
Thanks @hyouuu I can reproduce this. I don't have time to look at it right now, so I'm happy if someone else can look at it. Otherwise I'll do it as soon as I have time, probably next week.

hyouuu:
Thanks @friebetill ! I'm occupied at the moment too so will have to wait for your fix :p

PR #271, opened 2021-06-23
Replace deprecated textEditingValue method

singerdmx:


friebetill (changes requested):
As far as I can see, `userUpdateTextEditingValue` is called inside the `_handleSelectionHandleChanged` -> `onSelectionHandleChanged` -> `_handleDragUpdate` method. Therefore I think the `SelectionChangedCause` should be `drag` instead of `keyboard`. WDYT?

singerdmx:
> As far as I can see, `userUpdateTextEditingValue` is called inside the `_handleSelectionHandleChanged` -> `onSelectionHandleChanged` -> `_handleDragUpdate` method. Therefore I think the `SelectionChangedCause` should be `drag` instead of `keyboard`. WDYT?

You are correct. Thx!

PR #273, opened 2021-06-24
Support Multi Row Toolbar

florianh01:
Support of multi Line Toolbar

This PR is no breaking change


singerdmx:
Let's make this default

singerdmx:
So make it true

florianh01:
Making this default will have breaking changes for others

singerdmx:
I will make it version 1.6.0

PR #306, opened 2021-07-22
Native IOS keyboard input Chinese characters: bold ,underline ,color, italics will be invalid . how can i do it ?

quinoxz:
Native IOS keyboard input Chinese characters: bold ,underline ,color, italics will be invalid . how can i do it ?


quinoxz:
1

PR #316, opened 2021-08-03
Add Link as source for images and videos

rho-cassiopeiae:
Hosted images and videos are supported, but currently there's no way to actually insert them.
flutter_01

### DIFF
--- a/lib/src/utils/media_source.dart
+++ b/lib/src/utils/media_source.dart
@@ -1,0 +1,5 @@
+enum MediaSource {
+  Gallery,
+  Link,
+  GalleryAndLink,
+}
--- a/lib/src/utils/media_source.dart
+++ b/lib/src/utils/media_source.dart
@@ -1,5 +1,0 @@
-enum MediaSource {
-  Gallery,
-  Link,
-  GalleryAndLink,
-}
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -3,7 +3,7 @@ import 'dart:io';
 import 'package:flutter/material.dart';
 
 import '../models/documents/attribute.dart';
-import '../utils/media_source.dart';
+import '../utils/media_pick_setting.dart';
 import 'controller.dart';
 import 'toolbar/arrow_indicated_button_list.dart';
 import 'toolbar/camera_button.dart';
@@ -19,6 +19,7 @@ import 'toolbar/toggle_check_list_button.dart';
 import 'toolbar/toggle_style_button.dart';
 import 'toolbar/video_button.dart';
 
+export '../utils/media_pick_setting.dart';
 export 'toolbar/clear_format_button.dart';
 export 'toolbar/color_button.dart';
 export 'toolbar/history_button.dart';
@@ -31,6 +32,7 @@ export 'toolbar/quill_icon_button.dart';
 export 'toolbar/select_header_style_button.dart';
 export 'toolbar/toggle_check_list_button.dart';
 export 'toolbar/toggle_style_button.dart';
+export 'toolbar/video_button.dart';
 
 typedef OnImagePickCallback = Future<String?> Function(File file);
 typedef OnVideoPickCallback = Future<String?> Function(File file);
@@ -39,7 +41,7 @@ typedef WebImagePickImpl = Future<String?> Function(
     OnImagePickCallback onImagePickCallback);
 typedef WebVideoPickImpl = Future<String?> Function(
     OnVideoPickCallback onImagePickCallback);
-typedef MediaSourceSelectorBuilder = Future<MediaSource?> Function(
+typedef MediaPickSettingSelector = Future<MediaPickSetting?> Function(
     BuildContext context);
 
 // The default size of the icon of a button.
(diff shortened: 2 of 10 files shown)


lib/src/widgets/toolbar.dart
```diff
@@ -39,6 +39,8 @@ typedef WebImagePickImpl = Future<String?> Function(
     OnImagePickCallback onImagePickCallback);
```
singerdmx, line 39: Are you supposed to remove OnImagePickCallback ?

lib/src/widgets/toolbar.dart
```diff
@@ -78,8 +80,11 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
     bool showHorizontalRule = false,
     bool multiRowsDisplay = true,
     bool showCamera = true,
+    MediaSource? imageSource,
     OnImagePickCallback? onImagePickCallback,
```
singerdmx, line 84: Are you supposed to remove this?

lib/src/utils/media_source.dart
```diff
@@ -0,0 +1,5 @@
+enum MediaSource {
```
singerdmx, line 1: Is this some kind of configuration setting?
We need to export it so that our client can choose

lib/src/widgets/toolbar.dart
```diff
@@ -78,8 +80,11 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
     bool showHorizontalRule = false,
     bool multiRowsDisplay = true,
     bool showCamera = true,
+    MediaSource? imageSource,
```
singerdmx, line 83: Need to update `example` folder to add MediaSource

lib/src/utils/media_source.dart
singerdmx: Better rename it to something like `MediaSourceSetting` or something better

rho-cassiopeiae: yeah. If you specify `MediaSource.Link`, then `OnImagePickCallback` is never used (you can still specify it for use by `CameraButton`). If you specify `MediaSource.Gallery` or `MediaSource.GalleryAndLink`, then `OnImagePickCallback` must be provided as usual. Same goes for videos.

It's exported via flutter_quill.dart. Should I export it separately?

singerdmx: Could you update https://github.com/singerdmx/flutter-quill/blob/master/example/lib/pages/home_page.dart#L143 and related code that initializes toolbar?

rho-cassiopeiae: sure, will do it tomorrow.

singerdmx: Naming it `MediaSource` is confusing.
Try a better naming. thanks!

singerdmx: actually let's get rid of the configuring.
Make is default that link is alway possible. How about that?

singerdmx: That means user can always insert image/video using link. If `OnImagePickCallback` is provided, user can have additional choice to insert from gallery.
So you can get rid of `MediaSource`

singerdmx: I don't anyone would not want having "Link" as MediaSource.
It is always nice to have it.
Also this simplifies our code base.

singerdmx: If `OnImagePickCallback` is not provided, we can still hide the button.
But if `OnImagePickCallback` is provided, we show both gallery and link.

rho-cassiopeiae: For my case I simply don't want to allow users to upload videos from device, so your offer is fine by me.

singerdmx:
```
 if (onImagePickCallback != null)
          ImageButton(
            icon: Icons.image,
            iconSize: toolbarIconSize,
            controller: controller,
            onImagePickCallback: onImagePickCallback,
            filePickImpl: filePickImpl,
            webImagePickImpl: webImagePickImpl,
          ),
```

Here we still use  `if (onImagePickCallback != null)` and no need to update `example` folder

singerdmx: > For my case I simply don't want to allow users to upload videos from device, so your offer is fine by me.

I see, so we need to too bool to tune wether to show buttons

image

singerdmx: For your case, you still want to show the buttons while `onImagePickCallback` and `onVideoPickCallback` are null.
Maybe create two bool variables named `showVideoButton` and `snowImageButton`. Rename `showCamera` to `showCameraButton`.

singerdmx: But I would imagine having `link` is always available in addition to uploading from device

singerdmx: > sure, will do it tomorrow.

@rho-cassiopeiae Please let me know if you need any help.
I would love to see this checked in.

rho-cassiopeiae:
Made an update.
But I still want to be able to provide a custom "pick from" dialog, so there's got to be some config setting. Using bool would be ambiguous.

PR #332, opened 2021-08-10
add custom emotions button

lanwufeng:
Hello, I added a custom emotions button, next, how should I pass in a picture with custom attributes?
thanks a lot


singerdmx:
This code is totally unacceptable

Issue #338, opened 2021-08-12
Custom Text Selection Toolbar


jonasbernardo:
allow to customize the Text Selection Toolbar with widgets that receive the selected text and do some customize function to execute with the text selection.

singerdmx:
You can submit a PR

PR #339, opened 2021-08-12
Custom Text Selection Toolbar #338

jonasbernardo:
allow to customize the Text Selection Toolbar with widgets that receive the selected text and do some customize function to execute with the text selection.


singerdmx:
Fix it!

PR #351, opened 2021-08-19
Support for building custom inline styles

singerdmx:


ritheshSalyan:
@singerdmx yes. Those are the changes.

PR #372, opened 2021-08-31
Option added to controller to keep inline style on new lines

kevinDespoulains:
Hi !

I need to keep inline styles (bold, italic, underline, etc) even when I add a new line. It is annoying for the user to activate it multiple times for each line.
I think adding the optional parameter `keepStyleOnNewLine` with a default value of false can allow this behavior. When this parameter is set to true, inline attributes won't be removed from style.

Don't hesitate to tell me if this functionnality is breaking current code.

Also, maybe the name keepStyleOnNewLine isn't perfect, if you have a better idea, I'll take it !

Thank you ! :)

### DIFF
--- a/lib/src/widgets/controller.dart
+++ b/lib/src/widgets/controller.dart
@@ -14,7 +14,9 @@ class QuillController extends ChangeNotifier {
   QuillController({
     required this.document,
     required TextSelection selection,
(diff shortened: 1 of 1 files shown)


redvelvet-fan:
is not working versio ^6.3.2

PR #421, opened 2021-10-17
Adding translations to the toolbar!

Namli1:
I used the [i18n_extension](https://pub.dev/packages/i18n_extension) package (because of simplicity and clarity) to add translations to the toolbar. There are only a few places with text, namely:

- LinkDialog(): The description text in the dialog
- ColorButton(): The title text of the dialog
- ImageVideoUtils(): I found two words there and added them to the translations

In order to use the translations, I had to wrap the QuillToolbar in the I18n() widget, but that shouldn't change anything.
The translations can be found at [lib/src/translations/toolbar.i18n.dart](flutter-quill/lib/src/translations/toolbar.i18n.dart), where anyone could add their own translations as well.

I think that translating the editor will make it attractive to more developers, so I hope this feature  gets reviewed soon. Thanks for building the best flutter text editor out there!

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,20 @@ Define `mobileWidth`, `mobileHeight`, `mobileMargin`, `mobileAlignment` as follo
 }
 ```
 
+## Translation of toolbar
+The package offers translations for the quill toolbar, it will follow the system locale unless you set your own locale with:
+```
+QuillToolbar(locale: Locale('fr'), ...)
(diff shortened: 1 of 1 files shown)


singerdmx:
@Namli1 please check 200fbaa

Namli1:
Commit looks good, great to see that there are so many translations!

PR #433, opened 2021-10-28
Arabic Locale

vaygeth89:
Added Arabic language locale "ar", into the localization string


singerdmx:
@vaygeth89 please check 200fbaa

PR #437, opened 2021-11-02
implement change cursor to SystemMouseCursors.click when hovering a link (MacOS)

khanhnguyen-96:
This pull request makes cursor change to SystemMouseCursors.click when hovering a link (MacOS)

### DIFF
--- a/lib/src/widgets/text_line.dart
+++ b/lib/src/widgets/text_line.dart
@@ -184,6 +184,7 @@ class TextLine extends StatelessWidget {
     final style = textNode.style;
     var res = const TextStyle(); // This is inline text style
     final color = textNode.style.attributes[Attribute.color.key];
+    var hasLink = false;
 
     <String, TextStyle?>{
       Attribute.bold.key: defaultStyles.bold,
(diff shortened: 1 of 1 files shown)


singerdmx:
Do you need to check it is macOS system?

khanhnguyen-96:
I think it will make more sense if I remove the "MacOS" part. Because this should support cursor from any OS. What do you think?

singerdmx:
> I think it will make more sense if I remove the "MacOS" part. Because this should support cursor from any OS. What do you think?

What about mobile - iOS and Android?

singerdmx:
You can check out how to detect OS:

https://github.com/singerdmx/flutter-quill/blob/master/example/lib/pages/home_page.dart#L196

singerdmx:
NVM, maybe mobile is fine

khanhnguyen-96:
> NVM, maybe mobile is fine

Yeah, I totally agree since this is based on Flutter's behavior itself. Not many users will use a pointer on Android and iOS tho.

PR #461, opened 2021-11-15
Add a new parameter toolBarSectionSpacing to control the toolbar icon Wrap spacing

mark8044:
The toolbar icons are in a Wrap, with the sections separated by the `spacing` parameter.   This opens the parameter up for us to modify via `toolBarSectionSpacing`

If possible please review my code, I might have made some noob mistakes in how its written, but it does work well


singerdmx (changes requested):
Please remove extra blank line and space

singerdmx (changes requested):
Please fix line 436

mark8044:
Ok thanks I tried to fix that line I hope I got it correct

PR #462, opened 2021-11-15
Allow alignment of the toolbar icons to match WrapAlignment

mark8044:
Currently, the toolbar is inside a `Wrap` which is set to `WrapAlignment.center`, this opens up `WrapAlignment.start` and `WrapAlignment.end` by creating a new enum: `ToolbarAlignment` via `toolBarIconAlignment`

We now have `ToolbarAlignment.start`, `ToolbarAlignment.center` and `ToolbarAlignment.end` with `ToolbarAlignment.center` as the default so it looks the same as it was before

```
QuillToolbar.basic(
...
toolBarIconAlignment:ToolbarAlignment.start,
)
```

Let me know if I hosed anything up!

### DIFF
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -439,6 +439,7 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
 
   @override
   Size get preferredSize => Size.fromHeight(toolBarHeight);
+  Map<ToolbarAlignment, WrapAlignment> alignmentMap = {ToolbarAlignment.center: WrapAlignment.center}
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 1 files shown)


singerdmx (changes requested):
Please fix line 442 and format code

mark8044:
> Please fix line 442 and format code

I checked that line and it looked like a blank line separating

  Size get preferredSize => Size.fromHeight(toolBarHeight);
 from the next line?

Thanks for accepting this PR, this helps me a lot!

singerdmx:
image

singerdmx:
image

mark8044:
Ok thanks I'll try that next.

I have been doing all these edits directly on GitHub-web.  I think my workflow sucks.  When I try to change these files directly in Android-Studio they never take effect.  I have to change everything on Github then download the package and test.. over and over :)

PR #463, opened 2021-11-15
Added more enum to ToolbarAlignment to fully match WrapAlignment

mark8044:
Last PR only had `start`, `end` and `center` added.  This now adds all the `WrapAlignment` elements to fully match, incase someone needs the other ones which include: `spaceAround`, `spaceBetween`, `spaceEvenly`

### DIFF
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -49,7 +49,7 @@ typedef WebVideoPickImpl = Future<String?> Function(
 typedef MediaPickSettingSelector = Future<MediaPickSetting?> Function(
     BuildContext context);
 
-enum ToolbarAlignment { start, center, end }
+enum ToolbarAlignment { start, center, end, spaceAround, spaceBetween, spaceEvenly }
 
 // The default size of the icon of a button.
 const double kDefaultIconSize = 18;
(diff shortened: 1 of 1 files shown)


lib/src/widgets/toolbar.dart
```diff
@@ -452,7 +452,13 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
                       ? WrapAlignment.center
                       : (toolBarIconAlignment == ToolbarAlignment.end)
                           ? WrapAlignment.end
-                          : WrapAlignment.center,
+                          : (toolBarIconAlignment == ToolbarAlignment.spaceAround)
```
singerdmx, line 455: Use switch for this.
Actually please get rid of `enum ToolbarAlignment`

singerdmx: There is NO need to have such enum. Please remove it

mark8044: What is a better way? Just directly use the WrapAlignment?

```
QuillToolbar.basic(
...
toolBarIconAlignment:WrapAlignment.start,
)
```

Like that?

singerdmx: yes

PR #476, opened 2021-11-22
add turkish language

malikdoksoz:
support turkish language


singerdmx:
@malikdoksoz please check 200fbaa

PR #499, opened 2021-12-06
Add Danish translations to toolbar.i18n.dart

mit-mit:


singerdmx:
@mit-mit please check 200fbaa

PR #526, opened 2021-12-14
Option to disable floating cursor

MagTuxGit:
Moving cursor using keyboard on iOS is great, but the floating cursor works awful. 
I tried to fix it, spent some time debugging, but unfortunately I have no more time at the moment.

So I decided just to disable the floating cursor. Cursor moving works without a floating cursor.
I added new argument to QuillEditor.

### DIFF
--- a/lib/src/widgets/editor.dart
+++ b/lib/src/widgets/editor.dart
@@ -242,6 +242,7 @@ class QuillEditor extends StatefulWidget {
       this.onSingleLongTapEnd,
       this.embedBuilder = defaultEmbedBuilder,
       this.customStyleBuilder,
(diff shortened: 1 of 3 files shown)


adrian-moisa:
@MagTuxGit  Quick question, what is a floating cursor?

singerdmx:
https://www.youtube.com/watch?v=JmFvtfvhZ9k


Please google it out before asking

adrian-moisa:
Had I not been confused by the results I wouldn't have asked. It seems many pages indicate that this means cursor drift. The video you sent seems to indicate a completely different question. I've been long enough on the internet to know not to ask before a google search. Thanks for the support.

PR #579, opened 2022-01-06
feat: support multiple links insertion on the go

### LINKED ISSUE #578: feat: support insertion of multiple links
## Description

At the moment (`flutter_quill: 3.1.1`), there are some limitations or bugs when inserting links:

- The link attribute is only applied after a single space char insertion.
- When pasting text that includes links, the proper link attribute is not applied.
- The surrounding text is not always reformatted according to the resulting `Document` i.e. if an existing link gets broken, its link attribute should be removed, but it is not the case.
- Link insertion only uses the `Uri.parse` method and checks the schemas to check if it is valid, which is not that realiable.

I am about to submit a PR on this but would like you to check it and verify the link insertion rule behaviour.

mrverdant13:
## Details

- Apply link attribute as the user types.
- Support link attributes assignation when pasting a block of text.
- Apply or remove the link attribute to the words around the insertion, considering the resulting affected text segments.
- Use regular expressions to check if a URL is acceptable or valid. The pattern in use is `r'(https?:\/\/|www\.)[\w-\.]+\.[\w-\.]+(\/([\S]+)?)?'` but a better one might replace it.

## Manual testing

1. Run the `example` demo app.
2. Insert the following text:
	```
	Lorem ipsum dolor sit amet, consectetur www.google.com adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation http://google.com ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse https://www.google.com cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
	```
The included URLs should be formatted with the link attribute.

## Related issues

- Might close #578

### DIFF
--- a/lib/src/models/rules/insert.dart
+++ b/lib/src/models/rules/insert.dart
@@ -1,6 +1,6 @@
-import 'package:flutter_quill/flutter_quill.dart';
 import 'package:tuple/tuple.dart';
 
+import '../../../models/documents/document.dart';
 import '../documents/attribute.dart';
 import '../documents/style.dart';
 import '../quill_delta.dart';
@@ -291,8 +291,8 @@ class InsertEmbedsRule extends InsertRule {
 /// the URL pattern.
 ///
 /// The link attribute is applied as the user types.
-class AutoFormatLinksRule extends InsertRule {
-  const AutoFormatLinksRule();
+class AutoFormatMultipleLinksRule extends InsertRule {
+  const AutoFormatMultipleLinksRule();
 
   /// Link pattern.
   ///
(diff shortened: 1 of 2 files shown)


lib/src/models/rules/insert.dart
```diff
@@ -286,44 +287,107 @@ class InsertEmbedsRule extends InsertRule {
   }
 }
 
-/// Applies link format to text segment (which looks like a link) when user
-/// inserts space character after it.
+/// Applies link format to text segments within the inserted text that matches
+/// the URL pattern.
+///
+/// The link attribute is applied as the user types.
 class AutoFormatLinksRule extends InsertRule {
```
singerdmx, line 294: Can we add a new class for this?

mrverdant13: Sure thing!

lib/src/models/rules/insert.dart
```diff
@@ -286,44 +287,107 @@ class InsertEmbedsRule extends InsertRule {
...
   const AutoFormatLinksRule();
 
+  /// Link pattern.
+  ///
+  /// This pattern is used to match a links within a text segment.
+  static const _linkPattern =
```
singerdmx, line 300: Why do you need variable `_linkPattern` ?

singerdmx: You can just inline it

lib/src/models/rules/insert.dart
```diff
@@ -286,44 +287,107 @@ class InsertEmbedsRule extends InsertRule {
...
+      r'(https?:\/\/|www\.)[\w-\.]+\.[\w-\.]+(\/([\S]+)?)?';
```
singerdmx, line 301: I am concerned about this regex.
Could you add comments giving a few sample strings that this regex matches?

singerdmx: So you are ignoring `http`?

lib/src/models/rules/insert.dart
mrverdant13: I was thinking of using it later when inserting a link with the `LinkStyleButton` dialog. But an inline string would work for now.

lib/src/models/rules/insert.dart
mrverdant13: It works for the following testing URLs:
- www.google.com
- http://google.com
- https://www.google.com
- http://beginner.example.edu/#act
- https://birth.example.net/beds/ants.php#bait
- http://example.com/babies
- https://www.example.com/
- https://attack.example.edu/?acoustics=blade&bed=bed
- http://basketball.example.com/
- https://birthday.example.com/birthday
- http://www.example.com/
- https://example.com/addition/action
- http://example.com/
- https://bite.example.net/#adjustment
- http://www.example.net/badge.php?bedroom=anger
- https://brass.example.com/?anger=branch&actor=amusement#adjustment
- http://www.example.com/?action=birds&brass=apparatus
- https://example.net/

I used [this URL generator tool](https://www.randomlists.com/urls), btw.

singerdmx: looks good

singerdmx:
is it still draft?

lib/src/models/rules/insert.dart
mrverdant13: Done!

PR #603, opened 2022-01-23
Replace Toolbar showHistory group with individual showRedo and showUndo

mark8044:
Replace Toolbar showHistory group with individual showRedo and showUndo to give more granular control of icon display


mark8044:
FYI this is breaking change as `showHistory` was removed

Issue #304, opened 2021-07-21
[Web] Keyboard arrow keys does not work inside a ScrollView


wliumelb:
When the QuillEditor is a child of SingleChildScrollView, cannot move the cursor with keyboard arrow keys.

singerdmx:
Don't make it a child of SingleChildScrollView
Not our problem

wliumelb:
Found a workaround: 

Wrap QuillEditor with a RawKeyboardListener to intercept KeyEvent and prevent it from propagating to parents.

````
SingleChildScrollView(
        child: RawKeyboardListener(
              onKey: (RawKeyEvent event) {},
              focusNode: FocusNode(onKey: (FocusNode node, RawKeyEvent event) {
                  return KeyEventResult.skipRemainingHandlers;
              }),
              child: QuillEditor(
                    ...
              ),
       ),
)
````

jagrut-18:
> Found a workaround:
> 
> Wrap QuillEditor with a RawKeyboardListener to intercept KeyEvent and prevent it from propagating to parents.
> 
> ```
> SingleChildScrollView(
>         child: RawKeyboardListener(
>               onKey: (RawKeyEvent event) {},
>               focusNode: FocusNode(onKey: (FocusNode node, RawKeyEvent event) {
>                   return KeyEventResult.skipRemainingHandlers;
>               }),
>               child: QuillEditor(
>                     ...
>               ),
>        ),
> )
> ```

I was trying to create shortcuts to toggle bold and italic using `FocusableActionDetector`. But because of this workaround, I think the shortcut events are not triggered. I was thinking, it would be a really nice option to provide `actions` and `shortcuts` directly from the library.

Or can anyone suggest another approach to create shortcuts while using the editor?

PR #618, opened 2022-01-26
Fix certain keys not working on web when editor is in a scroll view (#304)

NicolasDionB:
Implements the fix proposed by @wliumelb from issue #304

### DIFF
--- a/lib/src/widgets/editor.dart
+++ b/lib/src/widgets/editor.dart
@@ -469,10 +469,26 @@ class QuillEditorState extends State<QuillEditor>
       floatingCursorDisabled: widget.floatingCursorDisabled,
     );
 
-    return _selectionGestureDetectorBuilder.build(
+    final editor = _selectionGestureDetectorBuilder.build(
       behavior: HitTestBehavior.translucent,
       child: child,
     );
+
+    if (kIsWeb) {
+      // Intercept RawKeyEvent on Web to prevent it from propagating to parents
+      // that might interfere with the editor key behavior, such as
+      // SingleChildScrollView. Thanks to @wliumelb for the workaround.
+      // See issue https://github.com/singerdmx/flutter-quill/issues/304
+      return RawKeyboardListener(
+        onKey: (_) {},
+        focusNode: FocusNode(
+          onKey: (node, event) => KeyEventResult.skipRemainingHandlers,
+        ),
+        child: editor,
+      );
+    }
+
+    return editor;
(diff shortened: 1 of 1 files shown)


galenwarren:
@NicolasDionB Doesn't the FocusNode introduced here need to be disposed?

NicolasDionB:
@galenwarren You may be right. Anyway a disposal could be good. If you could provide a PR that would be great!

galenwarren:
Thanks @NicolasDionB

I came across this trying to get keyboard-focus tab order to work cleanly in Flutter web with the QuillEditor control. There were two issues wrt this code:

```
      return KeyboardListener(
        onKeyEvent: (_) {},
        focusNode: FocusNode(
          onKeyEvent: (node, event) => KeyEventResult.skipRemainingHandlers,
        ),
        child: editor,
      );
```

- The FocusNode's not being reused and disposed was causing the focus tree to get corrupted on rebuild such that I couldn't tab into the control at all
- Its using the default value (true) for `canRequestFocus` was causing it to take focus when using tab, so it would take two presses of the tab key to get into the text content, instead of just one.

Creating it like so:

```
  final _keyboardListenerFocusNode = FocusNode(
    canRequestFocus: false,
    onKeyEvent: (node, event) => KeyEventResult.skipRemainingHandlers,
  );
```

... and then disposing it seems to fix those issues. This is what I'd include in the PR.

However, there's another thing I wanted to ask you about. In my case, I'd like to capture the tab key in the QuillEditor control and have it move to the next field on my form. As far as I can tell, there is no way to do that, because in the lowest level of the control it captures key events and always translates tab to some sort of content. I don't see any way to prevent this behavior or otherwise hook it.

What would you think of about my adding an `onKeyEvent` property to, say, `QuillEditorConfigurations`, to allow the user to specify a key handler that would get applied before applying the built-in handlers for tab, space, etc? It could have the same signature as the one on FocusNode. Or do you have a better idea for how I might capture the tab key and prevent the default behavior in the control?

galenwarren:
@NicolasDionB Wondering if you have any thoughts on the proposal to add an ```onKeyEvent``` handler to allow callers to override handling for particular keys? I can include that in the PR as well, if you're on board.

NicolasDionB:
@galenwarren Personally I like the idea but you should create a different PR for that, more specific to that addition. Also keep in mind that I'm not a senior on this project, just a casual user/contributor 😊

PR #652, opened 2022-02-07
ios simulator cursor fix

MagTuxGit:
https://github.com/singerdmx/flutter-quill/discussions/554


EchoEllet:
The iOS simulator is usually used for development mode, so we might consider removing this plugin and providing a workaround only when running the app in debug mode.

PR #817, opened 2022-05-18
Allow removing font size attribute from text

mark8044:
New changes have been great, but the problem is once you set a font size, you cannot unset the font size to no font size attribute in the delta.  

With this change, if a font size of `0` is used, then the font size attribute is remove from the text.  This would be a totally optional usage, as the default list of font sizes does not include an option of `0`, but a user can add it to a custom font size list if they wish

### DIFF
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -199,7 +199,15 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
             ],
             onSelected: (newSize) {
               controller.formatSelection(
-                  Attribute.fromKeyValue('size', getFontSize(newSize)));
+              if (newSize == '0') {
+                controller.formatSelection(
+                    Attribute.fromKeyValue('size', null));
+              }
+              else
+                {
+                  controller.formatSelection(
+                      Attribute.fromKeyValue('size', getFontSize(newSize)));
+                }
             },
             rawItemsMap: fontSizes,
           ),
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -198,7 +198,6 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
                 ),
             ],
             onSelected: (newSize) {
-              controller.formatSelection(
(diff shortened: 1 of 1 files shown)


singerdmx:
No, use clear button

mark8044:
> No, use clear button

But that would remove everything, this would work similar to other buttons in the toolbar.

For example, if you have some text with bold and italic and then click the 'B' it would only remove the bold and leave the italic.   

If you use had font size and bold and italic, you would have to reset the bold and italic afterwards.

This would follow the logic of the other buttons in the tool bar that can be 'on' or 'off'.  Maybe if `0` is a bad idea, you could use a different variable?

singerdmx:
For the least please format your code

mark8044:
> For the least please format your code

Sorry I edit everything in GitHub and hard, I will try to see if I can make it better :)

mark8044:
Ok I just updated it, I think the code is better?

singerdmx:
I have to say, you should improve coding

https://github.com/singerdmx/flutter-quill/commit/b401cc5518de56ab9f4168b5f1f91d33de41aa0c

singerdmx:
@mark8044 Also you probably can add similar feature for font family as below

image

singerdmx:
Sample JSON:
```
  {
    "attributes": {
      "font": "sans-serif"
    },
    "insert": "Sans Serif"
  },
  {
    "insert": " "
  },
  {
    "attributes": {
      "font": "serif"
    },
    "insert": "Serif"
  },
  {
    "insert": " "
  },
  {
    "attributes": {
      "font": "monospace"
    },
    "insert": "Monospace"
  },
```

mark8044:
> @mark8044 Also you probably can add similar feature for font family as below
> 
> image

Yes I would love this as well.  I can start on this in a few days.  

I originally had the font size drop down as a generic drop down so it could be re-used for this, but now its refactored it to be specific for font size.  Should I make a new drop down specific to font size, or refactor the font size to be a generic drop down that can be re-used for font family and font size?

singerdmx:
No need for generic use

singerdmx:
@mark8044  Did you start anything? If not I am going to implement this

Issue #809, opened 2022-05-11
Not all users are using 3.0.0. So please don't scream just because you are using 3.0.0.


deimantasa:
My issue is about [Mobile]
My issue is about [Desktop]

After switching to `flutter 3.0.0`, `flutter_quill 4.1.9` stopped working and is throwing exception on launch.

```
...
Running Gradle task 'assembleDebug'...
../../.pub-cache/hosted/pub.dartlang.org/flutter_quill-4.1.9/lib/src/widgets/raw_editor.dart:233:7: Error: The non-abstract class 'RawEditorState' is missing implementations for these members:
 - TextInputClient.insertTextPlaceholder
 - TextInputClient.removeTextPlaceholder
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.
...
```

Tried on Mobile and Desktop. Most likely it affects all platforms.

Details
```
[✓] Flutter (Channel stable, 3.0.0, on macOS 12.3.1 21E258 darwin-arm, locale en-HK)
    • Flutter version 3.0.0 at /Users/Aurimas/FlutterProject/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ee4e09cce0 (2 days ago), 2022-05-09 16:45:18 -0700
    • Engine revision d1b9a6938a
    • Dart version 2.17.0
    • DevTools version 2.12.2
 ```

singerdmx:
not all users are using 3.0.0.
So please don't scream just because you are using 3.0.0.

Issue #808, opened 2022-05-11
[Web] and [Mobile] Flutter 3.0.0 Support


myfknoll:
While updating Flutter to 3.0.0 during Google IO i am receiving following error:

../../sdk/.pub-cache/hosted/pub.dartlang.org/flutter_quill-4.1.9/lib/src/widgets/raw_editor.dart:233:7: Error: The non-abstract class 'RawEditorState' is missing implementations for these members:
 - TextInputClient.insertTextPlaceholder
 - TextInputClient.removeTextPlaceholder
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

myfknoll:
Why closed?

shkvoretz:
Yeah, why closed?

oliverbytes:
Please re-open. I got the issue after upgrading to Flutter 3.0 as well.

PR #832, opened 2022-05-22
flutter 3.0.0 support

DetachHead:
i'm not sure why #808 and #809 were closed, it looks like the fix is pretty simple


mark8044:
> i'm not sure why #808 and #809 were closed, it looks like the fix is pretty simple

It looks like 3.0.0 support was already merged to version 5.0.0

PR #877, opened 2022-07-21
Added CustomBlockEmbed and customElementsEmbedBuilder

LucazzP:
### Feature

Add the possibility to the developer to add custom blocks to the texts, so he will be able to use custom widgets inside the text. I added on example a possible custom block to add notes inside the text, but it can be anything.
I have a video with the example running with some notes added, to show that it works exactly as any other block, like the image or video.

### Motivations

The client of the app that I am developing requested some extra functionalities to the text editor, like the Notion does add tables and everything else that is not only text and images...

### How I made this

So to be able to add this feature, I needed to add this `CustomBlockEmbed` to create the data of the widget, and the `customElementsEmbedBuilder` to show the widget that I want with each type of Custom Block that I created.

### Example

https://user-images.githubusercontent.com/43015129/180117255-bc2e2591-5ae1-40a3-ad5e-4393b8baf5eb.mov


singerdmx:
Looks good!
Could you update README as well ?

LucazzP:
> Looks good! Could you update README as well ?

Yes! I will update as soon as I can.



> Also if you are adding new features like tables, please merge it here instead of CustomBlockEmbed so that it can be used by others. Thanks!

Of course! For now, he only wants custom widgets for the use of the app, tables is something that was not on our roadmap, but is a great idea to add haha, so when I add this, I will add here too.

singerdmx:
> > Looks good! Could you update README as well ?
> 
> Yes! I will update as soon as I can.
> 
> > Also if you are adding new features like tables, please merge it here instead of CustomBlockEmbed so that it can be used by others. Thanks!
> 
> Of course! For now, he only wants custom widgets for the use of the app, tables is something that was not on our roadmap, but is a great idea to add haha, so when I add this, I will add here too.

Thanks!

debmittarv:
Is there any option to customize the delete functionality of image. I am using embededBuilder for custom ui . I am able to add image , but I am not able to delete the particular image from  editor. please help me....
Thank you in advance.

dawarepramod4:
How to add the Button for this custom Builder in ToolBar..
please let us know

adinghub:
How do I transfer a block to export the original text and then read it using the original text?

karroze:
What delta JSON was used in this example?

EchoEllet:
> Not working with latest versions

Could you please provide more details?

Issue #898, opened 2022-08-02
[Feature Request] Maths Support


garv-shah:
I was wondering if it was possible to add maths support with LATEX? [This](https://pub.dev/packages/math_keyboard) plugin could be quite good to implement something like that.

I'd be happy to do it myself, but I'm not too sure how to add a new element to the page UI. I skimmed through the documentation but it wasn't quite clear, would I start with the `text_block.dart` file?

Thank you very much!! :)

garv-shah:
format for the maths input?

singerdmx:
Need the corresponding attribute for it. Please check quill documentation.

garv-shah:
i presume it wouldn't be within the main quill package but rather in something likes mathquill

singerdmx:
We need to make sure web version of react quill can be compatible

garv-shah:
I'm pretty sure it is, since there are quite a few "react-mathquill" type packages which seem to work together from a quick google search

singerdmx:
So what is it like for underlying json data?

garv-shah:
I think it would just have a tag like normal, and then the math would be stored as LATEX

If you can guide me through how to add additional components, I could try to add it myself?

singerdmx:
Please go through https://youtube.com/playlist?list=PLbhaS_83B97vONkOAWGJrSXWX58et9zZ2 if you don't understand what I mean by attribute.
Again, the first thing is to define json data as in

https://github.com/singerdmx/flutter-quill/blob/master/example/assets/sample_data.json

garv-shah:
It's all maths stored as LATEX (or KATEX) inside the JSON
If you look at https://quilljs.com/docs/formats/ it's the "Formula" attribute, the same as normal maths in quilljs, and it can be edited with the package I linked above
I can't find an example of a JSON with the data, but if you can find a comprehensive quilljs one, you should be able to find it

garv-shah:
oh, i watched the videos, and it seems you went through the same website I sent :P
so yeah, the attribute we're looking at is the embed that's not supported yet, which is formulas

singerdmx:
Try https://codesandbox.io/s/3zf03 and you can intercept web traffic to figure out the json

garv-shah:
i got it typed in, how do I get the JSON? sorry, i can't find it in dev tools

singerdmx:
[...]

PR #899, opened 2022-08-03
added formula to sample data json

garv-shah:


singerdmx:
OK I wasn't clear about this.
You need to add corresponding code to handle this attribute. Otherwise the editor will crash.

garv-shah:
almost already done in my fork from above, but I can't figure out where the attributes for embeds are, they don't seem to be defined in the same `attribute.dart` file as the rest. would you please be able to take a look [here](https://github.com/singerdmx/flutter-quill/compare/master...garv-shah:flutter-quill:master) at what I changed and help me out to where to add this?

singerdmx:
I think you can follow `codeBlock`

image

garv-shah:
but that scope is inside block, according to the quilljs documentation, this should be under the embed scope
i'd either have to follow image or video, but i can't find the line similar to 317 in `attribute.dart` for codeBlock

singerdmx:
Yes, it should be embed, not codeblock

singerdmx:
Then you don't need to define new attribute since it is embed

garv-shah:
oh nice, so should I submit a PR, and we can go back to https://github.com/singerdmx/flutter-quill/issues/898#issuecomment-1203351039 ?

singerdmx:
Ok it is new button anyway 
If it has issue other people can help fix

singerdmx:
Worst case you hide it

garv-shah:
cool, thanks!

PR #900, opened 2022-08-03
Added Formula Button (for maths support)

garv-shah:
https://github.com/singerdmx/flutter-quill/issues/898

### DIFF
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -110,6 +110,7 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
     bool multiRowsDisplay = true,
     bool showImageButton = true,
     bool showVideoButton = true,
+    bool showFormulaButton = false,
     bool showCameraButton = true,
     bool showDirection = false,
     bool showSearchButton = true,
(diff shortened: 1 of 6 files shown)


lib/src/widgets/toolbar.dart
```diff
@@ -364,6 +365,18 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
             webVideoPickImpl: webVideoPickImpl,
             iconTheme: iconTheme,
           ),
+        if (showImageButton)
```
singerdmx, line 368: Could you use a new variable?

garv-shah: oh yup, sure

garv-shah: done!

singerdmx:
@garv-shah https://github.com/singerdmx/flutter-quill/commit/c507e4fccd97ce7c586c2012f882e0887e69d929

garv-shah:
Ah, apologies, it must not have been handling the delta at all.
I believe you said that the attributes for embeds were automatic, it that's not the case would you know where I can handle the delta?

singerdmx:
Delta has to have a line break at the very end

singerdmx:
Don't insert at the end

garv-shah:
oh, whoops! alright, I'll mess around with it some more later, thanks :)

Issue #930, opened 2022-08-31
High dependency count


Jon-Salmon:
Firstly thank you very much for this plugin, it is awesome!

The embedded blocks are great features, but supporting them does add a significant overhead to the app (>2mb in android apk size and ~20 dependencies). It would be nice to have a way to reduce this overhead for apps that don't need these advanced features, but I'm aware this is easier said than done (and would be a major breaking change).

My proposal for this would be:
- Split the plugin into two plugins, a 'core' package and a 'embeds' package
- To use the full-featured option, a user would have to either add an extra parameter to their use of quill to load the extra widgets, or maybe there could be a static global method of loading the extra widgets on app startup.

I am very happy to perform such a refactor and submit a pull request, but wanted to check if this was something you were interested in supporting @singerdmx before I committed too much time to this as it would be a breaking change.

singerdmx:
Does it mean we need to publish two packages in https://pub.dev/ ?

Jon-Salmon:
Yes, that would unfortunately be required.

singerdmx:
This kind of things have been discussed.
But it seems not an ideal solution

singerdmx:
That being said, moving embed stuff to example and just asking users to copy/paste code may be a better solution than maintaining two packages.

singerdmx:
Let's start with something small.
First separate the code of `embed` into a standalone folder.
Then we decide if this folder should go to the example or a different package.

Jon-Salmon:
Ok, I'll have a crack at that and open a PR shortly

singerdmx:
![image (3)](https://user-images.githubusercontent.com/122956/187790641-37642a4a-22b4-4bb7-b890-9036f995927c.png)
Here is the old thought. The right side box is the `core`

singerdmx:
So instead of one example folder, we are going to have 3.
I prefer to keep current example folder for mobile.

singerdmx:
Main idea is still keeping one package. 
Users can refer to the folder to do whatever.

Jon-Salmon:
The only issue with that is that either the user would then have to add all the dependencies for images, videos etc themselves, or otherwise we would be back where we started with the bloat.

My original proposal would end up with a repo structure similar to this, with only the one example folder:

repo/
├─ flutter_quill/
│  ├─ lib/
│  ├─ example/
├─ flutter_quill_extensions/
│  ├─ lib/

singerdmx:
repo/
├─ lib/
[...]

PR #933, opened 2022-09-03
Embed refactor

Jon-Salmon:
Draft proposal for moving embed logic to separate package. See #930


lib/src/widgets/editor.dart
```diff
@@ -9,17 +9,13 @@ import 'package:flutter/services.dart';
 import 'package:i18n_extension/i18n_widget.dart';
 import 'package:tuple/tuple.dart';
 
-import '../models/documents/document.dart';
+import '../../flutter_quill.dart';
```
singerdmx, line 12: Please do not import everything

lib/src/widgets/embeds.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:flutter/material.dart';
+
+import '../models/documents/nodes/leaf.dart' as leaf;
+import '../models/themes/quill_dialog_theme.dart';
+import '../models/themes/quill_icon_theme.dart';
+import 'controller.dart';
+
+abstract class IEmbedBuilder {
```
singerdmx, line 8: Prefixing `I` seems to be a java convention.
Is it really necessary?

lib/src/widgets/editor.dart
singerdmx: Don't use `import '../../flutter_quill.dart'`

lib/src/widgets/embeds.dart
Jon-Salmon: Not sure if it is generally considered language specific, but it is definitely standard practice in C# as well. Happy to remove it though.

singerdmx: Yeah in C# it is

singerdmx:
Looks good!
Could you please update the main README to go in together with this change.
For the least, it should mention the second package.

singerdmx:
Looks good!
Just update the title or this PR and let's merge

pubspec.yaml
```diff
@@ -15,22 +15,15 @@ dependencies:
   collection: ^1.16.0
```
singerdmx, line 15: Let's bump the version to 6.0.0 on line 3

singerdmx:
I will do the rest

Jon-Salmon:
You're impressively quick off the mark, I've got one more tweak incoming. I'll open a new PR for it.

singerdmx:
OK

PR #950, opened 2022-09-15
Add ability to paste images

friebetill:


singerdmx:
Is it possible to implement a version of  `onImagePaste` in example folder HomePage?

singerdmx:
This seems to be extremely similar to `OnImagePickCallback`

singerdmx:
Maybe an example like this?

image

PR #951, opened 2022-09-15
Add onImagePaste in example

singerdmx:


friebetill:
I didn't get around to reviewing this yesterday. The example looks good 👍

singerdmx:
@friebetill dart analysis complains the import is redundant. If you need it you can add it back.
This import seems to be inside foundation already so there is no need

friebetill:
Ah I have a guess that it is because of the Flutter version. In Flutter 3.0.X the import is necessary. Probably it is redundant with 3.3.X. Unfortunately I can't test it right now, because I'm on the road and have hardly any internet. I think you are probably right and we should remove the import again, sorry.

singerdmx:
It is fine. We can remove those when most people are upgraded

Issue #955, opened 2022-09-17
[Desktop] Editor loses focus when toolbar buttons are pressed.


Blquinn:
I experienced this on desktop, it may be the same for web. Current master a855f59922a8efef1d654a23ea1d325e8f587eda

Title pretty much says it. Whenever a toolbar button is pressed, the editor loses focus. This makes to so that if you click bold, for example, the cursor is lost and you need to click back into the editor to recover it. When you do click the editor, the active text is no longer bolded. (This is when there is no selection.)

> Flutter 3.3.0 • channel stable • https://github.com/flutter/flutter.git
> Framework • revision ffccd96b62 (3 weeks ago) • 2022-08-29 17:28:57 -0700
> Engine • revision 5e9e0e0aa8
> Tools • Dart 2.18.0 • DevTools 2.15.0

Blquinn:
Closing this since, this can now be sovled by requesting focus in the afterButtonPressed callback of the editor.

hcanyz:
Possibly a better solution: set ```onTapOutside``` to ```QuillEditorConfigurations``` to manually control focus.

The focus thief is ```QuillRawEditorState``` ```_defaultOnTapOutside```

maelchiotti:
> Possibly a better solution: set `onTapOutside` to `QuillEditorConfigurations` to manually control focus.
> 
> The focus thief is `QuillRawEditorState` `_defaultOnTapOutside`

Hi, I have this issue, can you show how you would do that? I don't understand what you exactly mean by "manually control focus". Thanks

hcanyz:
@maelchiotti, this can be done

```
    QuillEditor.basic(
      focusNode: _focusNode,
      configurations: QuillEditorConfigurations(
        controller: _quillController,
        onTapOutside: (_, __) {
          // _focusNode.unfocus(); or doNothing
        },
      ),
    )
```
If ```onTapOutside``` is set then ```QuillEditor``` will not control focus. We can control it ourselves through ```_focusNode```.

md-rifatkhan:
@hcanyz this doesn't work when toolbar button is pressed

PR #956, opened 2022-09-17
Fixes toolbar buttons stealing focus from editor.

Blquinn:
ISSUE: [955](https://github.com/singerdmx/flutter-quill/issues/955)

I think another decent option would be to simply pass the FocusNode to the buttons instead of having this callback.


singerdmx (changes requested):
Update gitignore for vscode

Blquinn:
I was fixing the gitignore... Did you mean to merge that?

singerdmx:
It is fine

Blquinn:
Ok :+1:

Issue #617, opened 2022-01-25
bug: rules are not accessible any more


mrverdant13:
## Details

When working with custom controller behaviour, the user would use the `Document.setCustomRules` method, which is still available.

However, this method only receives a collection of instances of type `Rule`, which is not properly exported by the library.

The `lib/flutter_quill.dart` barrel file:
```dart
library flutter_quill;

export 'src/models/documents/attribute.dart';
export 'src/models/documents/document.dart';
export 'src/models/documents/nodes/embeddable.dart';
export 'src/models/documents/nodes/leaf.dart';
export 'src/models/quill_delta.dart';
export 'src/models/themes/quill_dialog_theme.dart';
export 'src/models/themes/quill_icon_theme.dart';
export 'src/widgets/controller.dart';
export 'src/widgets/default_styles.dart';
export 'src/widgets/editor.dart';
export 'src/widgets/link.dart' show LinkActionPickerDelegate, LinkMenuAction;
export 'src/widgets/style_widgets/style_widgets.dart';
export 'src/widgets/toolbar.dart';
```

singerdmx:
`CustomRules` should be removed and not supported anymore

singerdmx:
No more customization. Any rule needed should be in the repo. People should contribute their rules here.

Issue #582, opened 2022-01-11
feat: support custom inline attributes


mrverdant13:
## Details

Expose a way to register custom inline attributes.

Devs often need a way to use custom attributes depending on some specific use cases. i.e. mentions, tags, etc.

## Related issues and discussions

- Mentions: https://github.com/singerdmx/flutter-quill/discussions/564
- Custom image: https://github.com/singerdmx/flutter-quill/discussions/559
- Custom embed parameters: https://github.com/singerdmx/flutter-quill/discussions/569

I am working on a really simple PR that might enable this feature but would like to have additional feedback once I submit it.

singerdmx:
No I am against the idea

singerdmx:
Anything needs to be added should be directly added into repo

singerdmx:
Such PR will not be approved

singerdmx:
Custom styling is already supported 
No custom attributes 
If you need any new attributes, it should be in the repo

mrverdant13:
That's a shame.
This would promote the development of external packages to be plugged to `flutter_quill`.
No worries.

singerdmx:
We have enough bugs to fix, e.g. https://github.com/singerdmx/flutter-quill/issues/561 (I think you are working on this).
We don't need unnecessary complication to the repo. This will only make maintainability worse.
Also we don't want people just to use this repo without contributing to it such as adding more inline attribute feature.

PR #971, opened 2022-10-09
Changes to support custom attributes

jkidi:
Even though there is existing code to support custom attributes/rules there doesn't seem to be a way to actually use them without running into some type of issue. This PR addresses these issues along with some other related fixes.

1. Custom attribute with scope != IGNORE. (#582)
For example, if you attempt to add an INLINE custom attribute then you might see the following error:
```
Assertion failed: file:///lib/src/models/documents/nodes/leaf.dart:30:12
value.isInline || value.isIgnored || value.isEmpty
"Unable to apply Style to leaf: {Attribute{key: custom-attr, scope: AttributeScope.IGNORE, value: custom}}"
```

This is caused by remapping any attribute that is not found in the Attribute registry to scope IGNORE. So even though you might have defined your attribute as INLINE it will get changed to IGNORE when reading the document from json. The following code in lib/src/models/documents/style.dart:14 is causing this:

 ```dart
 static Style fromJson(Map<String, dynamic>? attributes) {
    if (attributes == null) {
      return Style();
    }

    final result = attributes.map((key, dynamic value) {
      final attr = Attribute.fromKeyValue(key, value);
      return MapEntry<String, Attribute>(
          key, attr ?? Attribute(key, AttributeScope.IGNORE, value));
    });
    return Style.attr(result);
  }
```

The easy way to handle this would be to allow adding custom attributes to the registry which is what this PR does. Ideally you just add the custom attribute to the registry when the app starts. This in turn allows you to add custom attributes that take advantage of existing rules.

```dart
void main() {
  Attribute.addCustomAttribute(const CustomAttribute());
  runApp(MyApp());
}
```

2. Custom attribute with scope == IGNORE
If you try to add a custom attribute with scope IGNORE then you will run into another issue, mainly not being able to add custom rules to handle the custom attribute. The Rule class not being exported means that new rules cannot be added. See issue #617. This PR exports the Rule class.

3. Custom button styling
Custom button styling is limited and currently doesn't apply the styles defined in icon theme. Made the necessary changes to fix that.

4. Fixed the read only demo page and added a demo page for custom attributes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,81 @@ QuillToolbar.basic(
     ]
 ```
 
+### Custom Attributes
+Often times we want to apply a style that is not available by default with the use of existing toolbar buttons. For such cases, we can simply add a custom toolbar button to apply a custom attribute to the selected text.
+
+We can create a custom attribute by defining the class as such:
+```dart
+class CustomAttribute extends Attribute<bool?> {
+  const CustomAttribute(bool? val)
+      : super(KEY, AttributeScope.INLINE, val);
+
+  static const String KEY = 'my-custom-attr';
+}
+```
+
+We should then register the attribute when the app is started.
+
+```dart
+void main() {
+  Attribute.addCustomAttribute(const CustomAttribute(null));
+  runApp(MyApp());
+}
+```
+
+After that we can define a custom style builder and handle the text style that will be applied for the given attribute. For example, the below code will apply a blue color and yellow background if a text contains this attribute.
+```dart
+  TextStyle _customStyleBuilder(Attribute attr) {
+    if (attr.key == CustomAttribute.KEY) {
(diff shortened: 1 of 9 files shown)


singerdmx:
Sorry no custom attribute 
All attributes should be explicitly defined in quill 
Or explicitly supported here

singerdmx:
Please break down your changes in the future 
Gigantic PR is not welcome 
If you work in industries you should break into small changes

jkidi:
All changes are related therefore a "gigantic" PR. I think not allowing custom attributes while most of the existing code already supports it is a little weird. In the end it's your choice whether to merge or not.

singerdmx:
No this can be broken into small changes and stop making excuses

PR #986, opened 2022-10-26
Add keyboard shortcuts for editor actions

Blquinn:
This adds a bunch of keyboard shortcuts to the editor.

I duplicated a lot of code from the toolbar buttons and I think it would be a good idea to figure out how to get that in one place. Perhaps the toolbar buttons can invoke the Intents/Actions?


singerdmx:
Thanks! I will digest this PR over the weekend

PR #994, opened 2022-11-02
Add typing shortcuts

Blquinn:
This adds typing shortcuts that are similar to quill-js and apple's notes application.

Start lists with "1." or "-". When you press space after either of these strings, it will automatically format them as actual lists.

Indent / Dedent lists with <Tab> and <Shift+Tab>.

### DIFF
--- a/lib/src/widgets/controller.dart
+++ b/lib/src/widgets/controller.dart
@@ -102,6 +102,26 @@ class QuillController extends ChangeNotifier {
         .mergeAll(toggledStyle);
   }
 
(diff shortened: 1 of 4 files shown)


lib/src/widgets/controller.dart
```diff
@@ -102,6 +102,26 @@ class QuillController extends ChangeNotifier {
         .mergeAll(toggledStyle);
   }
 
+  // Increases or decreases the indent of the current selection by 1.
```
Blquinn, line 105: I decided to move this method into the controller, instead of copying it into a third place.

PR #1041, opened 2022-12-31
fix: The non-abstract class 'RawEditorState' is missing implementations

Vorun5:
Fix:
Error: The non-abstract class 'RawEditorState' is missing implementations for these members:
 - TextInputClient.didChangeInputControl
 - TextInputClient.performSelector


singerdmx:
Which channel are you using ?

Vorun5:
> Which channel are you using ?

image

I got this error when using flutter pen 6.1.11 (OS: Windows 10). By changing the code as in the pull request, this error disappeared and everything worked.

mtallenca:
these changes look like they are required for the just released stable channel in flutter 3.7

Issue #1053, opened 2023-01-10
[Mobile] [Web] [Desktop] Align numerical list and bullet points with text content


Dnathan33:
My issue is about [Web] [Mobile] [Desktop]

When aligning a text today with numerical or bullet points applied, the text is aligned however, the numerical or bullet contents are not.

See screenshot below

![Simulator Screen Shot - iPhone 13 - 2023-01-10 at 09 54 40](https://user-images.githubusercontent.com/39175434/211584273-129c8614-765f-477f-8f1a-a0c64e9f71a8.png)

Dnathan33:
#1054 solves this issue

AndreVicenteW:
For me this error still happens in version 7.4.9

freshtechtips:
Thank you for the report, this issue has been open for a while and since there are no interactions with it, we had to close it

but if you still facing this issue, or other issues or you have any other questions, feel free to contact us anytime you want.

freshtechtips:
> For me this error still happens in version 7.4.9. Is there anything else I should do?

Hi, if you still facing this issue please create a new issue

PR #1054, opened 2023-01-10
Aligns numerical and bullet lists with text content

Dnathan33:
This PR fixes the issue with numerical and bullet points always being aligned to the left even when the text is center or right aligned. Quill.js has support for numerical and bullet points being aligned with the text content and this PR adds that support for Flutter Quill.

This PR solves issue #1053

Look at the screenshots below to see the differences.

**Flutter Quill: Current**

Simulator Screen Shot - iPhone 13 - 2023-01-10 at 09 54 40

**Flutter Quill: Updated**

Simulator Screen Shot - iPhone 13 - 2023-01-09 at 17 24 05

**Quill JS** - https://quilljs.com/docs/formats/

Screen Shot 2023-01-09 at 5 22 12 PM

### DIFF
--- a/lib/src/widgets/text_block.dart
+++ b/lib/src/widgets/text_block.dart
@@ -135,6 +135,7 @@ class EditableTextBlock extends StatelessWidget {
           _buildLeading(context, line, index, indentLevelCounts, count),
           TextLine(
             line: line,
+            index: index,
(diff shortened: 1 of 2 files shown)


Dnathan33:
Solves  #1054

PR #1073, opened 2023-01-31
Added `unknownEmbedBuilder` to QuillEditor

s0nerik:
Providing an `unknownEmbedBuilder` makes it possible to gracefully handle the unknown embed type. Without it, the `QuillEditor` throws an error, which can be undesirable in many cases.

### DIFF
--- a/lib/src/widgets/editor.dart
+++ b/lib/src/widgets/editor.dart
@@ -173,6 +173,7 @@ class QuillEditor extends StatefulWidget {
       this.onSingleLongTapMoveUpdate,
       this.onSingleLongTapEnd,
       this.embedBuilders,
+      this.unknownEmbedBuilder,
       this.linkActionPickerDelegate = defaultLinkActionPickerDelegate,
       this.customStyleBuilder,
       this.locale,
--- a/lib/src/widgets/editor.dart
+++ b/lib/src/widgets/editor.dart
@@ -359,6 +359,7 @@ class QuillEditor extends StatefulWidget {
       onSingleLongTapEnd;
 
   final Iterable<EmbedBuilder>? embedBuilders;
+  final EmbedsBuilder? unknownEmbedBuilder;
   final CustomStyleBuilder? customStyleBuilder;
 
   /// The locale to use for the editor toolbar, defaults to system locale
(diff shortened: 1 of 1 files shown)


singerdmx:
Why do we need this?

singerdmx:
Everything should be explicitly handled. And the name is really ugly.

s0nerik:
> Why do we need this?

Imagine a situation where the server adds support for a new type of embed. Another app (or even the same one, but a newer version, that also supports it) creates some content with that embed. The older app version should not break upon displaying the content with such an embed. It's much better to have a placeholder with the "content is not supported" part instead of an error.

With `EmbedBuilder` it is just impossible to provide a fallback for such a case.

> Everything should be explicitly handled. And the name is really ugly.

I can't think of a better name for it. Unless your app can see into the future - the embed received there is actually unknown to the app.

singerdmx:
correct approach is to make flutter quill support multiple custom embed builder.
Currently only one is supported.

s0nerik:
> correct approach is to make flutter quill support multiple custom embed builder.
Currently only one is supported.

I already does. And did that for a long time already.

singerdmx:
Sorry I don't understand you

s0nerik:
Here's how we were doing custom embeds for more than half a year now:

```dart
class CustomEmbeddable extends quill.Embeddable {
  const CustomEmbeddable._(super.type, super.data);

  static const String customType1 = 'custom1';
  static const String customType2 = 'custom2';

  static CustomEmbeddable custom1({
    required String field1,
    required String field2,
  }) =>
      CustomEmbeddable._(customType1, <String, dynamic>{
        'field1': field1,
        'field1': field1,
      });
  
  static CustomEmbeddable custom2({
    required String field1,
    required String field2,
  }) =>
      CustomEmbeddable._(customType2, <String, dynamic>{
        'field1': field1,
        'field1': field1,
      });

...

}
```

The pre-`v6.0.0` way of building the embedded widgets allowed for easy fallback widget definition. `v6.0.0` totally hides the ability to match the embed by `type` yourself and doesn't allow for specifying a fallback if the app doesn't know how to handle the embed type it got.

PR #1089, opened 2023-02-07
Fixed handling of mac intents

ericmartineau:
Without this method implemented, the quill editor is dropping platform-specific intents.  This may be happening on other platforms, but I know this fix works for macos


matthiasn:
Thanks a lot @ericmartineau, works as expected.

PR #1101, opened 2023-02-15
Vertical toolbar

adil192:
I've had to make the `QuillToolbar` constructor non-`const` to avoid breaking people's code (`toolbarHeight` is now deprecated and replaced with `toolbarSize`). If you'd like to keep the const constructor at the expense of breaking changes, let me know and I'll remove the `toolbarHeight` fallback.

image

### DIFF
--- a/lib/src/widgets/toolbar.dart
+++ b/lib/src/widgets/toolbar.dart
@@ -48,6 +48,7 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
     this.axis = Axis.horizontal,
     this.toolbarSize = 36,
     this.toolbarIconAlignment = WrapAlignment.center,
+    this.toolbarIconCrossAlignment = WrapCrossAlignment.center,
     this.toolbarSectionSpacing = 4,
     this.multiRowsDisplay = true,
     this.color,
@@ -63,6 +64,7 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
     double toolbarIconSize = kDefaultIconSize,
     double toolbarSectionSpacing = 4,
     WrapAlignment toolbarIconAlignment = WrapAlignment.center,
+    WrapCrossAlignment toolbarIconCrossAlignment = WrapCrossAlignment.center,
     bool showDividers = true,
     bool showFontFamily = true,
     bool showFontSize = true,
@@ -177,6 +179,7 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
       toolbarSize: toolbarIconSize * 2,
       toolbarSectionSpacing: toolbarSectionSpacing,
       toolbarIconAlignment: toolbarIconAlignment,
+      toolbarIconCrossAlignment: toolbarIconCrossAlignment,
       multiRowsDisplay: multiRowsDisplay,
       customButtons: customButtons,
       locale: locale,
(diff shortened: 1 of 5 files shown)


singerdmx:
what is the vertical about? Screenshot please?

adil192:
It arranges the toolbar icons in a column instead of a row, I'll upload a screenshot shortly :)

adil192:
image

This is how it looks currently (needs some work of course)

singerdmx:
As long as you make this as option (default should be horizontal) so that people get to use it the old way

lib/src/widgets/toolbar.dart
```diff
@@ -43,9 +43,12 @@ const double kDefaultIconSize = 18;
 const double kIconButtonFactor = 1.77;
 
 class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
-  const QuillToolbar({
+  QuillToolbar({
     required this.children,
-    this.toolbarHeight = 36,
+    this.axis = Axis.horizontal,
+    @Deprecated('Use toolbarSize instead')
```
singerdmx, line 49: Just remove this. We will bump a new version.
Please update REAME as necessary

adil192:
The default is horizontal yes

lib/src/widgets/toolbar.dart
singerdmx: Remove `toolbarHeight`

adil192: I've removed `toolbarHeight`.

I'm not sure what to update README with: does it need an example of using the `axis` parameter?

I assume CHANGELOG will also need to be updated to tell people about the change. Do you want me to add that or will you add it when you bump the version?

singerdmx: It looks like REAME does not have any reference to `toolbarHeight` so no action

singerdmx: Yes update CHANGELOG and bump it to 6.4.0

adil192: Okay I'll update CHANGELOG when I've finished the implementation

lib/src/widgets/toolbar.dart
```diff
@@ -532,6 +532,9 @@ class QuillToolbar extends StatelessWidget implements PreferredSizeWidget {
           ? Wrap(
               direction: axis,
               alignment: toolbarIconAlignment,
+              crossAxisAlignment: toolbarIconAlignment == WrapAlignment.center
+                ? WrapCrossAlignment.center
+                : WrapCrossAlignment.start,
```
adil192, line 537: Here I've used `toolbarIconAlignment` to center the icons horizontally in a column. Would you prefer it as a separate parameter in the constructor?

singerdmx: Let's remove `toolbarIconAlignment`

adil192: What should I replace it with?

singerdmx: Because mapping between two different types does not make sense

singerdmx: Please make sure horizontal looks the way it was or at least looks good

adil192: So make a new parameter for `crossAxisAlignment`?

singerdmx: yes

adil192:
The horizontal toolbar looks identical with and without this PR

Before:
image
After:
image

singerdmx:
I left comments

singerdmx:
Please mention crossAxisAlignment in change log

adil192:
> I left comments

I think I can't see them because I re-ordered some commits

singerdmx:
@adil192 Are you interested in fixing this?

image

adil192:
I'll have a look

singerdmx:
> I'll have a look

Thanks

PR #1105, opened 2023-02-16
Remove deprecated `ToolbarOptions`

adil192:
I'm not exactly sure how to test this, but I've observed that
- on Linux, no selection toolbar is shown but Ctrl+C and Ctrl+V still work
- on Android, the selection toolbar is shown

I believe this is correct behaviour

***

image


singerdmx:
@adil192 Why is it still showing this https://pub.dev/packages/flutter_quill/score ?

image

adil192:
I didn't see that, I'll fix that now

PR #1107, opened 2023-02-17
Update some dependencies

mark8044:
Update some dependencies, notable i18n_extension which conflicts with intl updates

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# [6.4.3]
+* Update dependencies (collection: 1.17.0, flutter_keyboard_visibility: 5.4.0, quiver: 3.2.1, tuple: 2.0.1, url_launcher: 6.1.9, characters: 1.2.1, i18n_extension: 7.0.0, device_info_plus: 8.1.0)
+
 # [6.4.2]
 * Replace `buildToolbar` with `contextMenuBuilder`.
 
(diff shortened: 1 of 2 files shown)


singerdmx:
Please update changelog

mark8044:
> Please update changelog

Ok I just did it

adil192:
I think we might need to revert `i18n_extension: ^7.0.0` because it depends on `intl: ^0.18.0` which most packages don't support.

Even in the example app, `flutter pub get` fails unless `intl: ^0.18.0` is added to `dependency_overrides` (which we ideally shouldn't have in a production app). As a temporary hack, I've done this in this commit: https://github.com/adil192/flutter-quill/commit/697759f34c8fcaa7b292925684b5e0b955bc60d9.

I also have to use `dependency_overrides` for `intl` in my app Saber, following https://github.com/provokateurin/nextcloud-neon/issues/195. This is because flutter_localizations (from the flutter sdk) depends on `intl: ^0.17.0`.

I propose that we set `i18n_extension` back to `^6.0.0`, and anyone who absolutely needs to use `^7.0.0` can use something like this in their app's pubspec since `i18n_extension` is less used than `intl`:
```yaml
dependency_overrides:
  i18n_extension: ^7.0.0
  intl: ^0.18.0
```

mark8044:
@adil192 @singerdmx You should actually go UP to `i18n_extension 8.0.0` as that is using `intl: ^0.17.0`

adil192:
I'll make another pull request, thanks for letting us know

mark8044:
@adil192 already there :)

PR #1125, opened 2023-03-06
Improve ImageEmbedBuilderWeb in example

clia:
When using the example in Web browsers, the image is displayed abnormally or not displayed.

The browser debug console outputs: 
```
Height of Platform View type: [https://some.com/some.jpg] may not be set. Defaulting to `height: 100%`.
Set `style.height` to any appropriate value to stop this message.

Width of Platform View type: [https://some.com/some.jpg] may not be set. Defaulting to `width: 100%`.
Set `style.width` to any appropriate value to stop this message.
```

So I added the `width` and `height` style for the image, setting it to 'auto', to make it show properly.

### DIFF
--- a/example/lib/universal_ui/universal_ui.dart
+++ b/example/lib/universal_ui/universal_ui.dart
@@ -44,8 +44,12 @@ class ImageEmbedBuilderWeb implements EmbedBuilder {
       return const SizedBox();
     }
     final size = MediaQuery.of(context).size;
-    UniversalUI().platformViewRegistry.registerViewFactory(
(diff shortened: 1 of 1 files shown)


mozartee:
Because of your setting of auto with width and height, how to set the ImageEmbedBuilderWeb's constraints property?

clia:
> Because of your setting of auto with width and height, how to set the ImageEmbedBuilderWeb's constraints property?

The behavior of these widths and heights is a bit messy and weird, sometimes they still don't show up, and I don't know how to handle them perfectly. Do you have any good ideas?

PR #1128, opened 2023-03-12
Remove tuples

adil192:
I'm working on removing all tuples from the code and replacing them with data classes. This makes the code much more readable and concise.


adil192:
https://github.com/singerdmx/flutter-quill/blob/a864e29ba115209e602003e8b92990a6d46bde2f/lib/src/models/documents/history.dart#L88-L111

Please can you help me understand what this `Tuple2` contains?

adil192:
Also now would be the perfect time to downgrade `i18n_extension` following https://github.com/singerdmx/flutter-quill/pull/1107#issuecomment-1465115838 (if you agree with the suggestion) since we already have a breaking change

singerdmx:
Thanks
Is it ready to merge?

singerdmx:
Downgrade is fine

adil192:
Ready to merge!

adil192:
This will also close
- #1116

ellet0:
Do you know what is `len`? we need to document everything
does it stands for length? length of the document or something? it's not very clear

adil192:
> Do you know what is `len`? we need to document everything does it stands for length? length of the document or something? it's not very clear

You're best off asking @singerdmx since I haven't modified this code:

https://github.com/singerdmx/flutter-quill/blob/8884aa9f7aff2e658ec678c3c3dfa27280b3aadd/lib/src/models/documents/history.dart#L93-L110

It looks like the number of characters that were inserted (positive) or deleted (negative) in the last history change, but I'm unfamiliar with the code

PR #1149, opened 2023-03-31
Highlight the beginning selected, empty lines.

Blquinn:
Highlights the beginning few pixels of empty lines. This makes it much easier to see that you're successfully selecting when trying to select empty regions of the document.

image


Blquinn:
I'm not sure what this looks like for RTL languages, or on mobile. I've only checked on desktop so far.

Issue #1092, opened 2023-02-09
[Desktop][IME] Input method panel place wrong on macos


pedia:
My issue is about [Desktop] of macos

I have tried running `example` directory successfully before creating an issue here. 

<img width="1241" alt="image" src="https://user-images.githubusercontent.com/170561/217816473-fdcc1fd3-6792-45eb-aa5e-ad426b984c79.png">

The `Input method panel` should placed just below the carer. 

# flutter doctor
```text
$ flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.7.2, on macOS 12.5.1 21G83 darwin-x64, locale en-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.3)
[✓] VS Code (version 1.75.0)
[✓] Connected device (2 available)
[✓] HTTP Host Availability

• No issues found!
```

This bug can be reproduced on versions below:
  6.0.10
  6.1.10
  6.1.12
  6.1.8
  6.1.9
  6.2.0
  6.2.1
master 32a69741

Thanks.

pedia:
Yes, It's fixed. I did not notice #1154.

Thanks all.

freshtechtips:
> Yes, It's fixed. I did not notice #1154.
> 
> Thanks all.

If you have any other questions or are facing any other issues 

Feel free to ask us at any time.

Issue #521, opened 2021-12-11
[Desktop]The  Input Method（maybe call it ‘typewriter’，some thing to input chinese） do not follow the cursor


RaySnow:
My issue is about [Desktop]

The  Input Method（maybe call it ‘typewriter’，some thing to input chinese） do not follow the cursor,  system may not think it's an input. on both MacOs and Windows
it works fine in TextField. 

And, there is another problem, in [Desktop] , only windows, only third pary  Input Method（such as Sogou Input or Baidu Input）.
When I type in chinese after H1, it will change the H1 to normal.
When T type first Enter key after H1, nothing happen, when I type the second Enter key, it will change the H1 to normal, and change the next line to H1.

Please help, thans very much.
 
[video](https://youtu.be/MMLVaENIeBk)

PR #1153, opened 2023-04-02
fix macos IME position bug

foreverpw:
Using TextInputConnection.setCaretRect to update IME position on Macos


foreverpw:
#1092 #521

chifandeyu:
image
The problem still exists. The y coordinate of the input method candidate popup is offset below the scrollbar when the cursor is more than one page high.

foreverpw:
#1159

PR #1179, opened 2023-04-17
Toolbar tweaks

bambinoua:
* Implement tooltips for embed `CameraButton`, `VideoButton`, `FormulaButton`, `ImageButton`.
* Extends customization for `SelectAlignmentButton`, `QuillFontFamilyButton`, `QuillFontSizeButton` adding `padding`, text `style`, `alignment`, `width`.
* Add `renderFontFamilieis` to `QuillFontFamilyButton` to show font faces in dropdown.
* Add `AxisDivider` and its named constructors for for use in parent project.
* Export `ToolbarButtons` enum to allow specify tooltips for `SelectAlignmentButton`.
* Export `QuillFontFamilyButton`, `SearchButton` as they were not exported before.
* Deprecate `items` property in `QuillFontFamilyButton`, `QuillFontSizeButton` as the it can be built usinr `rawItemsMap`. 
* Make `onSelection` `QuillFontFamilyButton`, `QuillFontSizeButton` omittable as no need to execute callback outside if `controller` is passed to widget.

Now the package is more friendly for web projects.

### DIFF
--- a/flutter_quill_extensions/lib/embeds/toolbar/camera_button.dart
+++ b/flutter_quill_extensions/lib/embeds/toolbar/camera_button.dart
@@ -55,6 +55,7 @@ class CameraButton extends StatelessWidget {
 
     return QuillIconButton(
       icon: Icon(icon, size: iconSize, color: iconColor),
+      tooltip: tooltip,
       highlightElevation: 0,
       hoverElevation: 0,
       size: iconSize * 1.77,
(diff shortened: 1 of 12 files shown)


Jon-Salmon:
@bambinoua This PR appears to have broken the example app toolbar (at least on Windows). Please fix: 
image

adil192:
Can confirm I'm also having this issue with the toolbar

adil192:
https://github.com/singerdmx/flutter-quill/blob/f3ad1affcab624eb2f8dd81bdf35b9d3d3f460c5/lib/src/widgets/toolbar.dart#L641-L653

I believe this section is wrong. If the toolbar is horizontal, we want a vertical divider.

PR #1182, opened 2023-04-18
Dropdown tweaks

bambinoua:
- Add `itemHeight`, `itemPadding`, `defaultItemColor` for customization of dropdown items.
- Remove `alignment` property as useless.
- Fix bugs with max width when `width` property is null.


bambinoua:
Thank you for quick response! It is a pleasure to collaborate with you. :)

PR #1186, opened 2023-04-20
Update `CHANGELOG.md`, `flutter_quill` version to 7.1.9 and `flutter_quill_extension` version to 0.2.1+1

bambinoua:
But you need to change the version of `flutter_quill_extensions` to `0.2.1+2` because it should has dependency on this `flutter_quill` version, i.e. `7.1.9`. And I could not set this version for `flutter_quill_extensions` as this version is not merged yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,188 +1,272 @@
+# [7.1.9]
+
+- Editor tweaks PR from [bambinoua](https://github.com/bambinoua).
+  - Shortcuts now working in Mac OS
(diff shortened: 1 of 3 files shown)


bambinoua:
So after merge this PR you need to change the version of `flutter_quill` package to `7.1.9` in `flutter_quill_extension`'s `pubspec.yaml` and update this version [here](https://pub.dev/packages/flutter_quill_extensions).

singerdmx:
It is published.
Need @Jon-Salmon  for flutter_quill_extension's

bambinoua:
@singerdmx, why do you move embed buttons to external package? What is the sense?

singerdmx:
Some people only need the core and this way it is less dependency

PR #1194, opened 2023-04-28
add superscript and subscript styles

huqigu:
I have added the superscript and subscript styles, please check for any bugs


lib/src/translations/toolbar.i18n.dart
```diff
@@ -40,6 +40,8 @@ extension Localization on String {
           'Font family': 'Font family',
           'Font size': 'Font size',
           'Bold': 'Bold',
+          'Subscript': 'Subscript',
```
singerdmx, line 43: Add these for all locales

huqigu: > Add these for all locales

Do you need me to translate all languages？

lib/src/models/documents/attribute.dart
```diff
@@ -359,7 +367,7 @@ class TokenAttribute extends Attribute<String> {
 // `script` is supposed to be inline attribute but it is not supported yet
```
singerdmx, line 367: Delete this line

lib/src/widgets/default_styles.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'dart:ui';
```
singerdmx, line 2: Why is it needed?

huqigu: Done.

huqigu: > Why is it needed?

to use FontFeature.subscripts()  and FontFeature.superscripts()

lib/src/models/documents/attribute.dart
huqigu: Done.

PR #1196, opened 2023-04-28
add custom recognizer

huqigu:
Now only links can be clicked, but I want to add gestures to any attribute text， so I refer to customStyleBuilder and added customRecognizerBuilder,  please review.


huqigu:
@singerdmx   If there are no issues,  please publish it,  because I want to use this feature to develop our app.  Thank you very much.

PR #1212, opened 2023-05-12
Updated CHANGELOG and pubspec.yalm

Mmisiek:


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: flutter_quill
 description: A rich text editor supporting mobile and web (Demo App @ bulletjournal.us)
-version: 7.1.17+1
+version: 7.1.18+1
```
adil192, line 3: This should be `7.1.18` or `7.2.0` depending on what @singerdmx prefers

adil192:
Also I think you need to branch off the latest singerdmx/flutter-quill commit rather than just using your previous branch from #1210

Mmisiek:
I agree but I honestly don't have time to deal with it now. Please feel free to do it correctly.

adil192:
Nevermind looks like this has already been done in https://github.com/singerdmx/flutter-quill/commit/0a3fe3f63c4636c68de1e44523eae3b9de787dcb

This PR can be closed

PR #1214, opened 2023-05-12
fixing some wrong doings

francksoudan:


MagTuxGit:
They are not wrong. They are for Flutter 3.10.

francksoudan:
Oh I see…

PR #1230, opened 2023-05-24
pass linestyle to embeded block

doonfrs:
I needed to embed icons in the editor, and I needed the icon to take the same line font size, I found that I had no access to the line style from the embedded builder.
This will be helpful for inline embedded blocks.

image

Example code: I can access the font size now ( and other styles for sure) 
image

### DIFF
--- a/lib/src/widgets/embeds.dart
+++ b/lib/src/widgets/embeds.dart
@@ -21,6 +21,7 @@ abstract class EmbedBuilder {
     leaf.Embed node,
     bool readOnly,
     bool inline,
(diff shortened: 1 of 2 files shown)


bambinoua:
@doonfrs man did you overview the code? Your change is broken! You updated the prototype of `EmbedBuild.build` (add new parameter `textStyle`) but this affects all widges which depends on it namely `ImageEmbedBuilder`, `ImageEmbedBuilderWeb`, `ImageVideoBuilder` and now the project which use them is not compiled!

PR #1231, opened 2023-05-24
Keep the embedded block style

doonfrs:
I inserted an icon as an embedded widget, but I still want it to take the same color/font size as the text; applying the style to the embedded element will allow the developer to read the attributes from the node and apply them to the element if needed.

Result: 
image

Usage example: 
textStyle now refer to the whole line style and node.style refers to the element style (selection).
```
@override
  Widget build(
    BuildContext context,
    flutter_quill.QuillController controller,
    flutter_quill.Embed node,
    bool readOnly,
    bool inline,
    TextStyle textStyle,
  ) {
    final iconInfo = jsonDecode(node.value.data);

    String? iconName;
    if (iconInfo != null) {
      iconName = iconInfo['iconName'];
    }
    iconName ??= 'accessibleIcon';

    var color = Colors.black;
    if (iconInfo != null) {
      var clr = node.style!.attributes['color']?.value?.toString();
      if (clr != null) {
        color = Color(int.parse(clr.replaceAll("#", "FF"), radix: 16));
      }
    }

    return Padding(
        padding: EdgeInsets.only(bottom: (textStyle.fontSize! > 24) ? 0 : 6),
        child: FaIcon(faIconNameMapping[iconName],
            color: color, size: textStyle.fontSize!));
  }

```

image

### DIFF
--- a/flutter_quill_extensions/pubspec.yaml
+++ b/flutter_quill_extensions/pubspec.yaml
@@ -5,7 +5,7 @@ homepage: https://bulletjournal.us/home/index.html
 repository: https://github.com/singerdmx/flutter-quill/tree/master/flutter_quill_extensions
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
+  sdk: ">=2.17.0 <4.0.0"
   flutter: ">=3.0.0"
 
 dependencies:


doonfrs:
I did not notice they are 3 commits, kindly remember to squash the commits into one.

PR #1234, opened 2023-05-25
enhance color picker, enter hex color and color palette option

doonfrs:
* Choose from the color palette
* The user can enter hex color value
* load the current selection color as an initial value

image


doonfrs:
Hi,
Do you have any problems with the PR?
The current picker is limited for actual usage; hex input and color picker are required.

Issue #1306, opened 2023-07-14
[Desktop] I use "enter" key where image taill  to new line , then use "backspace" key can't back to image taill.


TheHeBoy:
I use "enter" key where image taill  to new line , then use "backspace" key can't back to image taill.

https://github.com/singerdmx/flutter-quill/assets/49644098/ffc8d761-8fe4-4c91-95d9-5a838a1a7054

PR #1309, opened 2023-07-16
Fix around image use backspace key can't delete break line.

TheHeBoy:
Fix around image use backspace key can't delete break line. issue #1306

https://github.com/singerdmx/flutter-quill/assets/49644098/8ef9213f-774b-432d-9fb5-27a9b05e3312

I think EnsureEmbedLineRule and InsertEmbedsRule both apply for embed.EnsureEmbedLineRule only work for video, but InsertEmbedsRule for all embed. Must idea control block or inline element by expend of embed. 
```
/// Handles all format operations which manipulate embeds.
/// This rule wraps line breaks around video, not image.
class InsertEmbedsRule extends InsertRule {
  const InsertEmbedsRule();

  @override
  Delta? applyRule(Delta document, int index,
      {int? len, Object? data, Attribute? attribute}) {
    if (data is String) {
      return null;
    }

    assert(data is Map);

    if (!(data as Map).containsKey(BlockEmbed.videoType)) {
      return null;
    }
```

### DIFF
--- a/lib/src/models/rules/delete.dart
+++ b/lib/src/models/rules/delete.dart
@@ -1,4 +1,5 @@
 import '../documents/attribute.dart';
+import '../documents/nodes/embeddable.dart';
 import '../quill_delta.dart';
 import 'rule.dart';
 
@@ -118,6 +119,12 @@ class EnsureEmbedLineRule extends DeleteRule {
     final itr = DeltaIterator(document);
 
     var op = itr.skip(index);
+    final opAfter = itr.skip(index + 1);
+
+    if (!_isVideo(op) || !_isVideo(opAfter)) {
+      return null;
+    }
+
     int? indexDelta = 0, lengthDelta = 0, remain = len;
     var embedFound = op != null && op.data is! String;
     final hasLineBreakBefore =
@@ -156,5 +163,11 @@ class EnsureEmbedLineRule extends DeleteRule {
     return Delta()
       ..retain(index + indexDelta)
       ..delete(len! + lengthDelta);
+  }
+
+  bool _isVideo(op) {
+    return op != null &&
+        op.data is! String &&
+        !(op.data as Map).containsKey(BlockEmbed.videoType);
   }
 }
--- a/lib/src/models/rules/delete.dart
+++ b/lib/src/models/rules/delete.dart
@@ -109,7 +109,10 @@ class PreserveLineStyleOnMergeRule extends DeleteRule {
   }
 }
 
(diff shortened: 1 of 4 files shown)


singerdmx:
This PR contains previous changes

singerdmx:
Can you rebase？

singerdmx:
Is this about image or video？

TheHeBoy:
About image.

lib/src/models/rules/delete.dart
```diff
@@ -118,6 +119,12 @@ class EnsureEmbedLineRule extends DeleteRule {
     final itr = DeltaIterator(document);
 
     var op = itr.skip(index);
+    final opAfter = itr.skip(index + 1);
+
+    if (!_isVideo(op) || !_isVideo(opAfter)) {
```
singerdmx, line 124: Could you add comment what this is about?

singerdmx: It seems that EnsureEmbedLineRule should be renamed to EnsureVideoLineRule?
This rule is only applicable to video?
why is it not applicable to image?

singerdmx: Does this change work with content that has video?

singerdmx: If you want to exclude image, you can check:

```
if (!_isImage(op) || !_isImage(opAfter)) {
   return null;
}
```

singerdmx: You basically want to exclude image from this rule, right?

TheHeBoy: Because InsertEmbedsRule only works for video embed. I think If I  apply InsertEmbedsRule Rule when insert the embed. and must apply EnsureEmbedLineRule when delete the embed. If I do not apply InsertEmbedsRule, just don't need to apply EnsureEmbedLineRule.This is my idea.

singerdmx: Can you add one line to class comment:

/// This rule applies to video, not image.

singerdmx: Please update changelog. 
The change makes sense to me now.

TheHeBoy: Ok.

TheHeBoy: I can't understand when need to update changelog, Every time pr?

singerdmx: If this change needs to be published as a new version

PR #1320, opened 2023-07-21
Allow for custom context menu

Junoburger:
## Custom Context Menu 

For my app, I would like to customize the context menu that opens on text selection. 

I've been looking for a way to do it with the current implementation, but could not find a straight forward "clean" way. 

After exposing the `contextMenuBuilder` I was able to customize the menu.

editor

For which the implementation looks like this: 

```dart
QuillEditor(
  // .. params 
  textSelectionControls: CustomTextSelectionControls(),
  // enableSelectionToolbar: true,  // <-- Still hides the context menu
  contextMenuBuilder: (context, state) {
    return AdaptiveTextSelectionToolbar(
      anchors: state.contextMenuAnchors,
      children: [
        Container(
          // Container attributes here...
          child: Row(
            children: <Widget>[
              // (Icon) Buttons here
            ],
          ),
        ),
      ],
    );
  },
),

```

<hr>

And/Or give more control over the native context menu: 

<small>Like adding or removing buttons to it.</small>
editor2

```dart
contextMenuBuilder: (context, state) {
  List<ContextMenuButtonItem> btnItems =
      List.from(state.contextMenuButtonItems);

  btnItems.removeWhere(
    (item) => item.type == ContextMenuButtonType.selectAll
  );

  return AdaptiveTextSelectionToolbar.buttonItems(
    buttonItems: btnItems,
    anchors: state.contextMenuAnchors,
  );
},

```                                

<hr>

If there is a way to come to the first result, I'd be happy to learn about it, otherwise I think this would be a welcome addition. 

Please let me know if changes or additions are needed.

### DIFF
--- a/lib/src/widgets/editor.dart
+++ b/lib/src/widgets/editor.dart
@@ -191,6 +191,7 @@ class QuillEditor extends StatefulWidget {
     this.customLinkPrefixes = const <String>[],
     this.dialogTheme,
     this.contentInsertionConfiguration,
+    this.contextMenuBuilder,
     Key? key,
   }) : super(key: key);
 
@@ -432,6 +433,9 @@ class QuillEditor extends StatefulWidget {
   /// Configures the dialog theme.
   final QuillDialogTheme? dialogTheme;
 
+  // Allows for creating a custom context menu
+  final QuillEditorContextMenuBuilder? contextMenuBuilder;
+
   /// Configuration of handler for media content inserted via the system input
   /// method.
   ///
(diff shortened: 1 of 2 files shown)


singerdmx:
LGTM

Junoburger:
@singerdmx I've added some tests. 

Could make it simpler by just changing the name for one of the menu items and check for that. 
In any case, glad to receive feedback.

singerdmx:
Looks good!

fdaugs:
Hey @Junoburger. Thanks for you contribution. I've tried to open a PR just now. What a coincidence :D

PR #1362, opened 2023-08-17
Fix toolbar dividers condition

MilosKarakas:
There are only 6 groups of buttons but isButtonGroupShown has 10 members instead of 6. All conditions related to dividers access only members of isButtonGroupShown array with indexes 0-5. 

Alignment buttons should be grouped into one condition, which is what I adjusted and separators are showing as expected.


singerdmx:
info • The line length exceeds the 80-character limit • lib/src/widgets/toolbar.dart:178:81 • lines_longer_than_80_chars

singerdmx:
Please format

MilosKarakas:
@singerdmx  Sorry about that. Reformatted.

Issue #976, opened 2022-10-16
image pick is not supported in Web


MohammedSalehelShazly:
My issue is about Web platform

I have tried running `[flutter_quill example]` (https://github.com/singerdmx/flutter-quill/tree/master/example) 
The error occurred in **### getApplicationDocumentsDirectory**
```
  Future<String> onImagePickCallback(File file) async {
    // Copies the picked file from temporary cache to applications directory
    final appDocDir = await getApplicationDocumentsDirectory();
    final copiedFile =
        await file.copy('${appDocDir.path}/${basename(file.path)}');
    return copiedFile.path.toString();
  }
```

i'm user latest version of flutter and of the plugin
`flutter 3.3.4, flutter_quill: ^6.0.8+1`

any other solution?

true2thepen:
Still issue on flutter 3.3.7 and flutter_quill 6.1.3

LuchoTurtle:
This.
This won't ever work on the `web` because [`path_provider` ](https://pub.dev/packages/path_provider) does *not support* it. 
So, it fails when calling `getApplicationDocumentsDirectory()`.

Using [`universal_io`](https://pub.dev/packages/universal_io) might work to create a file from the array of bytes of the chosen image file.

hectorAguero:
I think this is a old issue, there's no need for universal_io, flutter_quill already has a `webImagePickImpl` property, also universal_io doesnt work for that. https://github.com/singerdmx/flutter-quill/blob/5dca6008cde707beb46291b6fbf71af6dbbf7bfb/example/lib/pages/home_page.dart#L272C29

Still, using the https://pub.dev/packages/cross_file will be better that ask for a webImplementation.

LuchoTurtle:
@hectorAguero sure, `cross_file` might be a better option. But uploading images does not work on the web when running the `example` app, even when `webImagePickImpl` is initialized.
This is because `onImagePickCallback` is called after the callback the code is executed in `webImagePickImpl`. So, the file that you return from `webImagePickImpl` will be passed on to `onImagePickCallback` 😞

freshtechtips:
Thank you for the report, this issue has been open for a while and since there are no interactions with it, we had to close it

but if you still facing this issue, or other issues or you have any other questions, feel free to contact us anytime you want.

PR #1367, opened 2023-08-22
refactor: Simplifying `universal_ui` in the example app

LuchoTurtle:
In the example app, in order to embed an image into the editor, we need to use `platformViewRegistry.registerViewFactory` so the image is properly displayed in `web` applications.

Because `platformViewRegistry.registerViewFactory(...)` was introduced in https://github.com/flutter/engine/pull/41877, we don't need to use a `fake_ui.dart` and `real_ui.dart` to do conditional imports as a workaround for that not compiling on `web`.

Therefore, we can simply use:

```dart
import 'dart:ui_web' as ui_web; // New import for all web-only stuff...

...

ui_web.platformViewRegistry.registerViewFactory(...); // as usual!
```

This PR makes this simple change and removes consequential dead code 😃 .

### DIFF
--- a/example/lib/universal_ui/fake_ui.dart
+++ b/example/lib/universal_ui/fake_ui.dart
@@ -1,3 +1,0 @@
-class PlatformViewRegistry {
-  static void registerViewFactory(String viewId, dynamic cb) {}
-}
--- a/example/lib/universal_ui/real_ui.dart
+++ b/example/lib/universal_ui/real_ui.dart
@@ -1,7 +1,0 @@
-import 'dart:ui' as ui;
-
-class PlatformViewRegistry {
-  static void registerViewFactory(String viewId, dynamic cb) {
-    ui.platformViewRegistry.registerViewFactory(viewId, cb);
-  }
-}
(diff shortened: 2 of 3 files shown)


LuchoTurtle:
Although embedding an image on `web` *still* doesn't work on the `example` app (as shown in https://github.com/singerdmx/flutter-quill/issues/976), I was able to verify this to work by returning an image URL in the `_onImagePickCallback` function inside `lib/pages/home_page.dart`.

https://github.com/singerdmx/flutter-quill/blob/5dca6008cde707beb46291b6fbf71af6dbbf7bfb/example/lib/pages/home_page.dart#L329-L335

Instead of running this, I'm returning a simple URL address, like so (again, just for testing purposes).

```dart
  Future<String> _onImagePickCallback(File file) async {
     return "https://i.ytimg.com/vi/CKU8la04tdo/mqdefault.jpg";
  }
```

image

LuchoTurtle:
Closing this as it doesn't  properly work on mobile devices :/ 

Sorry! You can dismiss this if you want to.

freshtechtips:
We have commented the universal_ui since it's already in the `flutter_quill_extensions`

PR #1379, opened 2023-08-28
fix: issue regarding onLiveTextInput being null

rynssnjn:
- Deleted onLiveTextInput parameter
- Format raw_editor.dart file


singerdmx:
Not accepted

singerdmx:
Please use older version of flutter quill since you are not using latest flutter

PR #1414, opened 2023-09-29
Code documentations and new event for confirm remove the image

ahmedhnewa:
Hi, I have already sent a pull request

[Previous pull request](https://github.com/singerdmx/flutter-quill/pull/1413)
[Previous Feature request](https://github.com/singerdmx/flutter-quill/issues/1412)

But I still made new changes

- Documented the code of flutter_quill_extensions
- Fix Arabic language translations
- Added a new event to allow the developer to confirm or not confirm removing the image which is necessary for production in my opinion

I should have organized the changes in a better way and only in one pull request

but even if you decide to reject the changes, I hope you will still document the code using comments
fix Arabic translations
and add more events and translations to pass

Remove unused plugins
as well as unnecessary packages ones 
(like `string_validator` and `url_launcher`) and update

all the packages and plugins

and continue to improve this package

Again, thank you for your efforts and time

PR #1413, opened 2023-09-29
Improvements and one new event

EchoEllet:
Solve [#1412](https://github.com/singerdmx/flutter-quill/issues/1412)

Some minor changes.

PR #1417, opened 2023-10-02
Fix my issues and changes

ahmedhnewa:
I have used gal instead of image_gallery_saver

I didn't change the packages and plugin versions

Please change it to fit your needs

[Code documentation and a new event for confirm removal of the image](https://github.com/singerdmx/flutter-quill/pull/1414)
[Improvements and one new event](https://github.com/singerdmx/flutter-quill/pull/1413)

### DIFF
--- a/flutter_quill_extensions/lib/flutter_quill_extensions.dart
+++ b/flutter_quill_extensions/lib/flutter_quill_extensions.dart
@@ -82,7 +82,8 @@ class FlutterQuillEmbeds {
               },
           shouldRemoveImageFromEditor: shouldRemoveImageFromEditor ??
               (imageFile) {
-                // TODO: Before pubish the changes, please consider change the name
+                // TODO: Before pubish the changes
+                // please consider change the name
                 // of the events if you want to
                 return Future.value(true);
               },
--- a/flutter_quill_extensions/lib/flutter_quill_extensions.dart
+++ b/flutter_quill_extensions/lib/flutter_quill_extensions.dart
@@ -20,11 +20,36 @@ export 'embeds/toolbar/video_button.dart';
 export 'embeds/utils.dart';
 
 class FlutterQuillEmbeds {
+  /// Returns a list of embed builders for Quill editors.
+  ///
+  /// [onVideoInit] is called when a video is initialized.
+  /// [onRemoveImage] is called when an image is removed from the editor.
+  /// By default, [onRemoveImage] deletes the cached image if it still exists.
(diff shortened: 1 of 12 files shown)


singerdmx:
You need to rebase

singerdmx:
@Jon-Salmon please have a look

flutter_quill_extensions/lib/flutter_quill_extensions.dart
```diff
@@ -72,21 +72,8 @@ class FlutterQuillEmbeds {
   }) =>
       [
         ImageEmbedBuilder(
-          afterRemoveImageFromEditor: afterRemoveImageFromEditor ??
-              (imageFile) async {
```
Jon-Salmon, line 76: I was mainly referring only to `shouldRemoveImageFromEdtior` about it being optional. There is nothing wrong with this being optional too, but we should still provide this sensible default here.

Jon-Salmon (changes requested):
Thanks for the updates! Just one small nit, but happy for this to be merged after changing that one small bit.

ahmedhnewa:
> Thanks for the updates! Just one small nit, but happy for this to be merged after changing that one small bit.

Thank you for your concern about this great plugin, I think we need to change the documentation of some of the changed code

I mentioned that the image by default will be removed from the local storage if it is removed from the editor from Quill Delta

But now it will not do anything by default

Instead, in the documentation, we should only show the developer who uses this package how to do it himself since it's not by default

I also suggest changing the names of the parameters that I have added

I mean, it could be shorter and more descriptive, but I was lazy, to be honest, since I already mentioned that I didn't have enough time to make more changes

If you want me to make these changes by myself, please tell me to do it

Or if you can handle it and you think you will do it better, then be my guest.

Thank you for your time and efforts again to the community.

flutter_quill_extensions/lib/flutter_quill_extensions.dart
ahmedhnewa: > I was mainly referring only to `shouldRemoveImageFromEdtior` about it being optional. There is nothing wrong with this being optional too, but we should still provide this sensible default here.

I agree with you, but then we shouldn't change the documentation code

As I suggest in my previous comment

And I was mainly refering to afterRemoveImageFromEditor

singerdmx:
LGTM

Jon-Salmon:
Thanks for the contributions. I'll make some documentation changes and release a new version shortly

PR #1403, opened 2023-09-19
flutter_quill_extensions: replace gallery_saver with gal

santiagos01:
Resolves https://github.com/singerdmx/flutter-quill/issues/1377

PR #1424, opened 2023-10-16
Pruning unused translations and adding relevant ones

LuchoTurtle:
I was initially going to work on finishing https://github.com/singerdmx/flutter-quill/pull/1403#issuecomment-1737079301 but since https://github.com/singerdmx/flutter-quill/pull/1414 has already been merged, I realised that some translations were needed and other weren't used. So I've refactored it!

In addition, I've added a warning in the `README` file regarding https://github.com/natsuk4ze/gal, as some permissions need to be added to some files (as noted in https://github.com/singerdmx/flutter-quill/pull/1403#discussion_r1335527280).

This should be a quick merge 👌


freshtechtips:
Thanks

freshtechtips:
@singerdmx I made the message clearer to know if we are saving the image using network or from local storage (cached) just for testing purposes, do we need to have separate translations, one for network saved image and another for storage?

LuchoTurtle:
Yeah, I was a bit hesitant about having specific translations for each scenario, as it doesn't seem really useful to the end user. 
But I kept it because I didn't want to meddle too much with the pre-existing code.

freshtechtips:
> Yeah, I was a bit hesitant about having specific translations for each scenario, as it doesn't seem really useful to the end user. But I kept it because I didn't want to meddle too much with the pre-existing code.

I thought the team would probably change the pull request code but I kept it like this so they could test it directly in the UI

so I remind them to change it if they want to

but it might be helpful to the end user to know if this image is from the cloud, by image URL, or locally in the sandboxed directory of the app

PR #1425, opened 2023-10-16
added the 'Swahili' locale  translation to toolbar.i18n.dart

SichangiFelix:
Locale = Swahili
Language code = sw


lib/src/translations/toolbar.i18n.dart
```diff
@@ -2105,6 +2105,74 @@ extension Localization on String {
           'Saved using the local storage': 'Saved using the local storage',
           'Error while saving image': 'Error while saving image',
         },
+          'sw': {
```
singerdmx, line 2108: indentation

Issue #1427, opened 2023-10-17
Provide a way to handle the errors of the image widget and the other widgets


freshtechtips:
Hi, in flutter_quill_extensions I don't think there is a way to customize the image widget

This issue is about the Flutter framework and it's not for a specific platform (like iOS or web etc.)

for example, use the CachedNetworkImage widget or custom image provider 
or at least handle the errors if the image doesn't exist, if there is no internet connection etc...

example:

```dart
Image.network(
   imageURL,
   errorBuilder: (context, error, stackTrace) {
     return Text('This is my custom error widget');
  },
)

```
I'm not sure but I think the same applies to the video widget and the others, or the errors in general

again I could do it myself but I messed a lot of things up in my last three pull requests,
the commit history, the branch names, the code, and the changes and it requires you guys to do a lot of reviews to make sure everything doesn't conflict with other codes, my problem was I didn't fetch the latest changes from the original repository and merge the changes and solve the conflicts and squash the commit message property, also some of the requirements were not very clear and it takes a long time and delay from the both sides

but if you don't mind and no one wants to do it then I will start fresh after you guys accept my last pull request with the changes and I will do it clean this time

again thank you for your time and efforts to the open-source community and the community in general.

freshtechtips:
> Yeah go ahead Why would the image not exist? But you are right we should have error handing Thanks

If the user enters a valid image URL but the image doesn't exist on the network,
or we upload the image to the network and replace it but the image got deleted somehow by the cloud

if you want me
 please merge my pull request or reject it if you want, so I can start fresh by fork it again because my forked repository is a little messed up so I can safely delete it

I will fix the conflict just give me a few minutes

No, thank you for this open-source library and your support to the community

freshtechtips:
This is no longer a issue

PR #1428, opened 2023-10-17
Bug fixes + Provide a way to handle the image errors and use custom image provider

freshtechtips:
Hi, I made a brand-new changes
but it's simple and not as big as the previous one

Features:

1. Allow to use custom image provider
2. Provide a way to handle all the errors of the image widget everywhere even in the preview
3. Add support for image resizing on desktop platforms when forced using the mobile context menu

My purpose of this pul is to add features but I made some improvements in my way.

Issue [#1427](https://github.com/singerdmx/flutter-quill/issues/1427)

Improvements:

- Little code orgnazation
- Fix two bugs, the first one when we use
  ```dart
   mediaPickSettingSelector: (context) async {
                    return MediaPickSetting.Camera;
                  },
  ```
  it will always use a pick image by link instead, this is why I prefer using a switch statement instead of if-else since we add more options all the time and we forgot to handle the others

- The second bug that will happen when we **showCameraButton** to true and we set **onImagePickCallback** but not **onVideoPickCallback**, then nothing will happen, some developers are not interested in using the camera for both photos and recording videos since they are scenarios when you just want to create simple note app or blog for example and you don't want videos, so throw an error will not be a solution, but the current behavior is not throwing an error so I change it to show only take a photo by the camera option if it set and take a photo by the video if set and the dev can always override this by **mediaPickSettingSelector** and **cameraPickSettingSelector**
- The way to check HTTP URL for some functions is handled in a more efficient way
- Improve performance by reducing the number of widgets rebuilt by listening to media query for only the needed things, for example instead of using `MediaQuery.of(context).size`, now we are using `MediaQuery.sizeOf(context)`
- New custom style attributes for platforms other than mobile

```json
[{"insert":{"image":"/Users/ahmedhnewa/Downloads/alrayada.net.png"},"attributes":{"style":"width: 704.8; height: 600.0; "}},{"insert":"\n"}]
```

Suggestions:
- in the flutter_quill_extensions.dart in line 130 in the function List<EmbedBuilder> builders
we have quite a lot of configuration for the image, how about making a class that hold all the configurations for each one, for the image, and all the different widgets
of course, this will require the developers that are overriding the property to update the code of the usage a little bit

- Please read all the todos, most of them I can't decide on my own because that is a community project, and some of them are in flutter_quill_extensions that require changes from flutter_quill but we need to update flutter_quill first so we can get the changes from flutter_quill_extensions, most of them are new custom style attributes


singerdmx (approved):

singerdmx:
@Jon-Salmon please have a look

freshtechtips:
> @Jon-Salmon please have a look

I might make more changes, so you might want to consider keeping this pull request open and not accepting the changes just yet, if possible.

singerdmx:
Ok

freshtechtips:
> Ok

I'm adding a new little feature to the desktop which is image resizer support for desktop
but to save some time, can you tell me how to change the width and height of the quill element using the quill controller??

Currently, this logic works only for Mobile:

```dart
final res = getEmbedNode(
                              controller,
                              controller.selection.start,
                            );
                            if (base.isMobile()) {
                              final attr = base.replaceStyleString(
                                getImageStyleString(controller),
                                w,
                                h,
                              );
                              controller
                                ..skipRequestKeyboard = true
                                ..formatText(
                                  res.offset,
                                  1,
                                  StyleAttribute(attr),
                                );
                              return;
                            }
``` 

I tried and I always get an exception of type String which is 'Apply delta rules failed' which is in

```dart
Delta apply(
    RuleType ruleType,
    Document document,
    int index, {
    int? len,
    Object? data,
    Attribute? attribute,
  }) {
    final delta = document.toDelta();
    for (final rule in _customRules + _rules) {
      if (rule.type != ruleType) {
        continue;
      }
      try {
        final result = rule.apply(delta, index,
            len: len, data: data, attribute: attribute);
        if (result != null) {
          return result..trim();
        }
      } catch (e) {
        rethrow;
      }
    }
    throw Exception('Apply delta rules failed');
  }
```

It's a matter of time until I manage this but to help save some time and make this pull request as closed as possible
if possible, can you tell me how to do it in more detail?? the code of flutter_quill is not documented with comments so I need more time

singerdmx:
We can add more logging to help debugging

singerdmx:
This indicates no rule is applied at all

singerdmx:
Apply delta rules failed means no rule is applied

freshtechtips:
> Apply delta rules failed means no rule is applied

I think I'm able to add this little feature

I just need you to tell me if this is a valid quill delta. 

```json
[{insert: {image: /Users/ahmedhnewa/Downloads/example.png}, attributes: {style: width: 412.0; height: 600.0; }}, {insert:
}]
```

It was style: `mobileWidth` and mobileHeight and I'm not sure if this applies to the original quill JSON rules but is it okay to store anything in the attributes of the `mobileWidth` and mobileHeight are not official and extra from this flutter library

but now on desktop, I'm using different attributes which is the width and height key

singerdmx:
mobileWidth is our newly defined special key which is not part of the standard quill delta

freshtechtips:
> mobileWidth is our newly defined special key which is not part of the standard quill delta

1- The attributes key, is it official and part of quill delta??
2- Is it okay to define more properties (width, height)

freshtechtips:
> Desktop should follow web I guess

Not really, The web uses  tag, and desktop can't do that without Chromium or webview

web uses the same one on mobile but as I said previously the context menu of macOS was quite buggy so I added an option for the developer to force the use mobile context menu on desktop, however the resizing option was not working so I had to add it only for mobile

in this pull request, I decided to add support for that too
but we were using mobileWidth and mobileHeight for mobile, what about desktop?? should we use something different like normal width and height like the JSON I sent in my previous message

I'm adding support for desktop context menu resizing image
so it will use properties (width, height) instead of mobile ones and this
is only for desktop and doesn't apply for mobile

singerdmx:
Ok make sense

freshtechtips:
@singerdmx 
@Jon-Salmon 

I know you guys are busy and I have sent pull requests and requested re-reviews too much but you might want to consider re-review the changes again just this time if possible because there are changes made by the others that conflict with this pull request and I had to solve conflicts and commit things to make this project the best it can be with the help of the community, I mean I could squash them but if we're able to get this pull request merged or rejected faster then I will be able to work more on new changes because I do plan to send more pull requests if you don't mind

singerdmx:
LGTM

singerdmx:
@Jon-Salmon I will go ahead and merge this. You can leave comments when you have time so it will be addressed in next PR

PR #1444, opened 2023-10-20
The first steps of the major update

freshtechtips:
# Pull Request

## Description

This is just the first steps of the major update, but it's not a big deal and you can publish it if you want it since the only change that is required is to remove the controller of `QuillEditor` and `QuillToolbar` and wrap both

in the following way

```dart
QuillProvider(
  configurations: QuillConfigurations(controller: _controller),
  child: Column(
  children: [
    QuillToolbar.basic(controller: _controller),
    Expanded(
      child: Container(
        child: QuillEditor.basic(
          readOnly: false, // true for view only mode
        ),
      ),
    )
  ],
),
)
```

instead of

```dart
Column(
  children: [
    QuillToolbar.basic(controller: _controller),
    Expanded(
      child: Container(
        child: QuillEditor.basic(
          controller: _controller,
          readOnly: false, // 为 true 时只读
        ),
      ),
    )
  ],
)
```

Example of how it will look like

```dart
   configurations: QuillConfigurations(
            controller: _controller,
            sharedConfigurations: const QuillSharedConfigurations(
              dialogBarrierColor: Colors.black26,
            ),
            editorConfigurations: const QuillEditorConfigurations(),
            toolbarConfigurations: const QuillToolbarConfigurations(),
          ),
```

I already have updated the `README.md` and `CHANGELOG.md` and the `pubspec.yaml`
 but I didn't update the Chinese one
I also updated
the tests, and the example

even though I noticed that the example has some issues on the macOS platform

please read the additional notes

## Improvements

- Improve code readability
- Improve the way to pass configurations (not complete, just the `controller` and `dialogBarrierColor` as an example)
- Fix the macOS example will not build and work at all, I have to increase the minimum deployment since `gal` requires increasing it
- Remove Multidex as it's not required in minSdkVersion API level 21, I think the maintainer of this library already knows what I'm talking about since he was a native Android developer in the past
- Rename the Quill text class name to `QuillText` to avoid annoying conflict with the material library and others
- Improve the example by cleaning the code and using more safer way to load the default quill document and other improvements
- Update the test by wrapping the test widgets where we use `QuillEditor` and `QuillToolbar`
- More changes and little improvements

This is just the showcase and not even the start of the major changes
I just showcase how it will look by moving two properties (`controller` and `dialogBarrierColor`) to a better place

## Additional notes
Again, you can merge this pull request into the `master` branch and I will move the rest of the configurations after that
you might wonder why I sent this pull request if I just started in step 1. For some reasons

1. I want to show you the showcase of the new changes and the new way to pass things and get them
2. To avoid some conflicts and the next pull requests features they will know where they should add configurations so I don't have to fix them all which requires too much time I would rather spend fixing bugs and adding more features and improvements
3. I want to know if the community is interested in this change

If you find these changes but don't want to merge them for some reason even though I prefer to merge them
You can cancel this pull request but tell me if it is OK or not so we start work on the next changes

if you merge them and publish the library, not much is required to change from the developers who using this library, this moves the controller to one place (they still can use different controllers for the toolbar and editor) but what we will benefit from that the new features that require more configurations the developers will know where to put them and we don't have to move things and solve conflicts in the future

## Checklist

- [x] I have added/updated relevant documentation 
- [x] I have tested these changes locally. 
- [x] I have updated the test to works 
- [x] I have followed the code style and guidelines. 
- [x] I have updated `CHANGELOG.md` with my changes in the next section 
- [x] I have run "dart format ." on the project 
- [x] I have run `flutter test` and `flutter analyze` and it passed successfully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# [7.4.17]
+# [7.5.0]
 
 - **Breaking change**: The widgets `QuillEditor` and `QuillToolbar` are no longer have controller parameter, instead you need to make sure in the widget tree you have wrapped them with `QuillProvider` widget and provide the controller and the require configurations
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# [7.4.17]
+
+- **Breaking change**: The widgets `QuillEditor` and `QuillToolbar` are no longer have controller parameter, instead you need to make sure in the widget tree you have wrapped them with `QuillProvider` widget and provide the controller and the require configurations
+
 # [7.4.16]
 
 - Update documentation and README.md
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -40,5 +40,6 @@ Closes #IssueNumber
 - [ ] I have tested these changes locally. <!-- REQUIRED -->
 - [ ] I have followed the code style and guidelines. <!-- REQUIRED -->
 - [ ] I have updated `CHANGELOG.md` with my changes in the next section <!-- REQUIRED -->
-- [ ] I have run "dart format ." on the project <!-- REQUIRED -->
+- [ ] I have run `dart format .`` on the project <!-- REQUIRED -->
+- [ ] I have run `dart fix --apply` on the project <!-- REQUIRED -->
 - [ ] I have run `flutter test` and `flutter analyze` and it passed successfully <!-- REQUIRED -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,18 +11,18 @@ Closes #IssueNumber
 (Replace "IssueNumber" with the actual issue number you are addressing.)
 
 ## Improvements
-<!-- Please tell us the improvemenets you made in a list -->
+<!-- Please tell us the improvements you made in a list -->
 
 <!-- Example: -->
 - Improve code readability
-- Improve peformance
+- Improve performance
 
 ## Features
 <!-- Please tell us the features you added in a list if you add any -->
 
 <!-- Example: -->
 - Add a new feature
-- Allow to custmize the widgets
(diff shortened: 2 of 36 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# [7.4.17]
```
singerdmx, line 1: Can we make it 8.0.0 or 7.5.0 since it is breaking change

freshtechtips: > Can we make it 8.0.0 or 7.5.0 since it is breaking change

We will make more and more changes so if we keep it that way we will reach 100.0.0 in no time
Developers will get an error that the controller is not defined and they can't define it, they will probably go the change log and look at the different and will easily change that in less than minute

if you still want to make it 8.0.0 or 7.5.0 tell me

singerdmx: at least 7.5.0

singerdmx: I don't even see 100.0.0 is an issue

singerdmx (approved):

CHANGELOG.md
freshtechtips: > at least 7.5.0

Done, you are okay with the first changes?? if yes then I will make the second pull request which will complete most of the changes and we will change the version to 8.0.0 this way we will not increase it too much in a short time

if no, consider telling me what could be better.

singerdmx: Ok ready to merge this?

freshtechtips: > Ok ready to merge this?

If you are then yes

freshtechtips: > Ok ready to merge this?

I just added the supported platforms to **flutter_quill** **flutter_quill_extensions**

flutter_quill supports all even though it doesn't have the best experience for the web because of the flutter engine (I might need to check the code deeper so I will make sure of it more)

I marked flutter_quill_extensions as support for all except for Linux since [gal](https://pub.dev/packages/gal) doesn't support Linux

In the future. I might go there and add support by sending a pull request to them since I don't think there is a plugin that supports saving images for Linux

I think that should be the last update for this pull request

PR #1454, opened 2023-10-22
Major update step 4

freshtechtips:
Managed to refactor most of the button configurations and move them to the new `QuillProvider` which will allow to customize every single button widget (except the Select Alignment buttons but we will work on that) and there are just three buttons left to refactor, the headline, search and insert link buttons

then we will refactor the `QuillToolbar` and then the editor which is the hardest part but we will manage

@singerdmx Feel free to merge this pull request but it's not necessary to publish yet, I will start on the others tomorrow so it's really not necessary to release two versions in a short time

and it's doesn't have much bug fixes, just improvements

I also added experimental animation support and it's disabled by default
and I just added it to the checkbox list for now it has one single issue that is not related to the animations or the changes we made but I will mention that later, I also marked it as experimental so developers know that when enabling it

which is the list of the checkboxes that will rebuild most of the items when there is a change
and it will cause the animation to replay for all the other items when there is a new change

a simple solution would be using the official animated lazy list from the Flutter framework but we will work on that later

let's fix more issues and add further improvements then we will release all in one go instead of too many releases
it won't take long just a few days

I also updated the and added quick instructions to use the library directly using git instead of the pub.dev so no need to publish the release, developers will be able to test this without publishing this release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## [7.8.0]
+- **Important note**: this is not test release yet, it works but need more test and changes and breaking changes, we don't have development version and it will help us if you try the latest version and report the issues in Github but if you want a stable version please use `7.4.16`. this refactoring process will not take long and should be done less than three weeks with the testing.
+- We managed to refactor most of the buttons configurations and customizations in the `QuillProvider`, only three lefts then will start on refactoring the toolbar configurations
+- Code improvemenets
+
 ## [7.7.0]
 
 - **Breaking change**: We have mirgrated more buttons in the toolbar configurations, you can do change them in the `QuillProvider`
--- a/README.md
+++ b/README.md
@@ -42,6 +42,37 @@ Pub: [FlutterQuill]
 
 ---
 
+## Installation
+
+```yaml
+dependencies:
+  flutter_quill: ^<latest-version-here>
+```
+
+<p align="center">OR</p>
+
+```yaml
+dependencies:
+  flutter_quill:
+    git: https://github.com/singerdmx/flutter-quill.git
+```
+
+> **Important note**
+>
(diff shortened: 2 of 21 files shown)


freshtechtips:
image
I don't have any analysis errors but to fix those we probably need to update `flutter_quill_extensions` in pub.dev first

singerdmx:
There is CI failure

singerdmx:
@freshtechtips 
Do you want to publish flutter_quill 7.8.0 to https://pub.dev (y/N)? y
Uploading... (2.0s)
Invalid `topics` value (`rich text editor`): must consist of lowercase alphanumerical characters or dash (but no double dash), starting with a-z and ending with a-z or 0-9.

freshtechtips:
> @freshtechtips 
> Do you want to publish flutter_quill 7.8.0 to https://pub.dev (y/N)? y
> Uploading... (2.0s)
> Invalid `topics` value (`rich text editor`): must consist of lowercase alphanumerical characters or dash (but no double dash), starting with a-z and ending with a-z or 0-9.
> 

I will remove it tomorrow  "rich text editor" topic item in pubspec.yanl


Because I'm in the middle of a dream right now

It's stringe that --dry-run success

singerdmx:
already fixed

freshtechtips:
> > @freshtechtips 
> > Do you want to publish flutter_quill 7.8.0 to https://pub.dev (y/N)? y
> > Uploading... (2.0s)
> > Invalid `topics` value (`rich text editor`): must consist of lowercase alphanumerical characters or dash (but no double dash), starting with a-z and ending with a-z or 0-9.
> > 
> 
> I will remove it tomorrow  "rich text editor" topic item in pubspec.yanl
> 
> 
> Because I'm in the middle of a dream right now
> 
> It's stringe that flutter pub publish --dry-run succeed

Issue #1288, opened 2023-07-01
[Mobile] or [Desktop]: app craches after inserting URL


Alspb:
App crashes after inserting URL (failed assertion).
Steps to reproduce:

1. Run the example (from `example` directory)
2. Place the cursor right after "diff-match-patch"
3. Press "Insert URL" button, enter text and URL and press "OK"

freshtechtips:
Hi, thank you for the report, this issue has been fixed, if you still facing any other issues or have any other questions feel free to ask us on GitHub

PR #1461, opened 2023-10-23
Major update 6

freshtechtips:
- The Quill editor and toolbar configurations are now refactored in one single class for each one
- After changing one of the checkbox list values the controller will not request the keyboard focus by default
- We have moved the configurations of the toolbar and the editor directly into the widget but we are still using inherited widgets internally
- Fixes to some of the code after the refactoring

More changes...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## [7.10.0]
 - **Breaking change**: `QuillToolbar.basic()` can be accessed from `QuillToolbar()` directly and the old `QuillToolbar` can be accessed from `QuillBaseToolbar`
-- The Quill editor and toolbar configurations is now refactord in one single single class for each one
-- After changing one of the check box list values the controller will not request the keyboard focus by default
-- We have moved the the configurations of the toolbar and the editor directly into the widget but we still using inheritied widgets internally
+- The Quill editor and toolbar configurations are now refactored in one single class for each one
+- After changing one of the checkbox list values the controller will not request the keyboard focus by default
+- We have moved the configurations of the toolbar and the editor directly into the widget but we are still using inherited widgets internally
 - Fixes to some of the code after the refactoring
 
 ## [7.9.0]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [7.10.0]
+- **Breaking change**: `QuillToolbar.basic()` can be accessed from `QuillToolbar()` directly and the old `QuillToolbar` can be accessed from `QuillBaseToolbar`
+- The Quill editor and toolbar configurations is now refactord in one single single class for each one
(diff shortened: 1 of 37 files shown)


freshtechtips:
**flutter_quill_extensions** will fail again, we probably need to publish the package to pub.dev and then update things from **flutter_quill_extensions**, also we should publish the flutter_quill_extensions updates to pub.dev

singerdmx:
Another option is that you publish the package instead

freshtechtips:
> Another option is that you publish the package instead

Are we talking about flutter_quill_extensions?? because you told me @Jon-Salmon has access to the publishing

singerdmx:
Yes I mean if you don't want to be blocked on that. Create a new package

freshtechtips:
> Yes I mean if you don't want to be blocked on that. Create a new package

Do you mean fork the repository of flutter_quill_extensions??

singerdmx:
Oh pub.dev check url uniqueness? I see. Probably have to work around it

freshtechtips:
> Oh pub.dev check url uniqueness? I see. Probably have to work around it

Still not sure what you mean by this message, is this what you mean by taking a copy of the flutter_quill_extensions publishing it with the new changes, and naming it with a different package name like flutter_quill_extensions_plus or something like that?? just

singerdmx:
Yes

freshtechtips:
> Yes

Fine by me but I think we should tell @Jon-Salmon first about this first
or another option is that we do this temporarily while he/she is not around

singerdmx:
Sure

freshtechtips:
> Sure



> Can't run the example - it shows failed assertion.

I didn't test the example because I decided to test the changes on a real project instead with extra functionality to catch more bugs

but give me a moments to check it...

freshtechtips:
> Can't run the example - it shows failed assertion.



> Can't run the example - it shows failed assertion.

Please clean the project and run it again and make sure to fetch the latest changes and rebase or merge

image

Can't encounter the issue you are facing, consider giving me more details so I can fix it

freshtechtips:
> Hm, it also works for me on Android emulator. But on Linux it fails with pretty large stack trace. If you don't face the error on desktop, I'll put the details here.

if it is related to the Linux binary application building. then I can't fix it right now since this library is not supported for Linux yet as it uses plugins that don't support Linux yet. I will look into that soon but it would be helpful to share more details

freshtechtips:
> Please also take a look at #1288, cause it seems to be connected with images which you're also working on.

Sure, I will checkout tomorrow. But can you upload some images or short video??

freshtechtips:
> > > Please also take a look at #1288, cause it seems to be connected with images which you're also working on.
> > 
> > Sure, I will checkout tomorrow. But can you upload some images or short video??
> Please find the video below:
> 
> insert_link_crash
> 
> 

Thank you for the report

As soon we finish the refactoring and other issues we will work on this one

PR #1470, opened 2023-10-24
Removed line numbers in code block

Alspb:
# Pull Request

## Description

Removes line number from the code block.

## Issues

Fixes the following problems:

- Code block line numbers interferes with ordered lists. For instance, in the example "Have trouble finding things?" line has No. 2 instead of 1 because of the code block above.
- Text in code block override line numbers anyway.

## Improvements

Removing line numbers also saves some space on mobile screens.
Finally, code blocks simply don't need line numbers, it's not an IDE.

## Checklist

- [ ] I have added/updated relevant documentation 
- [ ] I have tested these changes locally. 
- [ ] I have followed the code style and guidelines. 
- [ ] I have updated `CHANGELOG.md` with my changes in the next section 
- [ ] I have run `dart format .`` on the project 
- [ ] I have run `dart fix --apply` on the project 
- [ ] I have run `flutter test` and `flutter analyze` and it passed successfully 
- [ ] I have run `./before-push.sh` and everything is fine

PR #1471, opened 2023-10-25
Disable code line number in code block by default

freshtechtips:
Add a configuration class for the blocks in the editor

one of them is the code block and in there you will find enableLineNumbers and it's by default false

This solution thanks to @Alspb in his/her pull request [#1470](https://github.com/singerdmx/flutter-quill/pull/1470)


freshtechtips:
@singerdmx 
Can everything be considered a block?? because I'm adding a class configuration which will be called block options and inside it, we will find things like code block options and in there we will configure things

like enable line numbers

if the answer is yes then feel free to merge this pull request

singerdmx:
For the least, block vs inline means block at least occupy one line conceptually

singerdmx:
is this change ready to merge

freshtechtips:
> For the least, block vs inline means block at least occupy one line conceptually

So the items like custom embed builders, images, code block, checkbox etc..
can be considered as a block??

singerdmx:
why is checkbox a block?

singerdmx:
block is supposed to consist multiple lines (at least one)

freshtechtips:
> why is checkbox a block?

You are right this is not the name. give me a name that represent the items like images, checkbox, code block etc..

singerdmx:
Have you checked this https://www.youtube.com/watch?v=r502AfCX-Hw&t=284s ?

singerdmx:
I would ask chatgpt by describing my need. It is better with names

singerdmx:
I don't think checkbox should be treated the same way as block

freshtechtips:
> I would ask chatgpt by describing my need. It is better with names

it's blocked in my area so I can't use it, please use it instead and told him about this to get the perfect name

freshtechtips:
> I don't think checkbox should be treated the same way as block

Since we don't have the perfect name then should we call it something like
item, widget, view, or anything else??

singerdmx:
I am confused why is checkbox on the table?

freshtechtips:
> I am confused why is checkbox on the table?

I want to add a class configurations for views like checkbox, code block

to customize their behavior
for example, should we enable the line numbers in the code block??
should the checkbox have animations??

I want to follow the same approach for all the refactored code

Example of what it will be:
First the QuillEditorConfigurations
then

QuillEditorViewsConfigurations (or whatever that is called, this is why I'm asking)

- QuillEditorCheckboxConfigurations
   - animationEnabled

- QuillEditorCodeBlockConfigurations
  - enableLineNumbers

and so on

singerdmx:
If you're simply looking for a generic and versatile name to represent items like images, checkboxes, code blocks, etc., you can use "Widget" or "Element." These terms are broad and do not imply a specific type, making them suitable for a variety of components in your application.

singerdmx:
here are a few more generic terms you can consider:

Component
Module
Unit
Object
Item
Entity
Part
Element
Block
View

freshtechtips:
> If you're simply looking for a generic and versatile name to represent items like images, checkboxes, code blocks, etc., you can use "Widget" or "Element." These terms are broad and do not imply a specific type, making them suitable for a variety of components in your application.

Thanks, let me commit one final change, and let's merge this...

singerdmx:
if you're looking for terms more specific to the field of UI and design, you can consider:

UI Element
User Interface Object
Interface Component
UI Block
Display Object
User Control
Interface Item
Visual Component
Screen Element
User Interface Entity
These terms emphasize their connection to the user interface, making them more suitable for UI components in your application.

freshtechtips:
> if you're looking for terms more specific to the field of UI and design, you can consider:
> 
> UI Element User Interface Object Interface Component UI Block Display Object User Control Interface Item Visual Component Screen Element User Interface Entity These terms emphasize their connection to the user interface, making them more suitable for UI components in your application.

I think we are ready now to merge this

PR #1486, opened 2023-10-28
Code cleanup and fixes

freshtechtips:
This pull request is not ready yet to be merged but it will be soon

The biggest changes are:
1. Require Dart version 3 as a minimum (migrate to it as well)
2. Fix the bug of the link reg exp, since it was always using the default instead of the custom one
3. Code cleanup


singerdmx:
Hmm why automated publishing is not working

freshtechtips:
> Hmm why automated publishing is not working

It needs more changes, I can test it more update it, and search for it more, I don't prefer using automated publishing as we should not publish many versions with every single version, I don't mind but users don't like this

let me know if you are still interested, I will test this on `fresh_quill_extensions` and when it works in efficient way I will bring it here

singerdmx:
It is not totally automated.
You need to create a release in order to create a git tag and then is the automated publishing

PR #1496, opened 2023-11-01
Use screenshots section instead of using the images in README.md

freshtechtips:
Use screenshots section instead of using the images in README.md

the images will only shown in pub.dev in the package details and not in Github README


freshtechtips:
@singerdmx Consider publishing this changes with new minor version so we can see how it looks in pub.dev package details to make sure we didn't make any mistakes, even though the --dry-run and flutter pub get already take care of this but just in case

singerdmx:
Published

freshtechtips:
> Published

The screenshots is now visible, take a look at [flutter_quill](https://pub.dev/packages/flutter_quill) if you want to see it

PR #1518, opened 2023-11-10
Fix platform checking logic

freshtechtips:
I think we have messed up some of the code base but it was not working properly as it should

Let me make it simple, in flutter we can check the current platform using dart:ui which will works on all platforms except web since it working differently, however the flutter package include a solution to that, it will get the platform even if it’s on web

```dart
import 'package:flutter/foundation.dart'
    show kIsWeb, TargetPlatform, defaultTargetPlatform;
```

So we need to be more specific if we weren’t to check the current working app platform or the platform that is using a browser to run the web app

Feel free to share your thoughts about this

**Note**:
The parameter `supportWeb` might need to be renamed, share a suggestion before we merge the code

There is also a failure but it will be fixed once we publish the `flutter_quill` and `flutter_quill_extensions`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - **Breaking change**: the `customButtons` in the `QuillToolbarConfigurations` is now of type `List<QuillToolbarCustomButtonOptions>`
 - Bug fixes with the new `8.0.0` update
 - Update `README.md`
+- Improve the platform checking
 
 ## [8.3.0]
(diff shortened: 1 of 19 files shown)


freshtechtips:
@singerdmx
I merged this as we have more changes to `flutter_quill_extensions` and we want to focus on it a little

But let me know if we should revert or do any changes

singerdmx:
ok

singerdmx:
I need to publish right?

freshtechtips:
> I need to publish right?
> 

Sure, but the extension package is not ready yet for publishing

Issue #1100, opened 2023-02-14
[Request] Please add raw html string to initialize at controller


paraslamsong:
Faced the problem to use html string but the package is not allowing to use html string as input.

paraslamsong:
For now the solution is
```

 _controller.document = quill.Document.fromJson(
        jsonDecode(markdownToDelta(html2md.convert(htmlString))),
      );
```

With the use of other packages, 
[html2md](https://pub.dev/packages/html2md)
and 
[delta_markdown](https://pub.dev/packages/delta_markdown)

mmasdivins:
The problem with converting HTML to Markdown is that we lose all the following properties:
- Background Color
- Color
- Font
- Size
- Strikethrough
- Superscript/Subscript
- Underline
- Text Alignment
- Text Direction
- Formula
- Video

baoletrangia:
You can write your own converter from HTML to Delta. It works for me

YanusScevola:
You can use my **html to Delta** converter i wrote.

```kotlin

class HtmlToDeltaConverter {
  static const _COLOR_PATTERN = r'color: rgb\((\d+), (\d+), (\d+)\);';

  static quill.Delta _parseInlineStyles(htmlDom.Element element) {
    var delta = quill.Delta();

    for (final node in element.nodes) {
      final attributes = _parseElementStyles(element);

      if (node is htmlDom.Text) {
        delta.insert(node.text, attributes);
      }else if (node is htmlDom.Element && node.localName == 'img') {
        final src = node.attributes['src'];
        if (src != null) {
          delta..insert({'image': src});
        }
      } else if (node is htmlDom.Element) {
        delta = delta.concat(_parseInlineStyles(node));
      }

    }

    return delta;
  }

  static Map<String, dynamic> _parseElementStyles(htmlDom.Element element) {
    Map<String, dynamic> attributes = {};

    if (element.localName == 'strong') attributes['bold'] = true;
    if (element.localName == 'em') attributes['italic'] = true;
    if (element.localName == 'u') attributes['underline'] = true;
    if (element.localName == 'del') attributes['strike'] = true;

    final style = element.attributes['style'];
    if (style != null) {
      final colorValue = _parseColorFromStyle(style);
      if (colorValue != null) attributes['color'] = colorValue;

      final bgColorValue = _parseBackgroundColorFromStyle(style);
      if (bgColorValue != null) attributes['background'] = bgColorValue;
    }

    return attributes;
  }

[...]

PR #1523, opened 2023-11-12
feat: Html converter package

freshtechtips:
- *Feature #1100*

Feel free to [vote](https://github.com/singerdmx/flutter-quill/discussions/1524) the package name


scripts/enable_local_dev.sh
```diff
@@ -19,4 +19,9 @@ cp flutter_quill_test/pubspec_overrides.yaml.g flutter_quill_test/pubspec_overri
 
 echo ""
 
+echo "Enable local development for all the other packages..."
+cp packages/flutter_quill_html/pubspec_overrides.yaml.g packages/flutter_quill_html/pubspec_overrides.yaml
```
singerdmx, line 23: is `.g` a typo?

singerdmx (approved):

scripts/enable_local_dev.sh
freshtechtips: No, it's not, I recently add a way to enable and disable local development

freshtechtips: But I think it's a good idea to rename it with `.disabled` instead of `.g` to make it clear, let's do that next

singerdmx:
Are you sure we want to introduce md here?

singerdmx:
It would be nice only having logic for html and delta without use of md

freshtechtips:
> Are you sure we want to introduce md here?

Not sure what you mean, I already added the README.md of the package

singerdmx:
By md I mean markdown

freshtechtips:
> It would be nice only having logic for html and delta without use of md

Yes, they can easily without the `README.md` I also updated the `example`

It only have option to convert the delta to HTML, then convert the HTML back to quill delta and load it, which can emulate the process faster

singerdmx:
Is it like delta <-> mark down <-> html

freshtechtips:
> By md I mean markdown

Currently, there is no built-in way to convert HTML to Delta without building it from scratch, so I use this solution as a workaround, it's from a contributor in GitHub

Of course we will update this package and do this from scratch or do it in better way but for now I think we should focus on `flutter_quill` and `flutter_quill_extensions`

singerdmx:
Ok it does feel hacky but sure as long as it works

freshtechtips:
> hacky

The logic for converting Delta to HTML is fine

The logic for converting the HTML back to Delta is not ready and we will change it in the future

Issue #1318, opened 2023-07-21
[Desktop/Web] Support native triple click text selections


Blquinn:
Looking at https://github.com/flutter/flutter/pull/119046 it looks like triple clicking and some other keyboard text selections are now supported by flutter directly. Can we leverage the built in text selection in flutter to do selection in quill? This should replace the timer based stuff in the example project.

singerdmx:
I agree

DukeJeon:
https://github.com/singerdmx/flutter-quill/issues/1131#issue-1625852384

did you solve this problem?

Blquinn:
No

CatHood0:
I guess this alteady was added and it is supported by the editor currently. I will close by the reason mentione before. @Blquinn, if the feature is not supported, please reopen

EchoEllet:
There was an issue 2 months ago referencing this, so we probably want to keep it for now. We also would want to have some labels for the priority of the issue.

Critical, Major, and Minor.

Issue #468, opened 2021-11-16
[Desktop] triple click on a word to select paragraph


matthiasn:
My issue is about selecting text on macOS. The expectation is that triple clicking a word selects the current paragraph, as is the case in most text editing applications I have tried. Is there a way to achieve the same behavior with `flutter_quill`?

I ran the `example` application and it is the same behavior there, which feels unnatural to me. Thanks! Great library overall.

singerdmx:
@li3317  and I will take a look at it

simonbengtsson:
Would also love to see this! This is not currently supported by flutter text fields either, but seems to be active work happening regarding this recently https://github.com/flutter/flutter/issues/63576

PR #1530, opened 2023-11-14
Remaking the example

freshtechtips:
The first steps of remaking the example

This pull request is not ready yet


singerdmx (approved):

EchoEllet:
> In this PR home_page.dart file was removed, which contained a triple tap for paragraph selection implementation (see #468, #1318, #1758), PR #983.

I removed it in addition to the image and video embeds for the web because it wasn't quite implemented. The docs say the web is supported. Only for the example, though, you will get a runtime exception due to missing embeds in the production app.

The implementation in the extensions package was missing some things that are only implemented in the example.

I think we need to recreate the example once again. It could use a lot of improvements as I made it very quickly at the time of this PR, though this shouldn't be a high priority issue.

Regarding the mentioned issue, I'm not quite familiar with it.

PR #1547, opened 2023-11-21
Restore use of alternative `QuillToolbarLinkStyleButton2` widget

MacDeveloper1:
## Description

There is a `QuillToolbarLinkStyleButton2` widget which suggest an alternative dialog for `link` attribute. It was added in one of the previous versions but was not restored in latest version. Not this widget is available again.

## Breaking Change


singerdmx (approved):

singerdmx:
@freshtechtips please have a look

freshtechtips:
Thank you for your time and efforts to add this feature, but can you take a look at the review? you don't have to but we might make some changes in future releases. We would like to hear your opinion.

PR #1596, opened 2023-12-13
Fix the toggle style button

### LINKED ISSUE #1119: All Font Settings and Styles are Lost on Enter Press
The issue happens on all platforms.

When you change the font the font family, size, color, bold, etc are all dropped when enter is pressed.

For example,

Line 1: You type the first line in the default font.
Line 2: Then you type the second line with a newly selected font. (press enter)
Line 3: This third line returns to the first line default font, instead of continuing the second line font settings.

By default font, I simply mean whatever is set by your app theme or QuillEditor(customStyle: ).

I've run into a few issues `flutter_quill` but this is the only one that seems to stop me from really utilizing the editor. Advice or a fix would be great. Thank you.

freshtechtips:
## Related Issues

- *Fix #1119*

## Breaking Change

### DIFF
--- a/lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
+++ b/lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
@@ -219,7 +219,7 @@ mixin RawEditorStateTextInputClientMixin on EditorState
         );
       }
 
-      // TODO: A bug here too
+      // TODO: The same bug here
 
       if (widget.configurations.controller.selectedFontSize != null) {
         widget.configurations.controller.formatSelection(
@@ -232,6 +232,8 @@ mixin RawEditorStateTextInputClientMixin on EditorState
           ),
         );
       }
+
+      // TODO: The same bug here
 
       widget.configurations.controller.selectedStyles.forEach((key, value) {
         if (value ?? false) {
--- a/lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
+++ b/lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
@@ -208,8 +208,6 @@ mixin RawEditorStateTextInputClientMixin on EditorState
         value.selection,
       );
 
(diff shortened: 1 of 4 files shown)


lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
```diff
@@ -211,9 +211,7 @@ mixin RawEditorStateTextInputClientMixin on EditorState
       // TODO: There is a bug here, the first character is not being formatted
 
       if (widget.configurations.controller.selectedFontFamily != null) {
-        widget.configurations.controller.formatText(
-          diff.start,
-          diff.deleted.length,
```
freshtechtips, line 216: The old code I wrote

lib/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart
```diff
@@ -224,9 +222,7 @@ mixin RawEditorStateTextInputClientMixin on EditorState
       // TODO: A bug here too
 
       if (widget.configurations.controller.selectedFontSize != null) {
-        widget.configurations.controller.formatText(
-          diff.start,
-          diff.deleted.length,
```
freshtechtips, line 229: This too

lib/src/widgets/toolbar/buttons/toggle_style_button.dart
```diff
@@ -221,9 +221,13 @@ class QuillToolbarToggleStyleButtonState
...
-    controller.formatSelection(
-      _isToggled! ? Attribute.clone(widget.attribute, null) : widget.attribute,
-    );
+    controller
+      ..formatSelection(
+        (_isToggled ?? false)
+            ? Attribute.clone(widget.attribute, null)
+            : widget.attribute,
+      )
+      ..selectStyle(widget.attribute, _isToggled == true ? true : false);
```
freshtechtips, line 230: Save the style selection in the controller and use it then

freshtechtips:
@singerdmx Tell me if you would like to change anything

PR #1669, opened 2024-01-10
fix selection.affinity always downstream after updateEditingValue

Lamda303:
## Description
After editing text, the affinity of selection changes to  downstream. When editing text in front of  image, the cursor position is unintended.

https://github.com/singerdmx/flutter-quill/assets/36870873/889af161-5f53-4b0e-bad3-d3a2f89248c0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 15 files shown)


ellet0:
Hi, thank you

can you format the code?

singerdmx:
please pass CI

PR #1684, opened 2024-01-16
feat: Export Container node as QuillContainer

s0nerik:
## Description

Knowing about the `Container` (exported via `QuillContainer` alias) type is needed to perform a generalized lookup for a specific node in user code. Right now it requires handling `Root`, `Line` and `Block` separately.

## Checklist

- [ ] I have run the commands in `./scripts/before_push.sh` and it all passed successfully

## Breaking Change

### DIFF
--- a/lib/flutter_quill.dart
+++ b/lib/flutter_quill.dart
@@ -8,6 +8,7 @@ export 'src/models/config/toolbar/toolbar_configurations.dart';
 export 'src/models/documents/attribute.dart';
 export 'src/models/documents/document.dart';
 export 'src/models/documents/nodes/block.dart';
+export 'src/models/documents/nodes/container.dart';
 export 'src/models/documents/nodes/embeddable.dart';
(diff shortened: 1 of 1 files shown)


lib/flutter_quill.dart
```diff
@@ -37,3 +39,5 @@ export 'src/widgets/toolbar/buttons/alignment/select_alignment_button.dart';
 export 'src/widgets/toolbar/buttons/hearder_style/select_header_style_dropdown_button.dart';
 export 'src/widgets/toolbar/simple_toolbar.dart';
 export 'src/widgets/utils/provider.dart';
+
+typedef QuillContainer = Container;
```
ellet0, line 43: @singerdmx Why don't we refactor `Container` to `QuillContainer`, we could also extends the Container from QuillContainer to not make a breaking change for now and deprecate it and then remove it after a while to notify the users, because this make the source code a little bit more messy and difficult to understand

PR #1688, opened 2024-01-19
fix web cursor position / height (don't use iOS logic)

### LINKED ISSUE #1568: Cursor size is not consistented with font size.
You see the text with font size 36 but the cursor size is not changed but it should.

image

jezell:
## Description

Fixes #1568 

Web is improperly using iOS cursorLogic.


singerdmx:
@ellet0 please have a look

ellet0:
Are you sure? I mean, if it is not specific to native iOS app, we can use it on the iOS browser?

PR #1711, opened 2024-01-28
revert: move sub packages to `./packages`

Cierra-Runis:
## Description

The current project structure is too complex and we need to organize it better.

In fact, `flutter_quill` should not be directly in the root directory. It should also be moved to the `./packages` folder. Maybe I will open another pull request in the future.

## Checklist

## Breaking Change


ellet0:
I already did this in the past, it make hard to manage the packages and navigate through nested folders. In general, I wouldn't do this

Cierra-Runis:
> I already did this in the past, it make hard to manage the packages and navigate through nested folders. In general, I wouldn't do this

But as you can see, there are some files like `./doc`, `.github` and `./script` actually not belongs to `package:flutter_quill`. In fact, those packages such as `riverpod` and `isar` are follow the pattern that separate package content and project content, I consider it's the best practice to manage files.

Issue #1567, opened 2023-12-05
Editor for web app keeps showing "paste" dialog everytime in a read only content page from IOS web browser 


Salvador-dev:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill version

8.6.4

### Steps to reproduce

i'm trying to use the widget in a flutter web app but this issue doesn't have solution since 2022 where another person have the same problem

### Expected results

Don't show the dialog or only one time on the first time in the app.

### Actual results

Showed the dialog every time.

### Code sample

```dart
class ReadingDetailDialog extends StatefulWidget {
  final Reading reading;
  final User user;
  final BaseSubCategory? subCategory;

  const ReadingDetailDialog({
    super.key,
    required this.reading,
    required this.user,
    this.subCategory,
  });

  @override
  State<ReadingDetailDialog> createState() => _ReadingDetailDialogState();
}

class _ReadingDetailDialogState extends State<ReadingDetailDialog>
    with Notificable, Tools, UsesBreakpoints {
  final ReadingBloc _readingBloc;

  _ReadingDetailDialogState()
      : _readingBloc = getIt(),
        super();

  @override
  void initState() {
    _readingBloc.add(RequestOne(widget.reading.id));
    super.initState();
  }

  @override
  Widget build(BuildContext context) {
    return DialogsLayout(
      padding: const EdgeInsets.symmetric(horizontal: 25),
      appBarLeadingWidget: _clientAppBarLeading,
      appBarTitle: _adminAppBarTitle,
      child: BlocBuilder<ReadingBloc, BaseState>(
        bloc: _readingBloc,
        builder: (context, state) {
          if (state is ItemLoaded<DetailedReading>) {
            return _getContent(state.item);
          } else if (state is BaseError) {
            return const TryAgainLaterScreen();
          }
          return LoadingScreen(resource: tr('fields.audio'));
        },
      ),
    );
  }

  Widget get _clientAppBarLeading => widget.user.isClient
      ? CustomDialogDefaults.clientAppBarLeading(
          widget.reading.id,
          getIt<ReadingFavoriteCubit>(),
        )
      : Container();

  Widget? get _adminAppBarTitle =>
      widget.user.isAdmin && widget.subCategory != null
          ? CustomDialogDefaults.adminAppBarTitle(widget.subCategory!)
          : Container();

  Widget _getContent(DetailedReading reading) {
    return Column(
[...]

PR #1722, opened 2024-02-06
Fix safari clipboard bug

mtallenca:
With Safari 17.2.1 on Desktop and iOS 17 Safari - performing Clipboard status checks causes a "Paste" menu to be displayed and stops user input. The status checks were initialized by ClipboardStatusNotifier when a default value of "unknown" was used in it's constructor as well as each time the editor content changed. This PR gives the web a default value of ClipboardStatus.pasteable and doesn't force update calls when content change on web.

Related Issues
Fix Editor for web app keeps showing "paste" dialog everytime in a read only content page from IOS web browser  #1567

Suggestions
Cut/Copy/Paste with keyboard shortcuts work on web, but right click menu items cut/copy/paste do not work correctly. This PR does not address this existing issue.

Breaking Change

[ x] No, this is not a breaking change.

### DIFF
--- a/lib/src/widgets/others/text_selection.dart
+++ b/lib/src/widgets/others/text_selection.dart
@@ -82,7 +82,12 @@ class EditorTextSelectionOverlay {
     // our listener being created
     // we won't know the status unless there is forced update
     // i.e. occasionally no paste
-    clipboardStatus.update();
+    if (!kIsWeb) {
+      // Web - esp Safari Mac/iOS has security measures in place that restrict
+      // cliboard status checks w/o direct user interaction. So skip this
+      // for web
+      clipboardStatus.update();
+    }
   }
 
   TextEditingValue value;
--- a/lib/src/widgets/raw_editor/raw_editor_state.dart
+++ b/lib/src/widgets/raw_editor/raw_editor_state.dart
@@ -4,7 +4,7 @@ import 'dart:math' as math;
 import 'dart:ui' as ui hide TextStyle;
 
 import 'package:collection/collection.dart';
-import 'package:flutter/foundation.dart' show defaultTargetPlatform;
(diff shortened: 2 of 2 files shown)


ellet0:
Huge thanks once again

ellet0:
Thanks once again!

lib/src/widgets/others/text_selection.dart
```diff
@@ -82,7 +82,10 @@ class EditorTextSelectionOverlay {
     // our listener being created
     // we won't know the status unless there is forced update
     // i.e. occasionally no paste
-    if (clipboardStatus != null) {
+    if (clipboardStatus != null && !kIsWeb) {
+      // Web - esp Safari Mac/iOS has security measures in place that restrict
+      // cliboard status checks w/o direct user interaction. So skip this
+      // for web
```
EchoEllet, line 88: If this issue is specific to Safari (Mac/iOS), shouldn't we only introduce this change for safari instead of all browsers?

EchoEllet:
@mtallenca
Can explain why this change was made for all browsers instead of safari only?

mtallenca:
Only thinking I had was operating systems and browsers moving more and more
into the direction of privacy and any data sharing(i.e.
clipboard) requiring user interaction.  I'm still seeing messages in Chrome
console indicating "Third party cookies will be blocked in future Chrome
version as part of privacy sandbox".  I don't think it's a stretch that
additional constraints will be applied to clipboard viewing as well. It may
work now in Chrome, but it would not surprise me if that stops working in
the near future.

EchoEllet:
Thank you for the explanation, I was trying to fix #2220, and I couldn't find where the paste logic is performed for the web. It seems to be separated from non-web platforms. 

So, I initially thought this was related but didn't seem to be the issue.

PR #1746, opened 2024-02-16
Update deprecated code from flutter 3.19

ellet0:
## Description

*Update the project the project to work with flutter 3.19*

The CI always use the latest version, the latest stable version of Flutter 3.19 deprecate some of the code that we are using in the project so I had to update it, I'm not sure if the new api is exist or compatible with flutter 3.16

if not then we need to update the minimum version of flutter from 3.16 to 3.19

## Checklist

- [ ] I read the [Contributor Guide](../CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [ ] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).

## Breaking Change

### DIFF
--- a/dart_quill_delta/lib/src/delta/delta.dart
+++ b/dart_quill_delta/lib/src/delta/delta.dart
@@ -147,7 +147,7 @@ class Delta {
   Operation get last => operations.last;
 
   @override
-  bool operator ==(dynamic other) {
+  bool operator ==(Object other) {
     if (identical(this, other)) return true;
     if (other is! Delta) return false;
     final typedOther = other;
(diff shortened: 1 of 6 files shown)


singerdmx (approved):

dart_quill_delta/lib/src/delta/delta.dart
```diff
@@ -147,7 +147,7 @@ class Delta {
   Operation get last => operations.last;
 
   @override
-  bool operator ==(dynamic other) {
```
ellet0, line 150: I don't think we should use dynamic in the whole project, just to be more safe

singerdmx: yeah

ellet0:
I think there is a bug with the new updated code

image

I'm not sure though

ellet0:
This [commit](https://github.com/singerdmx/flutter-quill/commit/f7dcd55931c166e163392cc7ade91e414bcdecea) should fix some changes of this PR

PR #1774, opened 2024-03-15
fix dispose of text painter

jezell:
This fixes a leak of text painters in the RenderBaselineProxy


jezell:
Thousands of text painters in my heap snapshot, quill is leaking them every time these render objects are created because it doesn't dispose them.

PR #1843, opened 2024-05-02
! Support clipboard actions from the toolbar.

### LINKED ISSUE #1806: "Justify win width" tooltip should just be "Justify"

### Flutter Quill version

_No response_

### Steps to reproduce

hover justify button

### Expected results

tooltip should be "Justify"

### Actual results

tooltip is "Justify win width". This is confusing users because they have never seen justify be called "Justify win width" before.

### Code sample

```dart
[Paste your code here]
```

</details>
  

### Screenshots or Video

<details>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

### LINKED ISSUE #1816: Formatting across lists is broken

### Flutter Quill version

9.3.6

### Steps to reproduce

Starting with an empty editor add three lines of text. Select all and format with bullets. Select all again and then try toggling strikethrough on and off a few times.

### Expected results

It works

### Actual results

After first style is applied to the entire box, formatting won't work unless you only format within the bullet. Formatting buttons also seems to improperly detect whether they are actually applied to the current selection in the same case.

### Code sample

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

AtlasAutocode:
## Description

- Enables clipboard actions to be initiated from toolbar buttons. Previously, actions could only be initiated by keyboard shortcuts or from the popup menu after selecting some text, neither of which are user friendly.

- Fixes errors in copy-paste operations involving selection of partial attribute nodes. Previously, copying a selection with a mix of attribute types resulted in an incorrect result when pasted into the same document. For example: enter the text '**bolder** plainer' and select 'ld plain'. When pasted, the correct result should be '**ld** plain' but the current release shows '**ld pla**in', there are 6 bold characters rather than the expected 2.

- Allows checking if the quill document is readOnly so that direct changes will not be made when the editor is in readOnly mode.

## Related Issues

- *Fix #1816*
- *Fix #1806*
- *Fix: Dividers not shown in toolbar when multiRowDisplay==true*
- *Fix: Toolbar drop buttons are clipped at the text's bottom when multiRowsDisplay==false*

## Improvements
Provides better user control of clipboard actions, particularly for desktop systems.

## Additional notes
*Breaking change:* readonly flag moved from QuillEditorConfigurations to QuillController to enable readOnly state to be determined from outside of the editor. There are no effects if editor never sets readOnly.

## Checklist

- [X] I read the [Contributor Guide](../CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [X] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [X] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [X] All existing and new tests are passing.
- [X] I have run the commands in `./scripts/before_push.sh` and it all passed successfully

## Breaking Change

- [X] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [ ] No, this is *not* a breaking change.


lib/src/widgets/toolbar/buttons/clipboard_button.dart
```diff
@@ -0,0 +1,144 @@
...
+  bool get currentStateValue {
+    switch (widget.clipboardAction) {
+      case ClipboardAction.cut:
+        return !controller.readOnly && !controller.selection.isCollapsed;
+      case ClipboardAction.copy:
+        return !controller.selection.isCollapsed;
+      case ClipboardAction.paste:
+        return !controller.readOnly && (kIsWeb || widget._monitor.canPaste);
+    }
+  }
```
EchoEllet, line 69: I'm currently refactoring some things in #2230 and have removed `canPaste` completely to simplify the `ClipboardService` introduced in #1914. I wasn't aware it's used.

What is the usage of `currentStateValue`? From what I understand, it's used to update the state of the widget and enable the `onClick` handler.

So I had to provide this default:

```dart
/// If the Clipboard is not empty or has something to paste
  Future<bool> get hasClipboardContent async {
    final clipboardData = await Clipboard.getData(Clipboard.kTextPlain);
    return clipboardData != null;
  }
```

However the new `DefaultClipboardService` doesn't override it and will keep it as.

AtlasAutocode: `currentStateValue` is used by all toolbar buttons that have a toggled state.
It is used to determine if the button should be checked or not.

In the case of clipboard actions, if a clipboard action is possible then the button will be checked and the action is possible.
If the button is not checked then the action is not enabled and the user should not expect it to work.

For example: if nothing is selected then cut and copy are not possible.
If the controller is read-only then cut and paste are not possible.
If nothing is on the clipboard then paste is not possible.

> widget._monitor.canPaste

This is needed to monitor the clipboard state in real-time since the user might copy something to the clipboard from another app and the state of our editor needs to be updated. Failure to monitor the clipboard state will introduce a problem.

Specific example: A person starts up their computer and runs Quill Editor. Since the clipboard has no data, the paste button will not be enabled. User types in some text and realizes that they need something from a different app. They start that app, select the text and copy to the clipboard. They return to the quill editor and are surprised that the paste button is not enabled, and they cannot paste their text from the toolbar! With the monitor.canPaste, the monitor will see when the state of the clipboard changes and will update the state of the paste button giving the user feedback that paste is now possible.

Will your new `DefaultClipboardService` affect this process?

EchoEllet: Thank you for the details. 

> Will your new DefaultClipboardService affect this process?

Already did add this back, but the implementation with `super_clipboard` is different. 

Super clipboard can read the different items of the clipboard all at once. The implementation `super_clipboard` only returns the item you need at once.

Previously, this wasn't an issue as the clipboard can only read the data as plain text, but now, since the new implementation already supports almost all of `super_clipboard` features required by the project, like support for copying images from and to the editor, gif files, html text and plan on adding more once solving the existing issues.

The issue is that some of the items of the clipboard are not supported by Flutter `Clipboard` amd we're using it to see if we can paste plain text.

So the user might copy an image from another app or web browser and paste it to the editor, they will notice the paste toolbar button is not activated, since it check only plain text, not the new other elements.

BTW, are you familiar with where the paste logic is triggered on the web? I have disabled different methods related to the pasting, and it's still not disabled on the web. On other platforms, it works as I expect. 

There is probably something I have missed, I haven't tried to fix this bug too much, but if you can give me some hints, I will have more time to focus on #2230 and related issues.

See [#2220](https://github.com/singerdmx/flutter-quill/issues/2220#issuecomment-2351034648).

AtlasAutocode: > BTW, are you familiar with where the paste logic is triggered on the web?

I avoid the web! The only thing I have seen is that there is a lot of code/discussions/comments related to problems with security and permissions with clipboard access and function. You will also see a number of our issues that say their problem is seen only on one specific browser and not on others. So, you may want to check on more than one browser.

EchoEllet: I have tried this on three major browsers, and some other issues are specific to a browser. For example, the Clipboard API is not available on Firefox due to security restrictions. Some other issues on Safari.

I have commented/removed almost everything related to clipboard, text pasting, and it still pastes as plain text on the web.

**What is expected**: the paste functionality is no longer functional, which is the case on non web platforms.

**The actual issue**: The paste functionality is still working

I want to know where because I need to modify it to work with paste events to fix the duplication issue. For now, it will only support copying images to and from the clipboard using Clipboard API (which still doesn't work on Firefox).

We will focus on other issues for now. Thank you for your time and efforts.

EchoEllet:
> _Breaking change:_ readonly flag moved from QuillEditorConfigurations to QuillController to enable readOnly state to be determined from outside of the editor. There are no effects if editor never sets readOnly.

This breaking change was releases as a minor version and doesn't exist in `QuillController` or `QuillControllerConfigurations` when creating the controller using the `QuillController.basic`.

EchoEllet:
> _Breaking change:_ readonly flag moved from QuillEditorConfigurations to QuillController to enable readOnly state to be determined from outside of the editor

@AtlasAutocode It seems this was already possible, the user can create their own `_readOnly` and manage it from their toolbar and pass it to the `QuillEditor`. We should have more consideration before adding new features and that's something you and I agree on. I have introduced some changes or maybe many without creating an issue or discuss it, reviewing them with cautious.

AtlasAutocode:
At the time this was done, I found that toolbar buttons did not have access to the toolbar that contains them, or the editor, and hence did not have access to the read-only state maintained by the user.

I have just researched this and eventually found `requireQuillEditorConfigurations` which is not documented anywhere, very hard to find and barely used anywhere. This would place read-only back in the configurations which should be const?

I would argue that configurations define how the application works and, as such, should be static, const and never change.

In contrast, read-only is dynamic and can change frequently. As such, the controller is a better place for this. But this is your decision.

You are making a lot of changes as you review and reorganize the code with a view to stabilizing it. I commend your activity, but it leaves me uncertain as to where I can be of use.

EchoEllet:
> this is your decision.

This is a community-driven project and is not a project managed all by one person.

> `requireQuillEditorConfigurations`

This feature will be removed in the next breaking change release, it's only used internally and not meant to be used by user projects. It has design issues and full of bugs and is inefficient. It's similar to how Flutter widgets work internally though for our use case, we should not use it.

This feature is possible without `requireQuillEditorConfigurations` either but if we're talking about having the clipboard actions built-in in the user project, then this change is probably necessary.

The user can define their own `_readOnly`, update it from the toolbar, and pass to the `QuillEditorConfigurations`. However, my question is, if this is not a commonly requested feature, we should have at least introduced it in non-breaking change way.

> As such, the controller is a better place for this. But this is your decision.
> it leaves me uncertain as to where I can be of use.

It may sound like I'm against though actually, I'm with your change, your change itself doesn't have any issues, we should have discussed this, opened an issue, done more planning, and definitely avoided breaking changes in a new minor version, the output the user will get make it hard to debug the issues.

Confirming to [semver](https://semver.org/) guidelines is essential knowing that pub integrates with semver.

> In contrast, read-only is dynamic and can change frequently. As such, the controller is a better place for this. But this is your decision.
> I would argue that configurations define how the application works and, as such, should be static, const and never change.

We do have another parameter in the configuration that will prevent the use of the `const` keyword, is the performance our main concern or it's a matter of preference when making this decision? I suggested moving the `controller` since it always prevents the `const` keyword, I expect a negligible impact on performance and there are many aspects where we can improve the performance.

The issue is that things are being added randomly without understanding them, and we can't understand all of them if docs are missing.

At the moment I can't reply properly since I'm tight with other tasks which is why this message is not quite clear to the point.

lib/src/models/config/toolbar/simple_toolbar_configurations.dart
```diff
@@ -107,6 +107,9 @@ class QuillSimpleToolbarConfigurations extends QuillSharedToolbarProperties {
     this.showSearchButton = true,
     this.showSubscript = true,
     this.showSuperscript = true,
+    this.showClipboardCut = true,
```
EchoEllet, line 110: Showing new buttons in a minor version is also a breaking change. We should focus on having less buttons and an opinioned toolbar that fit the needs of most of our users, noticed that currently, the toolbar takes more than half of the screen, which is an issue I introduced in 2023 when I introduced support for Material 3.

We should make it take less space, with less buttons, and instead have sections or groups where each group have related buttons, users can navigate between them, we should not touch the `QuillSimpleToolbar` anymore and introduce a new toolbar unless most developers want something different, currently, we don't have any reports regarding the toolbar though I noticed some developers already started to develop their our toolbar.  We should introduce a new one instead of changing this. Or maybe provide those buttons so they can use them somewhere different than the toolbar.

The toolbar is one of the things that really need improvements. The UI is important for the end user.

lib/src/models/documents/nodes/line.dart
```diff
@@ -407,20 +407,20 @@ base class Line extends QuillContainer<Leaf?> {
     final data = queryChild(offset, true);
     var node = data.node as Leaf?;
     if (node != null) {
```
EchoEllet, line 409: Is this change related? Not sure why we need it for this feature.

lib/src/widgets/quill/quill_controller.dart
```diff
@@ -471,16 +474,157 @@ class QuillController extends ChangeNotifier {
     return document.querySegmentLeafNode(offset).leaf;
   }
 
-  /// Clipboard for image url and its corresponding style
-  ImageUrl? _copiedImageUrl;
+  // Notify toolbar buttons directly with attributes
+  Map<String, Attribute> toolbarButtonToggler = const {};
+
```
EchoEllet, line 479: IMO the controller should not have anything related or specific to the toolbar. Should annotate most of the newly added properties in the last year as experimental as we're removing them as we discover issues and more users start using them.

test/bug_fix_test.dart
```diff
@@ -65,7 +65,6 @@ void main() {
           configurations: QuillEditorConfigurations(
             controller: controller,
             // ignore: avoid_redundant_argument_values
```
EchoEllet, line 67: We should probably update our `analysis_options.yaml` and also remove those comments.

test/widgets/editor_test.dart
```diff
@@ -27,7 +27,6 @@ void main() {
             configurations: QuillEditorConfigurations(
               controller: controller,
               // ignore: avoid_redundant_argument_values
```
EchoEllet, line 29: Same as the previous comment, some lints in `analysis_options.yaml` need to be removed.

EchoEllet (commented):
I'm starting to finish #2230 (except lack of Windows support issue and error handling). A few things here could use some minor improvements.

lib/src/widgets/quill/quill_controller.dart
```diff
@@ -471,16 +474,157 @@ class QuillController extends ChangeNotifier {
...
+        replaceText(sel.start, sel.end - sel.start, '',
+            TextSelection.collapsed(offset: sel.start));
+      }
+      return true;
+    }
+    return false;
+  }
+
+  /// Returns whether paste operation was handled here.
+  /// updateEditor is called if paste operation was successful.
```
EchoEllet, line 515: We need to be more specific about this, it looks like it return `true` in read-only mode, and the doc comment indicate that this is a successful clipboard operation where it ignored the call due to `readOnly`.

EchoEllet:
> This breaking change was released as a minor version and doesn't exist in `QuillController` or `QuillControllerConfigurations` when creating the controller using the `QuillController.basic`.

We really need to solve this issue ASAP, see [this comment](https://github.com/singerdmx/flutter-quill/pull/2230#issuecomment-2386277135).

lib/src/widgets/quill/quill_controller.dart
```diff
@@ -471,16 +474,157 @@ class QuillController extends ChangeNotifier {
...
 
-  /// Clipboard for image url and its corresponding style
-  ImageUrl? _copiedImageUrl;
+  // Notify toolbar buttons directly with attributes
+  Map<String, Attribute> toolbarButtonToggler = const {};
+
+  /// Clipboard caches last copy to allow paste with styles. Static to allow paste between multiple instances of editor.
+  static String _pastePlainText = '';
+  static List<OffsetValue> _pasteStyleAndEmbed = <OffsetValue>[];
+
```
EchoEllet, line 483: > Static to allow paste between multiple instances of editor.

This might not be the expected behavior?  The copied image URL is not shared across different instances:

```dart
ImageUrl? _copiedImageUrl;
ImageUrl? get copiedImageUrl => _copiedImageUrl;
```

PR #1710, opened 2024-01-28
fix bug #1696 - Allow insert/replace styled text from clipboard

mtallenca:
## Description

When pasting html data, edit the current document and insert / replace at the correct location w/o creating a new document so listeners continue to work.

- *Fix #1696*

Does your PR require plugin users to manually update their apps to accommodate your change?

- [x ] No, this is *not* a breaking change.

PR #1890, opened 2024-05-23
fix: toolbar style change will be invalid when inputting more than 2 characters at a time

### LINKED ISSUE #1864: When inputting more than 2 characters at a time, the toolbar style change will be invalid.

### Flutter Quill version

9.3.11

### Steps to reproduce

1. Enter any text you want, as long as there are candidates in the input method that are more than 2 characters long.
2. Toolbar switches font size or color。
3. Click on the candidate content of more than 2 characters in the input method
4. The style selected in the second step does not take effect

### Expected results

No matter how much text is entered, the toolbar switching style will take effect

### Actual results

When more than two characters are entered at one time, the switching style will not take effect.

### Code sample

[After a day of hard work, I have found the code location that caused the bug.](https://github.com/singerdmx/flutter-quill/blob/9d84ef65439c93b4ee90ad1e8e5eddae587ad3d7/lib/src/widgets/quill/quill_controller.dart#L276)

This judgment condition will result in that when the user inputs more than 2 words, the toggledStyle cannot be obtained.

You should distinguish between user pasting and candidate input. The pasted content has styles, while the candidate content in the input method has no style.

### Screenshots or Video

<details>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

crasowas:
## Description

This is a regression caused by #1710. I believe there is a reason why pasted text should not use toggledStyle, but unfortunately, I couldn't find more details about it.

However, when I allowed pasted text to use toggledStyle to fix issue #1864, an [assertion](https://github.com/singerdmx/flutter-quill/blob/9c2e51ff73c244467e71c6f1861ec8a51518fb5e/lib/src/models/documents/nodes/line.dart#L148) failure occurred when pasted text that contains line breaks. This might be the reason why pasted text should not use toggledStyle.

This PR fixes issue #1864 and processes pasted text line by line.

## Related Issues

- Fix #1864 
- Related #1710 

## Checklist

## Breaking Change

### DIFF
--- a/lib/src/widgets/quill/quill_controller.dart
+++ b/lib/src/widgets/quill/quill_controller.dart
@@ -261,9 +261,7 @@ class QuillController extends ChangeNotifier {
       var shouldRetainDelta = toggledStyle.isNotEmpty &&
           delta.isNotEmpty &&
           delta.length <= 2 &&
-          delta.last.isInsert &&
-          // pasted text should not use toggledStyle
-          (data is! String || data.length < 2);
+          delta.last.isInsert;
       if (shouldRetainDelta &&
           toggledStyle.isNotEmpty &&
           delta.length == 2 &&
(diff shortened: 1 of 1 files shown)


ellet0:
> This is a regression caused by #1710

Thank you for pointing that out.

The goal of #1710 PR is to fix a bug that was caused by another PR sent by me. The feature I was implementing was improvement & and fixed another bug

 it appears that both PRs inadvertently introduced new issues.

PR #1904, opened 2024-05-28
Feat: search bar improved

Alspb:
## Description

Search bar improved: alignment option added, search settings changed from icons to checkboxes with titles, search happens on the fly, other minor improvements.

## Checklist

- [ ] I have run the commands in `./scripts/before_push.sh` and it all passed successfully

## Breaking Change


ellet0:
Thank you for your contributions

> \"Search\" button is removed since search is now on-the-fly (i.e. when the user enters text).

This might be considered as breaking change by some developers. It doesn't require additional changes to work but changes the behavior 

Can we add an option to change the behavior? Also, for large documents, this can be more resources 

>  I added searchBarAlignment parameter into search button options. It's just responsible for the search dialog alignment (previously the search dialog was always at the bottom thus overlapping the editor if the toolbar is on the top). Please double-check if you're fine with how this parameter is set in the code.

I'm okay in either case

> If you could help with the translation ('Case sensitive and whole word' is now split into 'Case sensitive' and 'Whole word' separately, also 'Close' and 'Search settings' tooltips added.), it would be perfect.

I'm not near my main machine, I will back soon to review, test the changes and the translations 

Thank you once again.

singerdmx:
I think we just need to bump major version.
It is ok to have breaking change that changes behavior

singerdmx:
I mean there is no need to have an option to go back to old behavior which is tedious to maintain and readability is bad.

lib/src/widgets/toolbar/buttons/search/search_button.dart
```diff
@@ -59,11 +59,11 @@ class QuillToolbarSearchButton extends StatelessWidget {
...
-  Color _dialogBarrierColor(BuildContext context) {
-    return options.dialogBarrierColor ??
-        context.quillSharedConfigurations?.dialogBarrierColor ??
-        Colors.black54;
-  }
+  // Color _dialogBarrierColor(BuildContext context) {
+  //   return options.dialogBarrierColor ??
+  //       context.quillSharedConfigurations?.dialogBarrierColor ??
+  //       Colors.black54;
+  // }
```
ellet0, line 66: This would allow customizing the dialog barrier color for all the dialogs, the previous code is far from perfect but if the new default for this dialog is `Colors.transparent` then we should consider applying it in this function, if we decide to remove function, it's better to remove the comments of the previous code as the code will get outdated quickly since the IDE will ignore the commented code, and it makes the code a little bit less readable

lib/src/widgets/toolbar/buttons/search/search_button.dart
```diff
@@ -124,13 +124,13 @@ class QuillToolbarSearchButton extends StatelessWidget {
       return;
     }
     await showDialog<String>(
-      barrierColor: _dialogBarrierColor(context),
+      barrierColor: Colors.transparent,
```
ellet0, line 127: If we keep the previous function, then this should be: `_dialogBarrierColor(context)`

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
```diff
@@ -110,100 +98,170 @@ class QuillToolbarSearchDialogState extends State<QuillToolbarSearchDialog> {
...
+    }
+
+    final searchBar = Container(
+      height: addBottomPadding ? 50 : 45,
+      padding: addBottomPadding ? const EdgeInsets.only(bottom: 12) : null,
+      child: Row(
+        children: [
+          IconButton(
+            icon: const Icon(Icons.close),
+            tooltip: 'Close',
```
ellet0, line 123: It doesn't need to be translated to 40 languages but we should consider adding this to `quill_en.arb` file, it will update the `untranslated.json` file to allow developers to update translations easily

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
```diff
@@ -110,100 +98,170 @@ class QuillToolbarSearchDialogState extends State<QuillToolbarSearchDialog> {
...
+      height: 45,
+      child: Row(
+        children: [
+          Expanded(
+            child: CheckboxListTile(
+              controlAffinity: ListTileControlAffinity.leading,
+              visualDensity: VisualDensity.compact,
+              contentPadding: EdgeInsets.zero,
+              title: const Text(
+                'Case sensitive',
```
ellet0, line 179: Same like above, const has to be removed once the change is done

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
```diff
@@ -110,100 +98,170 @@ class QuillToolbarSearchDialogState extends State<QuillToolbarSearchDialog> {
...
+            tooltip: 'Close',
+            visualDensity: VisualDensity.compact,
+            onPressed: () {
+              Navigator.of(context).pop();
+            },
+          ),
+          IconButton(
+            icon: const Icon(Icons.more_vert),
+            isSelected: _caseSensitive || _wholeWord,
+            tooltip: 'Search settings',
```
ellet0, line 132: Same as above

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
```diff
@@ -110,100 +98,170 @@ class QuillToolbarSearchDialogState extends State<QuillToolbarSearchDialog> {
...
+    final searchBarAtBottom = (searchBarAlignment == Alignment.bottomCenter) ||
+        (searchBarAlignment == Alignment.bottomLeft) ||
+        (searchBarAlignment == Alignment.bottomRight);
+    final addBottomPadding = searchBarAtBottom && isMobile(supportWeb: true);
+    var matchShown = '';
+    if (_text.isNotEmpty) {
+      if (_offsets.isEmpty) {
+        matchShown = '0/0';
+      } else {
+        matchShown = '${_index + 1}/${_offsets.length}';
```
ellet0, line 112: Maybe this should be translated too? with using some variables inside the translate so the result won't be a variable, instead, it will be a function and then we will pass the `${_index + 1}` and `${_offsets.length}`

the `/` and the direction of the text can be different in different languages

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
```diff
@@ -42,55 +45,40 @@ class QuillToolbarSearchDialog extends StatefulWidget {
   const QuillToolbarSearchDialog({
     required this.controller,
     this.dialogTheme,
-    this.text,
```
ellet0, line 45: This sounds like a breaking change, not all developers use the built-in toolbar, some of them use the widgets to create their own toolbar, we don't have to support this anymore, just add `@Deprecated` with a message of your choice that indicates this is no longer used

Or we could still use this if you want to, the `_text = widget.text ?? '';` shouldn't be removed from the `initState`

ellet0 (changes requested):
Just a few changes to be addressed, I will update the `quill_en.arb` file to include the new needed translations, you just have to pull the changes by merging them or rebase to your branch and use the new translations

ellet0:
> ('Case sensitive and whole word' is now split into 'Case sensitive' and 'Whole word' separately, also 'Close' and 'Search settings' tooltips added.), it would be perfect.

Added in [592da6c7438b995bfeb7231f66dfd834bda546b3](https://github.com/singerdmx/flutter-quill/commit/592da6c7438b995bfeb7231f66dfd834bda546b3) commit, pull the changes from the upstream master branch to your branch so you can start using the new translations, notice I only added the keys and used the script to update `untranslated.json` to help developers translate them to their languages more easily, maybe we should use some tool or script to help automating the translations as a default values as the translations we have are pretty simple and can be automated, this is not directly related to the PR

lib/src/widgets/toolbar/buttons/search/search_button.dart
ellet0: > Do you have any idea why barrierColor property could be of use here?

There was a feature request, I don't exactly remember the problem. Maybe they used a different theme or some different color/custom component something like this. They wanted an option to change it as the package depends on Material 3, developers already expecting this, we did too many breaking changes in the last year (mostly by me), and it was disturbing for developers who are trying to get their job done

I suggest minimizing the breaking changes and doing it only when it's really necessary 

In this case, it's possible to not break change without any disadvantages 

if we all agreed on removing this property, this change of barrier color, the change should be in a separate PR. There are other places, and almost all dialogs have this option. Having an option to work on all dialogs except one might be seems as a bug or inconsistency.

lib/src/widgets/toolbar/buttons/search/search_dialog.dart
ellet0: > In think it's better to be in-line with how it looks in Chrome, Firefox, Edge, etc. Are you able to check how it looks there for right-to-left languages?

Yes, I checked how it looks in Rtl last year 

Usually, flutter and flutter localizations take care of this without changing much in the UI code

If you want to see how it looks, change the text direction to Rtl in MaterialApp/WidgetsApp/CupertinoApp

Or use one of those languages:
https://www.andiamo.co.uk/resources/right-to-left-languages/

Either by temporarily changing the device/emulator language or explicit use of `local` in the app widget

Or use the`Directionality` widget to override this

I should have added an option in the example to change the language in the app

ellet0:
Adding an option to have a delay seems a good idea,
While I prefer the new widget,
I still suggest adding an option for different styles or widget implementation. An enum class with two options will do. If the logic is too different, then we could create another widget in the same folder for the old widget (before this PR) that has the bussniues logic, UI logic, and logic in general, the if check should be in the simple toolbar widget to minimize the conditional checks when using custom toolbar or using the components outside of the toolbar


- If you would like to do it yourself
- or leave it to me
- make a breaking change

I suggest any of the first two instead of the third

the choice is up to you.

ellet0:
> I tried to combine these two widgets but the logic is different in several places of the code, so that the whole widget becomes looking overcomplicated.

I agree with you, it's better to not combine the two in the same widget to avoid too many conditional checks and other issues like we both mentioned above.

What I suggested is that we use the old widget (before merging this PR), refactor the name to something else, or to avoid breaking changes, we could separate this new widget with a different name that's descriptive and clear and use that as a default in `QuillSimpleToolbar` which is what I suggest.

**In short**, I suggest to not change the current search bar/button widget (revert the changes) and instead introduce a new one and use it as default in the Toolbar.

> Could you please make a try?

Sure, will see your response on this first.

singerdmx:
@Alspb any update?

singerdmx:
@ellet0 do you think we can go ahead merging this change and follow up with additional changes as needed?

ellet0:
> @ellet0 do you think we can go ahead merging this change and follow up with additional changes as needed?

Yes, I will restore the old widget with a different name and make some changes that fit the rest of the options, configurations, and customization in other widgets as soon as I can.

An enum class will be used to change the widget to the new one in the `QuillSimpleToolbar`

Might make the old widget as default to avoid breaking behavior or the new one to simplify the development and configurations for new developers.

Developers who use a custom toolbar will have to use either the old or new widget to avoid conditional checks.

Or we could publish a new version directly with this new widget is as.

singerdmx:
Sounds good. I will go ahead merging this PR and you can patch it with changes later.

singerdmx:
hmm, not sure why CI failed on this commit

ellet0:
> hmm, not sure why CI failed on this commit

I tried running the workflow again, and it succeeded [26101906925](https://github.com/singerdmx/flutter-quill/actions/runs/9472757664/job/26101906925)

I'm not sure why it didn't show any details previously. The actions and steps or where it's failing.

PR #1955, opened 2024-06-28
Add support for html underline and videos 

### LINKED ISSUE #1917: Underline Styles

### The question

Hi, the issue is that when I write styles like underline, it takes it correctly when editing it, when saving the HTML value has something similar to `<p><u>Underline</u></p>, ` but when returning or setting that direct value in the controller does not work, it takes it as a simple paragraph `<p>` and does not assign the corresponding style. Can anyone guide me please?.

### LINKED ISSUE #1945: Converting from HTML not rendering the videos and the iframes

### The question

How do I go about this, I want to convert from HTML, the text and the Images are working but the videos and the iframes were not showing at all.

CatHood0:
## Description

If you try to use `DeltaX.fromHtml` with a tag `<u>` to use it as a underline style, then will be removed and just pass content as plain. Videos has the same result of the before one, if you try to use an `<iframe>` will be paste as a plain text or removed (idk really this last one case).

## Related Issues

*Fix #1917* 
*Fix #1945*

## Improvements

- Was added support for local underline syntax to detect and apply on resulting delta.
- Was added support for iframes with video url's (by now just can verify youtube videos)

## Checklist

## Breaking Change


ellet0:
Thank you for your contributions, I will review and test the changes soon.

lib/src/utils/html2md_utils.dart
```diff
@@ -0,0 +1,126 @@
+// ignore_for_file: implementation_imports
```
ellet0, line 1: Cam, can we avoid creating this file?

It's out of scope for `flutter_quill`, and the reason we have Markdown and HTML converting in `flutter_quill`

It's only to support rich text pasting support,  the reason the convert from HTML and Markdown functions is in `flutter_quill` instead of the converters as the convert to require additional dependencies that I prefer to not using them to avoid add too many dependencies to the developers project.

Once the convert logic is fully rewritten from scratch, we can solve this by adding the converter packages that have minimal dependencies, in short, I prefer to avoid adding logic related to HTML or Delta too much in `flutter_quill`

We only did this for the rich pasting feature. Usually, the user doesn't copy and paste videos, so the current solution works well for this use case. Adding more code to cover more use cases also means more maintenance over time, which is why I prefer to make this optional in the converters or move them somewhere else.

The current project structure could use a lot of refactoring and improvements, which is another reason I prefer to add anything extra.

ellet0 (changes requested):
Thank you for adding tests and your contributions once again, I haven't tested this change yet.

Just a few things to consider before releasing this change.

lib/src/utils/html2md_utils.dart
CatHood0: So, are you suggesting that I delete it completely and move the logic to Deltax again to avoid continuing to create unnecessary files?

lib/src/models/documents/delta_x.dart
```diff
@@ -32,9 +45,48 @@ class DeltaX {
   /// used for **production applications**.
   ///
   @experimental
-  static Delta fromHtml(String htmlText) {
-    final markdownText = html2md.convert(htmlText).replaceAll('unsafe:', '');
-
+  static Delta fromHtml(
+    String htmlText, {
+    Html2MdConfigs? configs,
```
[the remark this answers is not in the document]
CatHood0: Then, should i just put in what is necessary to solve the problems with italics, underlines and iframes instead of creating a whole class for its settings?

lib/src/utils/html2md_utils.dart
ellet0: > So, are you suggesting that I delete it completely and move the logic to Deltax again to avoid continuing to create unnecessary files, or just remove video implementation and left just underline fix?


I prefer to avoid creating new files or maybe create a new directory for this.

Yet you can keep both, it's might be a good idea to keep this logic in a new file as you did it to avoid exporting anything as the `DeltaX` class file is exported.

lib/src/models/documents/delta_x.dart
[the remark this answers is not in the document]
ellet0: That's what I suggest. It's not necessary, though.

lib/src/utils/html2md_utils.dart
singerdmx: @ellet0 Let's not block feature/fix. We can always refactor/reorganize code around later.

singerdmx:
@CatHood0 Please resolve the conflict. Please do not worry too much about where is the best place to put the code. I believe @ellet0 can always move around code to make sense OOD wise.

lib/src/utils/html2md_utils.dart
CatHood0: Sorry, i'm force a commit without update the local folder. I will see how fix this

CatHood0:
Yeah, i'm trying to resolve the before commit t

singerdmx:
@ellet0 I am going ahead and merging this. Please feel free to patch it after this.

lib/src/utils/html2md_utils.dart
ellet0: > @ellet0 Let's not block feature/fix. We can always refactor/reorganize code around later.

It would require making breaking changes.

singerdmx: @ellet0 By breaking do you mean not backward compatible?

singerdmx: If yes, I think it is acceptable and we just need to release with a major version change

PR #1964, opened 2024-07-02
Perf: Performance optimization

### LINKED ISSUE #997: [All] Various Performance Issues
The performance of the editor begins to degrade when the document size gets reasonably large, especially when performing edits.

I ran some profiles with 500 paragraphs of text loaded. (from https://loremipsum.io/generator/?n=500&t=p)

It seems like the main issue is that the entire widget is re-drawn and re-laid-out on every edit. Meaning, every line of text is re-drawn on every frame, regardless of if it changed, or not. I think this should be avoidable. I also profiled https://github.com/superlistapp/super_editor and there was no jank when editing large documents, perhaps insight can be gained from there.

image

A couple of smaller issues:

Rasterization is slow on every frame, this seems to be due to the cursor blinking, but it only happens on larger documents, so I assume it's redrawing the whole document when the cursor blinks, or something like that.

image

Also, in RawEditorStateTextInputClientMixin calling "Document.toPlainText()" and diffing that against the editor text seems to be very slow for large documents.

Alspb:
## Description

Performance optimization:

1. Caches Node.length and Node.offset since these two methods are called very often (partly due to recursion inside) and cause significant lags for large documents. The trickiest thing here is to clear the cache when necessary.
2. Removes unnecessary `Document.from` and changes the first argument type in several methods from `Delta document` to `Document document` (which is anyway reasonable for a variable named "document").
3. Change a check including `_root.toDelta()` to `assert`.

## Related Issues

- *Fix #997*

## Checklist

## Breaking Change


lib/src/models/documents/nodes/line.dart
```diff
@@ -145,7 +145,7 @@ base class Line extends QuillContainer<Leaf?> {
       assert(style.values.every((attr) =>
           attr.scope == AttributeScope.inline ||
           attr.scope == AttributeScope.ignore));
-      assert(index + local != thisLength);
+      assert(index + local != length);
```
EchoEllet, line 148: Out of scope for this PR however, can you add a message to this assert to help debug and make the code a bit easier to understand?

Thank you for your contributions.

EchoEllet: I'm unfamiliar with this code and unable to take a look due to the lack of time. I assumed you were as you have already updated similar parts.

> need some time to understand the code.

It's not a high-priority issue for now however you can take a look at it if you're interested. Thank you for your time.

lib/src/models/rules/delete.dart
```diff
@@ -121,9 +122,9 @@ class EnsureEmbedLineRule extends DeleteRule {
   const EnsureEmbedLineRule();
 
   @override
-  Delta? applyRule(Delta document, int index,
+  Delta? applyRule(Document document, int index,
       {int? len, Object? data, Attribute? attribute}) {
-    final itr = DeltaIterator(document);
+    final itr = DeltaIterator(document.toDelta());
```
EchoEllet, line 127: Is there any reason why the type changed from `Delta` to `Document`? It's looks like it's being converted back to Delta.

lib/src/models/rules/format.dart
```diff
@@ -159,7 +160,7 @@ class ResolveInlineFormatRule extends FormatRule {
 
   @override
   Delta? applyRule(
-    Delta document,
+    Document document,
```
EchoEllet, line 163: The same goes for other rules.

Issue #1697, opened 2024-01-23
Header Selection Dropdown opens under keyboard and is not visible


Sesa1988:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Steps to reproduce

- Create Editor
- Put Toolbar below it to the bottom
- Expand Header Selection Font Size

### Expected results

Dropdown opens to the top

### Actual results

Dropdown opens to the bottom below the keyboard

### Code sample

<details><summary>Code sample</summary>

```dart
return Column(
      children: [
        Expanded(
          child: QuillEditor.basic(
            configurations: QuillEditorConfigurations(
              controller: _controller,
              placeholder: AppLocalizations.of(context).translate('note'),
              readOnly: false,
              scrollable: true,
              expands: true,
              autoFocus: widget._shouldFocus,
              padding: const EdgeInsets.only(left: 18, right: 18),
              sharedConfigurations: const QuillSharedConfigurations(
                locale: Locale('en'),
              ),
            ),
            focusNode: widget._focusNode,
            scrollController: ScrollController(),
          ),
        ),
        Padding(
          padding: const EdgeInsets.only(bottom: 8),
          child: QuillToolbar.simple(
            configurations: QuillSimpleToolbarConfigurations(
              controller: _controller,
              showFontFamily: false,
              showFontSize: false,
              showSubscript: false,
              showSuperscript: false,
              multiRowsDisplay: showBigToolbar,
              showRedo: true,
              showUndo: true,
              buttonOptions: QuillSimpleToolbarButtonOptions(
                base: QuillToolbarBaseButtonOptions(
                  iconSize: showBigToolbar ? 14 : 16,
                ),
              ),
              sectionDividerSpace: showBigToolbar ? 0 : 16,
              toolbarSectionSpacing: showBigToolbar ? 0 : 4,
              headerStyleType: HeaderStyleType.original,
              sharedConfigurations: const QuillSharedConfigurations(
                locale: Locale('en'),
              ),
            ),
          ),
        ),
      ],
    );
```

</details>

### Screenshots or Video

<details>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

ellet0:
[...]

Issue #10, opened 2021-01-29
[web] Cursor error when click bold


artemsivcev:
Hello! I'm using your plugin on web and I found some bug. 
v.0.2.2 (last for now).
1. Type some text. Cursor must be at end of text
2. Select bold, italic, etc...
Expected result:
3. Cursor still at the end and next text become bold, italic, etc
Actual result:
3. Cursor move at start of text but next typed text become bold, italic, etc and its position right on the end
4. And after that there is some strange behaviour about delete some letters, some times delete button delete 2 or more letters...

<img width="416" alt="Screenshot 2021-01-29 at 09 35 49" src="https://user-images.githubusercontent.com/5911567/106251537-24c37280-6226-11eb-8c58-4c389f04f921.png">

singerdmx:
For web development, ReactQuill https://github.com/zenoamaro/react-quill is recommended to use for compatibility.
If you use our app https://bulletjournal.us/home/index.html, the way we solve it is to only use delta for data between mobile and web. HTML is actually not needed for mobile side. At web side, we use ReactQuill https://github.com/zenoamaro/react-quill to generate html.
Bullet Journal app https://bulletjournal.us/home/index.html is open sourced at https://github.com/singerdmx/BulletJournal

singerdmx:
@artemsivcev As mentioned in the README, this library has no intention of supporting web. ReactQuill https://github.com/zenoamaro/react-quill is recommended for web development.
Note that our underlying data format is 100% compatible with quill delta.

singerdmx:
Closing

PR #1972, opened 2024-07-04
Feat: Line height support

### LINKED ISSUE #1956: Feature: Line height attribute

### Proposal

I propose creating an attribute to be able to modify the space between lines, of course, I am not just proposing whether to do my part. I already have the implementation done with the button next to the attribute, however, I would like to know if there are plans to implement this feature, since I have seen several ones (like me) that for their applications, an attribute like line-height , it would be very good for us, since it makes the editor more complete (and the spice between words and letters is another story.

The issue is that, if it is planned, I would like to know where I could start looking to add it natively to the project.

This is how work by now:

https://github.com/singerdmx/flutter-quill/assets/114286961/927836e4-5c6c-4835-8124-90d4aac911fe

CatHood0:
## Description

Support was added for a feature that has been requested for some updates to dynamically change the spacing between lines

## Related Issues

*Fix #1956*

## Checklist

## Breaking Change

### DIFF
--- a/lib/src/models/documents/attribute.dart
+++ b/lib/src/models/documents/attribute.dart
@@ -40,6 +40,7 @@ class Attribute<T> extends Equatable {
     Attribute.background.key: Attribute.background,
     Attribute.placeholder.key: Attribute.placeholder,
     Attribute.header.key: Attribute.header,
+    Attribute.lineHeight.key: Attribute.lineHeight,
     Attribute.align.key: Attribute.align,
     Attribute.direction.key: Attribute.direction,
     Attribute.list.key: Attribute.list,
@@ -57,11 +58,9 @@ class Attribute<T> extends Equatable {
 
   static const BoldAttribute bold = BoldAttribute();
 
-  static final ScriptAttribute subscript =
-      ScriptAttribute(ScriptAttributes.sub);
+  static final ScriptAttribute subscript = ScriptAttribute(ScriptAttributes.sub);
 
-  static final ScriptAttribute superscript =
-      ScriptAttribute(ScriptAttributes.sup);
+  static final ScriptAttribute superscript = ScriptAttribute(ScriptAttributes.sup);
 
   static const ItalicAttribute italic = ItalicAttribute();
 
@@ -86,6 +85,8 @@ class Attribute<T> extends Equatable {
   static const PlaceholderAttribute placeholder = PlaceholderAttribute();
 
   static const HeaderAttribute header = HeaderAttribute();
+
+  static const LineHeightAttribute lineHeight = LineHeightAttribute();
 
   static const IndentAttribute indent = IndentAttribute();
 
@@ -135,6 +136,7 @@ class Attribute<T> extends Equatable {
     Attribute.blockQuote.key,
     Attribute.indent.key,
     Attribute.direction.key,
+    Attribute.lineHeight.key,
   });
 
   static final Set<String> blockKeysExceptHeader = LinkedHashSet.of({
@@ -142,6 +144,7 @@ class Attribute<T> extends Equatable {
     Attribute.align.key,
     Attribute.codeBlock.key,
     Attribute.blockQuote.key,
+    Attribute.lineHeight.key,
     Attribute.indent.key,
     Attribute.direction.key,
   });
@@ -175,6 +178,18 @@ class Attribute<T> extends Equatable {
 
   /// "attributes":{"header": 6 }
   static const Attribute<int?> h6 = HeaderAttribute(level: 6);
+
+  /// "attributes":{"line-height": 1.0 }
+  static const Attribute<double?> lineHeightNormal = LineHeightAttribute(lineHeight: 1);
+
+  /// "attributes":{"line-height": 1.15 }
+  static const Attribute<double?> lineHeightTight = LineHeightAttribute(lineHeight: 1.15);
+
+  /// "attributes":{"line-height": 1.5 }
+  static const Attribute<double?> lineHeightOneAndHalf = LineHeightAttribute(lineHeight: 1.5);
+
+  /// "attributes":{"line-height": 2.0 }
+  static const Attribute<double?> lineHeightDouble = LineHeightAttribute(lineHeight: 2);
 
   // "attributes":{"align":"left"}
   static const Attribute<String?> leftAlignment = AlignAttribute('left');
@@ -262,9 +277,7 @@ class Attribute<T> extends Equatable {
     if (identical(this, other)) return true;
     if (other is! Attribute) return false;
     final typedOther = other;
-    return key == typedOther.key &&
-        scope == typedOther.scope &&
-        value == typedOther.value;
+    return key == typedOther.key && scope == typedOther.scope && value == typedOther.value;
   }
 
   // This might not needed anymore because of equatable
@@ -317,29 +330,28 @@ class LinkAttribute extends Attribute<String?> {
 }
 
 class ColorAttribute extends Attribute<String?> {
-  const ColorAttribute(String? val)
-      : super('color', AttributeScope.inline, val);
+  const ColorAttribute(String? val) : super('color', AttributeScope.inline, val);
 }
 
 class BackgroundAttribute extends Attribute<String?> {
-  const BackgroundAttribute(String? val)
-      : super('background', AttributeScope.inline, val);
+  const BackgroundAttribute(String? val) : super('background', AttributeScope.inline, val);
 }
 
 /// This is custom attribute for hint
 class PlaceholderAttribute extends Attribute<bool> {
-  const PlaceholderAttribute()
-      : super('placeholder', AttributeScope.inline, true);
+  const PlaceholderAttribute() : super('placeholder', AttributeScope.inline, true);
 }
 
 class HeaderAttribute extends Attribute<int?> {
-  const HeaderAttribute({int? level})
-      : super('header', AttributeScope.block, level);
+  const HeaderAttribute({int? level}) : super('header', AttributeScope.block, level);
+}
+
+class LineHeightAttribute extends Attribute<double?> {
+  const LineHeightAttribute({double? lineHeight}) : super('line-height', AttributeScope.block, lineHeight);
 }
 
 class IndentAttribute extends Attribute<int?> {
-  const IndentAttribute({int? level})
-      : super('indent', AttributeScope.block, level);
+  const IndentAttribute({int? level}) : super('indent', AttributeScope.block, level);
 }
 
 class AlignAttribute extends Attribute<String?> {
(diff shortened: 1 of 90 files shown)


ellet0:
Thank you for your contributions,

Regarding this [comment](https://github.com/singerdmx/flutter-quill/issues/1956#issuecomment-2197801350), I would make this feature part of `flutter_quill_extensions` as some developers expect this to be compatible with Quill Js, if you decided to keep it part of `flutter_quill` then I would suggest to include a docs comment, for `LineHeightAttribute` class, and for some other usages, like the `showLineHeightButton` and also for the button itself as the developer might use this button in custom Toolbar. 

Other than that, we should be good to go.

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
```diff
@@ -0,0 +1,194 @@
...
+      return attr;
+    }
+    return widget.controller
+            .getSelectionStyle()
+            .attributes[Attribute.lineHeight.key] ??
+        Attribute.lineHeight;
+  }
+
+  String _label(Attribute<dynamic> attribute) {
+    var label = '1.0';
```
ellet0, line 95: We could use the value `Attribute.lineHeightNormal` as default?

lib/src/models/documents/attribute.dart
```diff
@@ -176,6 +181,22 @@ class Attribute<T> extends Equatable {
...
+      LineHeightAttribute(lineHeight: 1.15);
+
+  /// "attributes":{"line-height": 1.5 }
+  static const Attribute<double?> lineHeightOneAndHalf =
+      LineHeightAttribute(lineHeight: 1.5);
+
+  /// "attributes":{"line-height": 2.0 }
+  static const Attribute<double?> lineHeightDouble =
+      LineHeightAttribute(lineHeight: 2);
+
```
ellet0, line 199: We could also make those part of `LineHeightAttribute` instead to make this a bit organized.

lib/src/widgets/others/default_styles.dart
```diff
@@ -245,186 +254,211 @@ class DefaultStyles {
...
-              decoration: TextDecoration.none,
-            ),
-            const VerticalSpacing(8, 0),
-            const VerticalSpacing(0, 0),
-            null),
-        h3: DefaultTextBlockStyle(
+      h1: DefaultTextBlockStyle(
           defaultTextStyle.style.copyWith(
-            fontSize: 24,
+            fontSize: 34,
```
ellet0, line 259: Are these changes related?

lib/src/widgets/others/default_styles.dart
```diff
@@ -245,186 +254,211 @@ class DefaultStyles {
...
             color: defaultTextStyle.style.color,
-            letterSpacing: -0.5,
-            height: 1.083,
+            letterSpacing: -1,
+            height: 1.030,
             fontWeight: FontWeight.bold,
             decoration: TextDecoration.none,
```
ellet0, line 264: Are these changes related?

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
```diff
@@ -0,0 +1,194 @@
...
+          context: context,
+          controller: widget.controller,
+          onPressed: () {
+            throw UnimplementedError('Not implemented yet.');
+          },
+        ),
+      );
+    }
+
+    return MenuAnchor(
```
ellet0, line 136: This is not directly related, though we have an active issue for all the dropdown buttons, it was introduced after moving to `MenuAnchor` as per release notes of Flutter 3.16, the old widget still works though the Flutter team recommends it for Desktop support

- #1697

lib/src/widgets/others/default_styles.dart
CatHood0: Idk why headers has changes when i didn't make it. I'll restore the original values

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
CatHood0: I just add this dropdown by now. I prefer my original implementation

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
CatHood0: Yeah, it's better by that way😯🙌

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
ellet0: > I just add this dropdown by now. I prefer my original implementation

Sure, though the `MenuAnchor` is used by other widgets too, if we couldn't solve the reported issue, we might revert to use `PopupMenuButton` later.

lib/src/models/documents/attribute.dart
CatHood0: You mean, create this part into `LineHeightAttribute` as static const?

lib/src/widgets/toolbar/buttons/select_line_height_dropdown_button.dart
ellet0: > Yeah, it's better by that way😯🙌

Not necessarily give better results, though it makes things easier to track and maintain. If we want to see the usages we can do it easily this way.

lib/src/models/documents/attribute.dart
ellet0: > You mean about create this part into `LineHeightAttribute` as static const, right?

Yes, it's up to you, though.

ellet0:
> >  I would make this feature part of flutter_quill_extensions as some developers expect this to be compatible with Quill Js. 
> 
> I thought about this first before starting to create this feature, however, I don't understand exactly how the attribute could then be applied since I was forced to modify some lines of `TextLine` so that said attribute could be applied. I apologize, but it still seems complicated to move it without having to remove the `TextLine` lines for this attribute. If you have any suggestions, it would be welcome since it seems better to me, as you say, to keep the package as similar to Quill Js and do the rest as extensions of it.

I haven't looked into this, though the easiest solution is to create an interface with a class that implement the default behavior with static instance, another class that implements this interface in the extensions package, the developer would have to call some function to override the default implementation.

We could also simplify the solution or use a similar method.

Though if you prefer not to, you don't have to, docs for all usages of this feature that will be used by the developer would be sufficient.

CatHood0:
This is not related, but, this stacktrace is always throwed after start the app and try to move the cursor at any position (i need more testing to know the cases when this happens):

'package:flutter_quill/src/widgets/quill/text_block.dart':
Failed assertion: line 694 pos 12: 'container.containsOffset(position.offset)'

When the exception was thrown, this was the stack:
#2      RenderEditableTextBlock.globalToLocalPosition (package:flutter_quill/src/widgets/quill/text_block.dart:694:12)
#3      RenderEditor.getLocalRectForCaret (package:flutter_quill/src/widgets/editor/editor.dart:1225:39)
#4      RawEditorStateTextInputClientMixin._updateComposingRectIfNeeded
(package:flutter_quill/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart:94:24)
#5      RawEditorStateTextInputClientMixin.openConnectionIfNeeded
(package:flutter_quill/src/widgets/raw_editor/raw_editor_state_text_input_client_mixin.dart:79:7)
#6      QuillRawEditorState.requestKeyboard (package:flutter_quill/src/widgets/raw_editor/raw_editor_state.dart:1482:7)
#7      QuillRawEditorState._handleFocusChanged (package:flutter_quill/src/widgets/raw_editor/raw_editor_state.dart:1380:7)
#8      ChangeNotifier.notifyListeners (package:flutter/src/foundation/change_notifier.dart:433:24)
#9      FocusNode._notify (package:flutter/src/widgets/focus_manager.dart:1090:5)
#10     FocusManager.applyFocusChangesIfNeeded (package:flutter/src/widgets/focus_manager.dart:1850:12)
(elided 4 frames from class _AssertionError and dart:async)

CatHood0:
> some developers expect this to be compatible with Quill Js

If this could be a problem, we could just force `showLineHeightButton` to be false by default and only those who want this functionality activate it. Not if you like this solution, but I'll take some time to try to move all of this to `flutter_quill_extensions`

ellet0:
> > some developers expect this to be compatible with Quill Js
> 
> If this could be a problem, we could just force `showLineHeightButton` to be false by default and only those who want this functionality must activate it. i don't know what do you think about this, but if you really prefer to move it, then I'll take some time to try to move all of this to `flutter_quill_extensions`

Moving this feature is something optional, the current method is okay, docs and setting `showLineHeightButton` to false by default should be sufficient, though if we keep adding more features in the future that do not confirm or are not compatible with Quill JS, we will move them to `flutter_quill_extensions` later in the future.

singerdmx (approved):

CatHood0:
> I would suggest including a docs comment, for the LineHeightAttribute class, and for some other usages, like the showLineHeightButton and also for the button itself as the developer might use this button in custom Toolbar.

I don't completely understand what you mean by this (sorry). I understand the part about adding more docs comments (although I honestly wouldn't know how to start with this part since many buttons do not contain docs comments, which doesn't give me many examples about this) to the button. But not about adding docs comments to `showLineHeightButton` since it seems to me that by itself it is understood what this same variable does (although maybe I misunderstood this part)

ellet0:
There is no need to be sorry about it.

The project exposes the buttons so the developers can develop their own toolbar, take a look at [Flutter Quill Custom Toolbar](https://github.com/singerdmx/flutter-quill/blob/master/doc%2Fcustom_toolbar.md).

The developer might find this button and use it without noticing that this button use a feature that is not compatible with Quill JS delta in case they expect or want the data to be compatible.

The other buttons could use docs comments, though it's not required. If it's clear, there is no direct way to know if this feature is not compatible with Quill JS from the developer perspective.

A simple docs comments will do. This way, the developer will know this before enabling this feature.

CatHood0:
I already added the warning comments for possible conflicts with the original Quill Js API. I would have liked to add tests for this new feature, but I still don't understand testing with flutter_quill very well. I'll probably try to add the test soon

ellet0:
> I already added the warning comments for possible conflicts with the original Quill Js API. I would have liked to add tests for this new feature, but I still don't understand testing with flutter_quill very well. I'll probably try to add the test soon

The project wasn't designed from the ground to prioritize tests. The testing process is manual, and that's something we could improve, though I'm unfamiliar with the source code that has not been updated much since 3 years (e.g., Raw editor, Rules, Delta and other internal Apis)

I only splitted the code a bit for organization.

CatHood0:
> The project wasn't designed from the ground to prioritize tests. The testing process is manu

Then i won't add it by now. When it be necessary i'll create it

lib/src/models/documents/attribute.dart
```diff
@@ -337,6 +342,27 @@ class HeaderAttribute extends Attribute<int?> {
       : super('header', AttributeScope.block, level);
 }
 
+/// This attribute represents the space between text lines. The line height can be
+/// adjusted using predefined constants or custom values
+///
+/// The attribute at the json looks like: "attributes":{"line-height": 1.5 }
+class LineHeightAttribute extends Attribute<double?> {
```
EchoEllet, line 349: I moved this class to `custom_attributes.dart` without breaking the change in #2032.

To make it a bit more readable and in case we add more attributes that are not available as a standard feature in Quill JS.

CatHood0: I think it's great. I'll triedto move to `flutter_quill_extensions` but i couldn't import any file from this last one. Thanks for make this change!

EchoEllet: > I tried to move to `flutter_quill_extensions` but i couldn't import any file from this last one. Thanks for make this change!

This file is already exported, if it wasn't we can import the file manually from `src` in `flutter_quill_extensions` and ignore the warning since if there are any internal breaking changes we can handle it, and usually we use the same version for all packages to ensure compatibility. We could also export it only in 
`package:flutter_quill/extensions.dart`

CherishMvp:
How should I set the line-height when initializing globally?

PR #1976, opened 2024-07-05
fix: context menu behavior in mouse, desktop env

### LINKED ISSUE #1742: Disable context menu after selection for desktop platform

### The question

How could I change default behaviour of context menu showing after selection? On desktop platforms context menu is showing after user selects the text, selects all with double click etc. I think this is not normal behaviour on desktop and all of our users think that this is weird.

How can I make it that for desktop platform context menu for selection will show only when right clicking the selection?

n7484443:
## Description
When drag or select text with the mouse in a desktop environment, a context menu always appears.
The typical behavior in a desktop environment is that it does not appear when dragging/selecting, but when additionally press the right-click button.

## Related Issues
- *Fix #1742*

## Checklist

- [X] I read the [Contributor Guide](../CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [X] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [X] I did not modify the `CHANGELOG.md` nor the package version in `pubspec.yaml` files.
- [X] All existing and new tests are passing.
- [X] I have run the commands in `./scripts/before_push.sh` and it all passed successfully

## Breaking Change

- [X] No, this is *not* a breaking change.


lib/src/widgets/others/delegate.dart
```diff
@@ -167,7 +169,7 @@ class EditorTextSelectionGestureDetectorBuilder {
       null,
       SelectionChangedCause.forcePress,
     );
-    if (shouldShowSelectionToolbar) {
+    if (shouldShowSelectionToolbar && !requiresAdditionalActionForToolbar) {
```
ellet0, line 172: We might consider extractring this check into its own function or getter as it's become a bit harder to maintain. In the future, someone might add new features or bugs to fix and update some of the checks and forget the others.

And we also don't have tests to cover those cases.

Issue #1714, opened 2024-01-31
[Web] Any way to disable the clipboard share permission request?


c-seeger:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### The question

When using web by default quill shows a permission notification asking for "See text and images copied to the clipboard". Is there any way to disable this feature completely? I don't want to bother my users with this request at all.

darwin-morocho:
@singerdmx

darwin-morocho:
@c-seeger maybe this can help if you only want to show your delta list

```dart
import 'package:flutter/material.dart';
import 'package:flutter_quill/markdown_quill.dart';
import 'package:flutter_quill/quill_delta.dart';
import 'package:markdown_widget/markdown_widget.dart';

class QuillViewer extends StatelessWidget {
  const QuillViewer({
    super.key,
    required this.data,
  });
  final List data;

  @override
  Widget build(BuildContext context) {
    return MarkdownBlock(
      data: DeltaToMarkdown().convert(
        Delta.fromJson(data),
      ),
    );
  }
}
```

IWantToKnowAboutThis:
I want to use readOnly mode.
I don't know why I keep trying Clipboard share permissions requests on my web browser. 
Does anyone know how to fix this?
😢

PR #2009, opened 2024-07-13
feat(web): rich text paste from Clipboard using HTML

### LINKED ISSUE #1998: [Web] Pasting HTML Doesn't Work Propertly

### Flutter Quill version

9.5.13

### Steps to reproduce

Copy HTML with bullets on web, paste.

A few problems here:

First off the browser won't actually let non plain text formats through using the clipboard API, so the clipboard service always thinks there is no html on the clipboard. On web, non text formats are only available from the paste event.

Steps to fix:

1) Handle "paste" event on window instead of using standard paste hotkey logic (note that the same path is also needed to paste things like images from the clipboard on web). If a quill node is focused, handle the paste event and insert the data into the quill editor. The paste event will contain, files, and multiple clipboard formats that are unavailable to the browser clipboard API.
2) If button action is pressed and initiates a paste, unfortunately, the only way to get the data is using a browser extension (this is what google docs does). Until there is a browser extension for flutter quill, quill should do what google docs does when you don't have the browser extension installed, which is display a dialog when pressing the paste button that explains the limitation and shows the hotkey they need to press in order to paste. Once there is an extension, this dialog can provide a link to install the browser extension.

### Expected results

Bullets are preserved

### Actual results

Bullets are not preserved

### Code sample

```dart
[Paste your code here]
```

</details>

### Screenshots or Video

<details>

[Upload media here]

</details>

### Logs

```console
[Paste your logs here]
```

</details>

EchoEllet:
## Description

*Support Rich Text Pasting with HTML on the Web.*

Importing any file from [`web`](https://pub.dev/packages/web) directly will cause a compile issue on non-web platforms even if we're using `kIsWeb` to check the platform, so I used conditional import by importing a file that mocks the implementation for non-web platforms, for web will import the real implementation that is intended for web and use the `web` package.

This process is error-prone and can cause issues when moving the files and refactoring the imports since the IDE doesn't update the conditional import path and the compiler only causes issues when compiling to the web, which is why the GitHub workflow build the example app for both Web and Linux (a non-web platform).

Also the `_pasteMarkdown` and `_pasteHTML` are no longer accessible to `quill_controller_web_stub.dart` as it's in a separate file, I didn't make them public since they are intended for internal use only, so I moved them into a separate file that's not exported, I have noticed we're making some APIs public because we need them somewhere else internally, making more breaking changes when there is a need.

The deprecated APIs should be in their own separate file since we can move them somewhere else and export the file and still be backward compatible, maybe we need a new directory such as `deprecated` at some point

## Related Issues

- *Partial Fix #1998*
- *Related #1843*
- *Related #2001*
- *Related #1714*
- *Related #1719*
- *Related #1722*

Previously, some assertions were failing, some of them were not caused by this feature but related to it.

## Suggestions

- Remove the `meta` package from `flutter_quill` since it's already available in Flutter and exposed as a public API, it's required in `dart_quill_delta`
- We should probably deprecate the experimental function `QuillController.setContents`, it's misleading and doesn't work the same as Quill JS. It's not being used internally.
- `disableClipboard` is not quite a descriptive name or I might have not understood it correctly, it says:

  > this disables the clipboard notification for requesting permissions

 ```dart
   if (!widget.configurations.disableClipboard) {
      // Web - esp Safari Mac/iOS has security measures in place that restrict
      // cliboard status checks w/o direct user interaction. Initializing the
      // ClipboardStatusNotifier with a default value of unknown will cause the
      // clipboard status to be checked w/o user interaction which fails. Default
      // to pasteable for web.
      if (kIsWeb) {
        _clipboardStatus = ClipboardStatusNotifier(
          value: ClipboardStatus.pasteable,
        );
      }
    }
  ```
  I'm not sure if it disables all Clipboard features including the `Clipboard` from Flutter or only the Rich text pasting feature.  See related PR #1719.

## Checklist

## Breaking Change

### DIFF
--- a/lib/src/controller/quill_controller.dart
+++ b/lib/src/controller/quill_controller.dart
@@ -1,8 +1,8 @@
 import 'dart:math' as math;
 
+import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter/services.dart' show ClipboardData, Clipboard;
 import 'package:flutter/widgets.dart';
-import 'package:html/parser.dart' as html_parser;
 import 'package:meta/meta.dart' show experimental;
 
 import '../../quill_delta.dart';
@@ -10,7 +10,6 @@ import '../common/structs/image_url.dart';
 import '../common/structs/offset_value.dart';
 import '../common/utils/embeds.dart';
 import '../delta/delta_diff.dart';
-import '../delta/delta_x.dart';
 import '../document/attribute.dart';
 import '../document/document.dart';
 import '../document/nodes/embeddable.dart';
@@ -18,9 +17,12 @@ import '../document/nodes/leaf.dart';
 import '../document/structs/doc_change.dart';
 import '../document/style.dart';
 import '../editor/config/editor_configurations.dart';
-import '../editor_toolbar_controller_shared/clipboard/clipboard_service_provider.dart';
 import '../toolbar/config/simple_toolbar_configurations.dart';
 import 'quill_controller_configurations.dart';
+import 'quill_controller_rich_paste.dart';
+
+import 'web/quill_controller_web_stub.dart'
+    if (dart.library.html) 'web/quill_controller_web_real.dart';
 
 typedef ReplaceTextCallback = bool Function(int index, int len, Object? data);
 typedef DeleteCallback = void Function(int cursorPosition, bool forward);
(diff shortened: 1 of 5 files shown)


singerdmx (approved):

AtlasAutocode (approved):
I don't see any obvious problems.

EchoEllet:
I was already able to reproduce this issue at the time of this PR, and I wasn't able to again after the last update, I'm not quite sure what I did but I'm able to reproduce it once again:

Text pasted twice

The text is pasted twice, with and without rich text paste feature when there is an HTML in the clipboard content.

PR #2023, opened 2024-07-16
Fix(rule): PreserveInlineStyleRule assume the type of the operation data and throw stacktrace

CatHood0:
## Description

I understand if you take this as something that normally wouldn't happen, but in the event that some developer creates their own `CustomEmbedBlock`, these types of errors are better to avoid from the beginning.

In this case, I have a video that explains a little better how I reproduced this error (it is not easy to reproduce. I had to record several times until I finally managed it)

https://github.com/user-attachments/assets/8902f1c8-d5d9-445b-88ed-9a066f8bc518

## Checklist

## Breaking Change

### DIFF
--- a/lib/src/models/rules/insert.dart
+++ b/lib/src/models/rules/insert.dart
@@ -568,7 +568,8 @@ class PreserveInlineStylesRule extends InsertRule {
       if (prevData.endsWith('\n')) {
         /// If current line is empty get attributes from a prior line
         final currLine = itr.next();
-        final currData = currLine.data is String ? currLine.data as String : null;
+        final currData =
+            currLine.data is String ? currLine.data as String : null;
         if (currData != null && (currData.isEmpty || currData[0] == '\n')) {
           if (prevData.trimRight().isEmpty) {
             final back =
--- a/lib/src/models/rules/insert.dart
+++ b/lib/src/models/rules/insert.dart
@@ -568,17 +568,19 @@ class PreserveInlineStylesRule extends InsertRule {
       if (prevData.endsWith('\n')) {
         /// If current line is empty get attributes from a prior line
         final currLine = itr.next();
-        final currData = currLine.data as String?;
-        if (currData != null && (currData.isEmpty || currData[0] == '\n')) {
-          if (prevData.trimRight().isEmpty) {
-            final back =
-                DeltaIterator(documentDelta).skip(index - prevData.length);
-            if (back != null && back.data is String) {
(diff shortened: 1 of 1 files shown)


AtlasAutocode:
Thanks for the catch: I'm used to other languages that have a conditional cast operation without throwing an exception.
It appears that currLine.data is a Map<>

L571 change
> final currData = currLine.data as String?;

to:
> final currData = currLine.data is String ? currLine.data as String : null;

Can you test this change and include in your PR?
Let me know if it doesn't fix the exception.

CatHood0:
> Can you test this change and include in your PR? 

I'll try it

CatHood0:
I see the line L572 that has a conditional
> `if (currData != null && (currData.isEmpty || currData[0] == '\n'))`

Is not better use `currData.startWith('\n')` instead accessing to the character index?

CatHood0:
Oh god, i miss this

AtlasAutocode:
> Isn't it better to use currData.startWith('\n') instead of accessing the character index?

Sounds good.
Since I program in multiple languages, I have had problems with situations where a string is empty and tend to be cautious.

You could probably use:

> if ( currData?.startsWith('\n') == true )

I don't know whether it is better to be 'compact' or 'explicit'.
Reading code (and understanding it) 6 months after writing code is my usual state of affairs.

CatHood0:
> if ( currData?.startsWith('\n') == true )

Use `== true` with a function that return a bool is redundant, at least on Dart. 

> Since I program in multiple languages, I have had problems with situations where a string is empty and tend to be cautious.

Yeah, i understand you

AtlasAutocode:
> Use == true with a function that return a bool is redundant, at least on Dart.

currData can be null so it is needed.

> if (currData != null && (currData.isEmpty || currData.startsWith(RegExp('^\n')))) {

why not use the simple single statement to combine all 3 conditions: if ( currData?.startsWith('\n') == true )

Also, I do not like using RegExp for simple cases.
Reason == Performance.
Testing for null, empty and starts with NL is fast. Using a regular expression is time costly. (This is why I used [0] which is fast)
Maybe it doesn't matter in this case, but I usually deal with situations where a program with RexEx would take an hour to run but the simpler method takes 5 minutes to run.
Programmers love RegExp - users want fast.

CatHood0:
> why not use the simple single statement to combine all 3 conditions: if ( currData?.startsWith('\n') == true )

You have a point. I will change it

PR #2028, opened 2024-07-16
 Fix(rule): PreserveInlineStyleRule assume the type of the operation data and throw stacktrace

CatHood0:
## Description

I understand if you take this as something that normally wouldn't happen, but in the event that some developer creates their own `CustomEmbedBlock`, these types of errors are better to avoid from the beginning.

In this case, I have a video that explains a little better how I reproduced this error (it is not easy to reproduce. I had to record several times until I finally managed it)

_Yeah, the PR was repeated because I accidentally deleted the branch_

https://github.com/user-attachments/assets/8902f1c8-d5d9-445b-88ed-9a066f8bc518

## Checklist

## Breaking Change

### DIFF
--- a/lib/src/models/rules/insert.dart
+++ b/lib/src/models/rules/insert.dart
@@ -568,7 +568,8 @@ class PreserveInlineStylesRule extends InsertRule {
       if (prevData.endsWith('\n')) {
         /// If current line is empty get attributes from a prior line
         final currLine = itr.next();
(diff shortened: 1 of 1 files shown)


CatHood0:
@AtlasAutocode I see the line L572 that has a conditional
> `if (currData != null && (currData.isEmpty || currData[0] == '\n'))`

Isn't it better to use `currData.startWith('\n')` instead of accessing the character index?

PR #2040, opened 2024-07-19
Have the ability to create custom rules, closes #1162

### LINKED ISSUE #1162: Override default rules to remove some rules
I can see that the `setCustomRules` function can be used to append a list of custom rules. However, what am i supposed to do if I want to remove some of the default rules to match my use case. Is there a way to do this within the current build?

Guillergood:
## Description

Have the ability to create custom rules, right now we do not have a way to create a new class to put in ``setCustomRules``, the closest thing is:

``class QuillCustomDeleteRule extends InsertRule`` 

which is not ideal, so why not import all rules folder?

## Related Issues

It fixes #1162

## Type of Change


lib/extensions.dart
```diff
@@ -4,4 +4,7 @@ export 'src/common/utils/platform.dart';
 export 'src/common/utils/string.dart';
 export 'src/common/utils/widgets.dart';
 export 'src/document/nodes/leaf.dart';
+export 'src/rules/delete.dart';
+export 'src/rules/format.dart';
 export 'src/rules/insert.dart';
+export 'src/rules/rule.dart';
```
EchoEllet, line 10: Those 3 files weren't exported by default, there might be a reason why only one of them is exported or they might have some functions or classes that meant to be used internally, exposing public API for the existing code that's avaliable internally can lead to breaking change if we want to fix or change something later.

Guillergood: I understand, but then I cannot see another option if we want to create custom rules...

EchoEllet: I'm unfamiliar with this code, though certain parts might not be intended to be public API.

In case there is an issue, we can deprecate it, and then avoid exposing it within the next breaking change release.

PR #2049, opened 2024-07-22
if magnifier removes toolbar, restore it when it is hidden

### LINKED ISSUE #2046: Selection menu doesn't display after drag

### Flutter Quill version

10.0.2

### Steps to reproduce

1. Create a note with lots of text.
2. Long press a word 
3. Drag one of the selection handles
4. End the drag

### Expected results

I'm expecting the selection menu to reappear after the drag is finished.

### Actual results

The selection menu does not reappear.  So after selecting text, no way to cut/copy.

### Code sample

_No response_

### Additional Context

_No response_

mtallenca:
## Description

Magnifier is removing selection context menu, but not restoring the menu after it's dismissed. This PR fixes that bug

- *Fix #2046*

## Type of Change

### DIFF
--- a/lib/src/editor/widgets/text/text_selection.dart
+++ b/lib/src/editor/widgets/text/text_selection.dart
@@ -181,6 +181,7 @@ class EditorTextSelectionOverlay {
 
   /// A copy/paste toolbar.
   OverlayEntry? toolbar;
+  bool _restoreToolbar = false;
 
   TextSelection get _selection => value.selection;
 
@@ -411,8 +412,12 @@ class EditorTextSelectionOverlay {
   void _showMagnifier(MagnifierInfo initialMagnifierInfo) {
     // 隐藏toolbar
     if (toolbar != null) {
+      _restoreToolbar = true;
       hideToolbar();
+    } else {
+      _restoreToolbar = false;
     }
+
     // 更新 magnifierInfo
     _magnifierInfo.value = initialMagnifierInfo;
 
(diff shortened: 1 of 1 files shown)


lib/src/editor/widgets/text/text_selection.dart
```diff
@@ -181,6 +181,7 @@ class EditorTextSelectionOverlay {
 
   /// A copy/paste toolbar.
   OverlayEntry? toolbar;
+  bool _restoreToolbar = false;
```
EchoEllet, line 184: Is `_restoreToolbar` something specific to the magnifier or something only needed to fix an issue related to the magnifier and is not meant to be used outside of this feature?

EchoEllet (commented):
@mtallenca Thank you for the quick fix, working on moving this feature into it's own separate file so more details would be helpful to address it correctly.

lib/src/editor/widgets/text/text_selection.dart
EchoEllet: If yes, we probably need a better way to document it or maybe more specific name:

```dart
/// Restore the selection context menu after the magnifier is dismissed.
bool _restoreToolbar = false;
```

Do you have suggestions?

mtallenca:
This is a flag only used by the magnifier.  If the magnifier is going to
hide the toolbar, it needs to restore the toolbar when dismissed.  A better
name would be _restoreToolbarAfterMagnifier

EchoEllet:
Thank you for confirming.

PR #2060, opened 2024-07-24
Fix: RTL issues

### LINKED ISSUE #1928: problems when using languages ​​that use ltr instead of rtl

### Flutter Quill version

9.4.2

### Steps to reproduce

problem with languages ​​that use ltr instead of rtl,   DirectionAttribute('ltr');
When I touch the enumerated list or checkbox option, even when using ltr in languages ​​like Hebrew, the checkboxes and enumerated lists do not change, they remain in the position as if using rtl

### Expected results

checkbox and the list number go to the beginning of the line on the right side

### Actual results

checkbox and list number are at the beginning of the line on the left side

### Code sample

default quill

### Screenshots or Video

_No response_

### Logs

_No response_

CatHood0:
## Description

As we know, we always have issues related with the RTL directionality. At this moment, the current implementation of List, Checkbox or Blockquote, force the direction to be LTR, even if the system language is a RTL. 

_To test this i use some languages as **hebrew** or **arabic** that are RTL (as far i know)_

### Before these changes

Before 1 Before 2(1)

### After these changes

After1(1) After2 After3

## Related Issues

*Fix #1928*

## Suggestions

We really **need** to avoid harcode the widgets on the editor. 

I'll try to separate the logic to another file and make more extensible soon

### DIFF
--- a/lib/src/common/utils/directionality.dart
+++ b/lib/src/common/utils/directionality.dart
@@ -1,0 +1,25 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+
+bool isRTLLanguage(Locale locale) {
+  final rtlLanguages = [
+    'ar', // Arabic
+    'dv', // Divehi (Maldivian)
+    'fa', // Persian (Farsi)
+    'ha', // Hausa (in Ajami script)
+    'he', // Hebrew
+    'khw', // Khowar (in Arabic script)
+    'ks', // Kashmiri (in Arabic script)
+    'ku', // Kurdish (Sorani)
+    'ps', // Pashto
+    'ur', // Urdu
+    'yi', // Yiddish
+    'sd', // Sindhi
+    'ug', // Uyghur
+  ];
+  return rtlLanguages.contains(locale.languageCode);
+}
+
+bool isRTL(BuildContext context) {
+  return isRTLLanguage(Localizations.localeOf(context));
+}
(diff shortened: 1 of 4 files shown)


lib/src/common/utils/directionality.dart
```diff
@@ -0,0 +1,25 @@
...
+    'yi', // Yiddish
+    'sd', // Sindhi
+    'ug', // Uyghur
+  ];
+  return rtlLanguages.contains(locale.languageCode);
+}
+
+bool isRTL(BuildContext context) {
+  return isRTLLanguage(Localizations.localeOf(context));
+}
```
EchoEllet, line 25: I'm not sure if there is a reason or if we're already using `Directionality.of(context)` instead?

CatHood0: Can you explain more about it? I'm using these functions to verify first if the language of the system is a RTL language. Does `Directionality.of(context)` already do this?

EchoEllet: It does and it's compatible with Flutter, it supports the [Directionality](https://api.flutter.dev/flutter/widgets/Directionality-class.html) widget and should work in case the user changes the local or the direction

Eng-Mohamed-Elsayed:
ok how can add button ltr or rtl

jonasbernardo:
It doesn't work, watch the video, its ontap only works as if it were in the ltr position

https://github.com/user-attachments/assets/ccd43be3-9977-48ee-b057-5ff0580147fa

Issue #2065, opened 2024-07-25
toPlainText method from any Embeddable return text that cannot be used outside of the editor


CatHood0:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### The question

I watch this code on [Line](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/document/nodes/line.dart) class:
```dart
  int _getNodeText(Leaf node, StringBuffer buffer, int offset, int remaining) {
    final text = node.toPlainText();
    if (text == Embed.kObjectReplacementCharacter) {
      buffer.write(Embed.kObjectReplacementCharacter); 
      return remaining - node.length;
    }

    final end = math.min(offset + remaining, text.length);
    buffer.write(text.substring(offset, end));
    return remaining - (end - offset);
  }
```

It's really useful the `Embed` can be copied and paste multiple times, but, it's difficult if you have just a `CustomBlockEmbed` with just plainText and this text is not copied on the clipboard as plainText. I don't know if i already fully explain what is exactly the issue, then you can see this #2050 to undertand better this.

If you try to copy any `Embed` , you can see the text is just empty even the `Embed` contains data as `String` or the `toPlainText` from the `EmbedBuilder` is overrided. I fixed this partially adding this part of code:

```dart
    if (text == Embed.kObjectReplacementCharacter) {
      final nodeValue = node.value;
      if (nodeValue is Embeddable && !['video','image','custom','formula','table'].contains(nodeValue.type) && nodeValue.data != null && nodeValue.data is String) {
        buffer.write(nodeValue.data);
      } else {
        buffer.write(Embed.kObjectReplacementCharacter);
      }
      return remaining - node.length;
    }
```

But i prefer the root error. Could anyone suggest where should i take a look?

CatHood0:
@EchoEllet  are you familiar with these parts of code?

EchoEllet:
> @EchoEllet  are you familiar with these parts of code?

I didn't work on them, and I was planning to before. I moved some parts to different files for organization.

CatHood0:
> I didn't work on them, and I was planning to before

I fix partially to copy just the plain text for custom embeds, but, this could break the feature that let the editor paste the Embeds to the editor

singerdmx:
`toPlainText` is more of debugging purpose like dumping the log

singerdmx:
[...]

PR #2067, opened 2024-07-25
Feat: support for customize copy and cut Embeddables to Clipboard

### LINKED ISSUE #2050: How to copy text from a custom widget

### The question

How to copy text from a custom widget

CatHood0:
## Description

Before this PR, we always copy or cut an `Embeddable` with a empty text, making also more easy to getting back a `Embed` when the user paste it, but, this could be an issue for users that only want to get the plain text from a `Embed`. To fix this i add some classes to override this behavior.

### `CopyCutService` and `CopyCutServiceProvider`

`CopyCutService` is an interface that let us modify the behavior of `getCopyCutAction` that let us return some text or an object that will be copied to the `Clipboard`. 

It looks like:

```dart
typedef CopyCutAction = Object? Function(dynamic data);

/// An abstraction to make it easy to provide different implementations
/// For copy or cut actions from a Line (just for embeddable blocks)
@immutable
abstract class CopyCutService {
  /// Get the CopyCutAction by the type
  /// of the embeddable (this type is decided by
  /// the property type of that class)
  CopyCutAction? getCopyCutAction(String type);
}
```

`CopyCutServiceProvider` it's very similiar to `ClipboardServiceProvider`. Just provide an instance to be modified and used by any resource of the project.

It looks like:
```dart
class CopyCutServiceProvider {
  const CopyCutServiceProvider._();
  static CopyCutService _instance = DefaultCopyCutService();

  static CopyCutService get instance => _instance;

  static void setInstance(CopyCutService service) {
    _instance = service;
  }

  static void setInstanceToDefault() {
    _instance = DefaultCopyCutService();
  }
}
```

> [!NOTE]
> By default `CopyCutServiceProvider` use `DefaultCopyCutService` that just return an empty text.
> 
> It looks like:
> ```dart
> class DefaultCopyCutService extends CopyCutService {
>   @override
>   CopyCutAction? getCopyCutAction(String type) {
>     return (data) => Embed.kObjectReplacementCharacter;
>   }
> }
> ```

### Example

If we have a `Embeddable` like:

```dart
class ContentTagEmbed extends Embeddable {
  const ContentTagEmbed(
    String value,
  ) : super(tagType, value);

  static const String tagType = 'content_tag';

  static ContentTagEmbed fromDocument(Document document) => ContentTagEmbed(jsonEncode(document.toDelta().toJson()));

  Document get document => Document.fromJson(jsonDecode(data));
}
```

To detect this `Embed`, we just need to create a custom implementation of `CopyCutService`

```dart
class CopyCutServiceTagContent extends CopyCutService {
  
 @override
  CopyCutPlainText? getCopyCutAction(String type) {
    if (type == 'content_tag') {
      return (data) => data.toString();
    } 
    // returning this String maintain the default behavior for others Embed that doesn't need to be plain
    return (data) => '\uFFFC';
  }
}
```

And after we just need to set this instance to the `CopyCutServiceProvider`:

```dart
CopyCutServiceProvider.setInstance(CopyCutServiceTagContent());
```

https://github.com/user-attachments/assets/6795e59d-463f-42bd-9b0e-99487af6559f

## Related Issues

- *Related #2050*
- *Related #2065*


singerdmx:
Nice

lib/src/editor_toolbar_controller_shared/copy_cut_service/copy_cut_service.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/foundation.dart';
+
+typedef CopyCutAction = Object? Function(dynamic data);
```
EchoEllet, line 3: Can we use `Object?` instead of `dynamic`?

CatHood0: I use dynamic to avoid transforming the data from Embeddable objects. But, i'll change it soon to avoid unexpected behaviors

PR #2078, opened 2024-07-30
Move Controller outside of configurations data class

AtlasAutocode:
#Reference:  https://github.com/singerdmx/flutter-quill/discussions/2037

The presence of the controller in the configuration classes prevents use of const.
This PR is a non-breaking first step in achieving this.
Configuration class usage has been reorganized to deprecate the controller member and enable the controller to be set directly in the editor and toolbar. In the future, the deprecated code will be removed - that will be a breaking change.

Current code should work without change.
**Recommendation:** Move setting the controller from configuration classes and set directly in the editor and toolbar.

**Future:** controller will be removed from the configuration classes.

**Current code**
> final _controller = QuillController.basic();
QuillSimpleToolbar(
    configurations: **QuillSimpleToolbarConfigurations(controller: _controller,**)),

> QuillEditor.basic(
    configurations: **QuillEditorConfigurations(controller: _controller,**),
),

**Recommended** (Move controller up a line)
> final _controller = QuillController.basic();
**QuillSimpleToolbar(controller: _controller,**
    configurations: const QuillSimpleToolbarConfigurations()
)),

> **QuillEditor.basic(controller: _controller,**
    configurations: const QuillEditorConfigurations(),
),

## Type of Change

- [ X ] 🧹 **Code refactor:** Code restructuring that does not affect behavior.
- [ X ] 🧪 **Tests:** Adds new tests or modifies existing tests.
- [ X ] 📝 **Documentation:** Updates or additions to documentation.


lib/src/editor/config/editor_configurations.dart
```diff
@@ -97,7 +100,8 @@ class QuillEditorConfigurations extends Equatable {
   /// by any shortcut or keyboard operation. The text is still selectable.
   ///
   /// Defaults to `false`. Must not be `null`.
-  bool get readOnly => controller.readOnly;
+  // ignore: deprecated_member_use_from_same_package
+  bool get readOnly => controller?.readOnly != false;
```
EchoEllet, line 104: We have missed something here: The readOnly property will be from the `QuillController` (the one that's deprecated and no longer used)

lib/src/toolbar/simple_toolbar.dart
```diff
@@ -9,13 +11,37 @@ import 'simple_toolbar_provider.dart';
...
+  }) {
+    // ignore: deprecated_member_use_from_same_package
+    controller ??= configurations?.controller;
+    assert(controller != null,
+        'controller required. Provide controller directly (preferred) or indirectly through configurations (not recommended - will be removed in future versions).');
+    controller ??= QuillController(
+        document: Document(),
+        selection: const TextSelection.collapsed(offset: 0));
+    //
+    controller.toolbarConfigurations = configurations;
```
EchoEllet, line 27: We should not store the `QuillSimpleToolbar` configuration or anything related in the `QuillController`. Our toolbar design is poor and the `QuillController` should be more extensible and less specific to it to allow developers to design their toolbar.

I will remove it in `11.0.0` since it's confusing.

PR #2089, opened 2024-08-02
Feature/allow shortcut override

InstrinsicAutomations:
## Description

This change modifies the merge priority of shortcut actions with `widget.configurations.customShortcuts`. Specifically they allow the programmer's own defined shortcuts to take priority when activation triggers are the same. We had some use cases where we wanted to disable or completely change the default keyboard shortcuts. This change allowed us to do that while preserving the original behavior if untouched.

## Related Issues

While there were some tickets that only seemed tangential to this feature, none were a good fit.

## Type of Change


chadtomkiss:
@InstrinsicAutomations @singerdmx does this change work as intended? I'm finding the second Map of `mergeMaps` is the one that's taking priority and so isn't letting me override Quill shortcuts.

For `actions`, custom actions are passed as the second argument to `mergeMaps` 🤔

PR #2090, opened 2024-08-04
Enhancement: Search within Embed objects

### LINKED ISSUE #2075: Custom content cannot be searched.

### The question

352365054-713b593a-caaf-4e71-aaba-67a30c6490b1

AtlasAutocode:
## Description

Adds the ability to search within custom Embed objects.

For how to enable: refer to documentation, Configurations, Links, Search

https://github.com/singerdmx/flutter-quill/blob/master/doc/configurations/search.md

## Related Issues

- *Fix #2075*

## Type of Change

- [ X ] ✨ **New feature:** Adds new functionality without breaking existing features.
- [ X ] 🧪 **Tests:** Adds new tests or modifies existing tests.
- [ X ] 📝 **Documentation:** Updates or additions to documentation.


lib/src/document/document.dart
```diff
@@ -250,10 +241,23 @@ class Document {
...
+  /// Caches configuration set in QuillController.
+  /// Allows access to embedBuilders and search configurations
+  QuillEditorConfigurations? _editorConfigurations;
+  QuillEditorConfigurations get editorConfigurations =>
+      _editorConfigurations ?? const QuillEditorConfigurations();
+  set editorConfigurations(QuillEditorConfigurations? value) =>
+      _editorConfigurations = value;
+  QuillSearchConfigurations get searchConfigurations =>
+      editorConfigurations.searchConfigurations;
+
```
EchoEllet, line 255: I'm uncertain if storing the editor configuration in the document and controller is a good idea, I don't have a better solution as I have questions. Let's keep it but keep it private and internal API so we can fix things later with fewer breaking changes, also document the places where you added the config parameter in the `line.dart` file (e.g., `_getNodeText()`). 

Some methods like `getPlainText()`:

```dart
/// Returns plain text within the specified text range.
  String getPlainText(int offset, int len, [QuillEditorConfig? config]) {
    final plainText = StringBuffer();
    _getPlainText(offset, len, plainText, config);
    return plainText.toString();
  }
```

Could accept `embedBuilders` and `unknownEmbedBuilder` instead to keep it a bit clearer in case someone is using that method so they don't have to create and pass the full config of the editor.

PR #2118, opened 2024-08-17
Feat: Spellchecker for Flutter Quill

CatHood0:
## Description

For many versions I have been wondering how we could add spell-checking to the project or when it would even be implemented. However, now here is a feature that surely more than one of the thousands who use the package have wanted for their project, but that, unfortunately, had not been easy to implement.

Thanks to the [simple_spell_checker](https://pub.dev/packages/simple_spell_checker) package, getting which words are not spelled correctly from those that are was never so easy and fast.

Now, in order to activate this Feature we would have to take into account the existence of the `SpellcheckerService` class which allows us to create any custom implementation for our editor to directly use without any cover-ups.

```dart
abstract class SpellcheckerService {
  SpellcheckerService({required this.language});

  final String language;

  /// dispose all the resources used for SpellcheckerService
  ///
  /// if [onlyPartial] is true just dispose a part of the SpellcheckerService
  /// (this comes from the implementation)
  ///
  /// if [onlyPartial] is false dispose all resources
  void dispose({bool onlyPartial = false});

  /// Facilitates a spell check request.
  ///
  /// Returns a [List<TextSpan>] with all misspelled words divide from the right words.
  List<TextSpan>? fetchSpellchecker(String text,
      {LongPressGestureRecognizer Function(String)?
          customLongPressRecognizerOnWrongSpan});
}
```

By default `DefaultSpellcheckerService` is the one that the editor uses, which as we can see, has no function because it directly returns a null value.

```dart
class DefaultSpellcheckerService extends SpellcheckerService {
  DefaultSpellcheckerService() : super(language: 'en');

  @override
  void dispose({bool onlyPartial = false}) {}

  @override
  List<TextSpan>? fetchSpellchecker(String text,
      {LongPressGestureRecognizer Function(String p1)?
          customLongPressRecognizerOnWrongSpan}) {
    return null;
  }
}
```

And now, if we want to activate this new functionality that we added, we would have to use `SimpleSpellCheckerServiceImpl` which contains the implementation that we need to be able to activate spell-checking as if we were in Word or LibreOffice (obviously I'm exaggerating)

```dart
class SimpleSpellCheckerImpl extends SpellcheckerService {
  SimpleSpellCheckerImpl({required super.language})
      : checker = SimpleSpellChecker(
          language: language,
          safeDictionaryLoad: true,
        );

  /// [SimpleSpellChecker] comes from the package [simple_spell_checker]
  /// that give us all necessary methods for get our spans with highlighting
  /// where needed
  final SimpleSpellChecker checker;

  @override
  List<TextSpan>? fetchSpellchecker(
    String text, {
    LongPressGestureRecognizer Function(String word)?
        customLongPressRecognizerOnWrongSpan,
  }) {
    return checker.check(
      text,
      customLongPressRecognizerOnWrongSpan:
          customLongPressRecognizerOnWrongSpan,
    );
  }

  @override
  void dispose({bool onlyPartial = false}) {
    if (onlyPartial) {
      checker.disposeControllers();
      return;
    }
    checker.dispose();
  }
}
```

In order to make our instance valid, we have to use `SpellcheckerServiceProvider` with the setInstance() method
```dart
SpellcheckerServiceProvider.setInstance(SimpleSpellCheckerServiceImpl(language: '<your_language>'));
```
And voilá, we will have at our disposal a powerful tool that will be useful for those who write novels, notes, or any type of document that requires correct spelling and grammar.

> [!NOTE]
> It should be noted that the `dispose()` method of `SpellcheckerService` is not just for show. This method, at least for `SimpleSpellCheckerServiceImpl`, is responsible for closing the spell-checker itself once we no longer use it. This is important to note because `SimpleSpellChecker` is used within our custom service, and it contains several `StreamControllers` that need to be closed to avoid memory issues.
>
> ```dart
> SpellcheckerServiceProvider.instance.dispose(); // call when spell-checker will no longer used
> ```

## A little preview of this Feat:

_there's some issue with some words, but this is an error from [simple_spell_checker](https://pub.dev/packages/simple_spell_checker) since i delete case sensitive (i'll fix it later)_

https://github.com/user-attachments/assets/e164d208-c3db-4871-8413-01d24b90c3e8

### DIFF
--- a/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
+++ b/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
@@ -4,7 +4,7 @@ import 'package:simple_spell_checker/simple_spell_checker.dart';
 
 import 'spellchecker_service.dart';
 
-/// SimpleSpellCheckerImpl is a simple spell checker for get 
+/// SimpleSpellCheckerImpl is a simple spell checker for get
 /// all words divide on different objects if them are wrong or not
 class SimpleSpellCheckerImpl extends SpellcheckerService {
   SimpleSpellCheckerImpl({required super.language})
@@ -12,9 +12,10 @@ class SimpleSpellCheckerImpl extends SpellcheckerService {
           language: language,
           safeDictionaryLoad: true,
         );
+
   /// [SimpleSpellChecker] comes from the package [simple_spell_checker]
   /// that give us all necessary methods for get our spans with highlighting
-  /// where needed 
+  /// where needed
   final SimpleSpellChecker checker;
 
   @override
@@ -32,7 +33,7 @@ class SimpleSpellCheckerImpl extends SpellcheckerService {
 
   @override
   void dispose({bool onlyPartial = false}) {
-    if(onlyPartial) {
+    if (onlyPartial) {
       checker.disposeControllers();
       return;
     }
--- a/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
+++ b/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
@@ -1,13 +1,32 @@
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
+import 'package:simple_spell_checker/simple_spell_checker.dart';
 
 import 'spellchecker_service.dart';
 
 class SimpleSpellCheckerImpl extends SpellcheckerService {
-  SimpleSpellCheckerImpl({required super.language});
+  SimpleSpellCheckerImpl({required super.language})
+      : checker = SimpleSpellChecker(
+          language: language,
+          safeDictionaryLoad: true,
+        );
+  final SimpleSpellChecker checker;
 
   @override
-  List<TextSpan>? fetchSpellchecker(String text) {
-    return null;
+  List<TextSpan>? fetchSpellchecker(
+    String text, {
+    LongPressGestureRecognizer Function(String word)?
+        customLongPressRecognizerOnWrongSpan,
+  }) {
+    return checker.check(
+      text,
+      customLongPressRecognizerOnWrongSpan:
+          customLongPressRecognizerOnWrongSpan,
+    );
   }
 
+  @override
+  void dispose() {
+    checker.dispose(closeDirectionary: true);
+  }
 }
--- a/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
+++ b/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
@@ -1,11 +1,13 @@
 import 'package:flutter/material.dart';
+
 import 'spellchecker_service.dart';
 
-class DefaultSpellcheckerService extends SpellcheckerService{
-  DefaultSpellcheckerService() : super(language: 'en');
+class SimpleSpellCheckerImpl extends SpellcheckerService {
+  SimpleSpellCheckerImpl({required super.language});
 
   @override
   List<TextSpan>? fetchSpellchecker(String text) {
     return null;
   }
+
 }
--- a/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
+++ b/lib/src/editor/spellchecker/simple_spellchecker_impl.dart
@@ -4,12 +4,17 @@ import 'package:simple_spell_checker/simple_spell_checker.dart';
 
 import 'spellchecker_service.dart';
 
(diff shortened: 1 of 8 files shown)


CatHood0:
Note: i'll move `SimpleSpellCheckerServiceImpl` to flutter_quill_extensions but by now i need first add this to the new version of the repo and after i'll can be able to add it in the extensions

singerdmx (approved):
Nice

lib/src/editor/spellchecker/spellchecker_service_provider.dart
```diff
@@ -0,0 +1,19 @@
...
+
+  static SpellcheckerService get instance => _instance;
+
+  static void setInstance(SpellcheckerService service) {
+    _instance = service;
+  }
+
+  static void setInstanceToDefault() {
+    _instance = DefaultSpellcheckerService();
+  }
```
EchoEllet, line 18: We don't have this method in other services such as Clipboard?

pubspec.yaml
```diff
@@ -45,6 +45,7 @@ dependencies:
   # Dart Packages
   intl: ^0.19.0
   dart_quill_delta: ^10.0.0
+  simple_spell_checker: ^1.0.6
```
EchoEllet, line 48: Can we move this package and feature in `flutter_quill_extensions`? Mainly to make the bundle size smaller in addition to other reasons.

If a user wants a different implementation, they will have two packages instead of one.

lib/src/editor/spellchecker/simple_spellchecker_impl.dart
```diff
@@ -0,0 +1,42 @@
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:simple_spell_checker/simple_spell_checker.dart';
+
+import 'spellchecker_service.dart';
+
+/// SimpleSpellCheckerImpl is a simple spell checker for get
+/// all words divide on different objects if them are wrong or not
+class SimpleSpellCheckerImpl extends SpellcheckerService {
```
EchoEllet, line 9: A side note, we can change the name to be a bit more consistent since we're adding more services.

The previous name I chose for Clipboard service can be a bit random.

We might want to discuss the naming for future services.

EchoEllet:
> first add this to the new version of the repo and after i'll can be able to add it in the extensions

You can use the local development mode, which will use packages from your local machine. The CI will force push the publishing, and then we only need to increase the minimum version of `flutter_quill`.

You will find a dart script in the scripts` directory.

FelixMittermeier:
@CatHood0 Thanks for the useful contribution! I wanted to test it in my app but was not able to make it work. Is there any setup required to make it work and highlight the misspelled words? Thanks

CatHood0:
@FelixMittermeier Can you show the code that are you using?

lib/src/editor/spellchecker/simple_spellchecker_impl.dart
CatHood0: Yeah, that's my fault, i'm too bad naming things🤣 i'll fix this soon to avoid confusions

FelixMittermeier:
@CatHood0 Thanks for your answer. I missed the part to call:
```
SpellcheckerServiceProvider.setInstance(SimpleSpellCheckerImpl(language: 'de'));
```
first. By the way, in your first message you wrote: `SimpleSpellCheckerServiceImpl` but this is a different name which does not exist. Maybe you can modify it there as well to avoid confusion.

Also I noticed it's unfortunately not ready for production yet. For example if the text contains special characters ("Umlaute" in German) like ä, ö, ü etc. then the spell checker somehow completely removes them. My string input was for example:

`damit es ausgewählt ist`
then the `ä` was just removed from the string by the spell checker package:

IMG_217BBC0A5F6D-1

CatHood0:
@FelixMittermeier this is an issue from the Spell checker package. I'll fix it. Thanks for show that character that is being ignored!

CatHood0:
@FelixMittermeier please, take a look newly about the PR since i change some parts to be make more sense with the next version that moves this spell checker to extensions.

lib/src/editor/spellchecker/simple_spellchecker_impl.dart
EchoEllet: > i'm too bad naming things

I'm in the same boat.

Thank you for the fix. Though the name wasn't an issue, being consistent with the naming conventions is.

PR #2120, opened 2024-08-18
Chore: Move spellchecker service to extensions

CatHood0:
## Description

Since `SimpleSpellCheckerService` is a custom implementation that is not needed by default, was moved to Flutter Quill Extensions and can be used such as SuperClipboard implementation:

```dart
FlutterQuillExtensions.useSpellCheckerService('<your_language>');
```

And was fixed some issue with some missing characters while typing (this is for languages such as portuguese, deutsch, italian, etc)

## Suggestions

Please, after update the version of the main repo, also update it into `flutter_quill_extensions` to avoid conflicts since the new version change the names of the classes and methods.


lib/src/editor/spellchecker/spellchecker_service_provider.dart
```diff
@@ -3,17 +3,26 @@ import 'default_spellchecker_service.dart';
...
-  static SpellcheckerService _instance = DefaultSpellcheckerService();
+class SpellCheckerServiceProvider {
+  const SpellCheckerServiceProvider._();
+  static SpellCheckerService _instance = DefaultSpellCheckerService();
 
-  static SpellcheckerService get instance => _instance;
+  static SpellCheckerService get instance => _instance;
 
-  static void setInstance(SpellcheckerService service) {
+  static void setNewCheckerService(SpellCheckerService service) {
```
EchoEllet, line 12: We might want to enable this service in the example as well.

lib/src/editor/spellchecker/spellchecker_service_provider.dart
```diff
@@ -3,17 +3,26 @@ import 'default_spellchecker_service.dart';
...
+  static void dispose({bool onlyPartial = false}) {
+    _instance.dispose(onlyPartial: onlyPartial);
+  }
+
+  static void setNewLanguageState({required String language}) {
+    assert(language.isNotEmpty);
+    _instance.setNewLanguageState(language: language);
+  }
+
+  static void turnOffService() {
```
EchoEllet, line 25: This can be a breaking change, though it hasn't been used yet.

I'm interested in knowing the reason for the naming changes if available. 

Same for the static method `setInstance`

lib/src/editor/spellchecker/spellchecker_service_provider.dart
CatHood0: I'll add it after. I'm still doing the documentation about the Feature to the README

lib/src/editor/spellchecker/spellchecker_service_provider.dart
CatHood0: I changed it because it's not used by many users at the moment, and to make the method make more sense in the context of what the service does.

PR #2133, opened 2024-08-21
Chore: improve Spell checker API to the example

CatHood0:
## Description

It's already known that we have a simple `API` to be able to use our spell checker in the editor, but nevertheless, if we looked closely at how it was working, it was becoming quite cumbersome to activate and deactivate it due to everything that was involved. For this reason, it was updated to the latest version of `simple_spell_checker` in order to be able to use new methods like `toggleChecker()` (called into the provider as `toggleState()`) that just deactivates the service without having to close or replace it (something that we did before since there was no real way to deactivate it without changing the service), and `isServiceActive()` so that the developer can always get the status of the service.

This not only makes the `API` simpler for other developers but also improved performance since we don't have to change the instance of the service every time we deactivate or activate it.

### Not related

It was updated to the latest version because if we had the service active there were commonly used characters that were lost or ignored or even checked as if they were a word (for example, in the previous version the character ":" was taken as a word)

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,8 @@ One use of this would be having the opportunity to **activate and deactivate** t
 
 ```dart
  SpellCheckerServiceProvider.toggleState();
- isActivatedSpellChecker = SpellCheckerServiceProvider.isServiceActive();
+ // use isServiceActive to get the state of the service
+ SpellCheckerServiceProvider.isServiceActive();
  setState(() {});
 ```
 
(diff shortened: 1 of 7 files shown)


EchoEllet:
Unrelated but while testing in my current branch, I'm always having this exception:

```console
The following assertion was thrown building TextLine(dependencies: [MediaQuery, _LocalizationsScope-[GlobalKey#d9bd8], _QuillPressedKeysAccess], state: _TextLineState#09a95):
You cannot reuse this SimpleSpellchecker since you dispose it before
'package:simple_spell_checker/src/spell_checker_interface/abtract_checker.dart':
Failed assertion: line 256 pos 12: '!_disposed'
```

Sometimes the spell-checking is already disabled.

PR #2146, opened 2024-08-25
Feat: custom leading builder

CatHood0:
## Description

A new property has been added that will allow us to configure the leading of the elements in a list without any limitations.

### Example of a custom bullet list:

This code: 

```dart
 customLeadingBlockBuilder: (node, config) {
     if (config.attribute == Attribute.ul) {
            final indent = (config.attrs[Attribute.indent.key]?.value as int?) ?? 0;
             return Container(
                  alignment: AlignmentDirectional.topEnd,
                  padding: EdgeInsetsDirectional.only(end: config.padding ?? 0),
                  child: Text(
                       indent == 0
                           ? '😃'
                           : indent == 1
                                ? '👽'
                                : indent == 2
                                     ? '♥︎'
                                     : indent == 3
                                         ? '🫸'
                                         : indent == 4
                                             ? '🌏'
                                             : '🤕',
                            style: TextStyle(fontSize: 13),
               ),
          );
     }
     return null;
},
```

Will have this effect on the view:

Captura de pantalla_2024-08-26_12-36-49

## Related Issues

- *Fix #2138*
- *Fix #2149*
- *Fix #1611*
- *Fix  #2131*

## Not related issues

- *Fix #1974*

These issues are not related but was fixed while creating the feature.

- [x] ✨ **New feature:** Adds new functionality without breaking existing features.
- [x] 🛠️ **Bug fix:** Resolves an issue without altering current behavior.
- [ ] 🧹 **Code refactor:** Code restructuring that does not affect behavior.
- [ ] ❌ **Breaking change:** Alters existing functionality and requires updates.
- [ ] 🧪 **Tests:** Adds new tests or modifies existing tests.
- [ ] 📝 **Documentation:** Updates or additions to documentation.
- [ ] 🗑️ **Chore:** Routine tasks, or maintenance.
- [ ] ✅ **Build configuration change:** Changes to build or deploy processes.

## Suggestions

We should stop creating such hardcoded code as it is really getting complicated trying to create default blocks to give users the opportunity to create their own independent components.

PR #2182, opened 2024-09-04
style: refactor build leading function style, width, and padding parameters for custom node leading builder

EchoEllet:
## Description

*Move the style logic from a conditional expression into a function call for improved readability.*

The refactoring simplifies the code by replacing nested ternary operators with if statements, improving the readability a bit.

## Related Issues

- *Related #2146*

## Type of Change

### DIFF
--- a/lib/src/editor/widgets/text/text_block.dart
+++ b/lib/src/editor/widgets/text/text_block.dart
@@ -269,30 +269,35 @@ class EditableTextBlock extends StatelessWidget {
       index: isOrdered || isCodeBlock ? index : null,
       count: count,
       enabled: !isCheck ? null : !(checkBoxReadOnly ?? readOnly),
-      style: isOrdered
-          ? defaultStyles.leading!.style.copyWith(
-              fontSize: size,
-              color: context.quillEditorElementOptions?.orderedList
-                          .useTextColorForDot ==
-                      true
-                  ? fontColor
-                  : null,
-            )
-          : isUnordered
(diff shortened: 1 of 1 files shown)


CatHood0 (approved):
I hadn't thought of this way! Thanks for the change. I think this refactor is more than excellent, now it's much easier for anyone to read.

We could use this PR to do the same with other parameters that have this one.

lib/src/editor/widgets/text/text_block.dart
```diff
@@ -269,30 +269,35 @@ class EditableTextBlock extends StatelessWidget {
...
+          );
+        }
+        if (isCheck) {
+          return null;
+        }
+        return defaultStyles.code!.style.copyWith(
+          color: defaultStyles.code!.style.color!.withOpacity(0.4),
+        );
+      }(),
       width: isOrdered || isCodeBlock
```
CatHood0, line 301: As I said above, we could do the same with width which can be improved

EchoEllet: I will update the PR soon.

> As i said before

I'm not quite sure where you said it or missed the notification.

EchoEllet: > As i said before,

I'm not quite sure where you said it or missed the notification.

Will do the same with `width` and `padding` soon.

singerdmx (approved):

CatHood0 (approved):
Nice! Thanks for taking the time to do these changes.

PR #2188, opened 2024-09-05
chore: improve platform check by using constants and defaultTargetPlatform

EchoEllet:
## Description

*Improve the platform check by using `kIsWeb` and `defaultTargetPlatform` to allow it to be tested and tree shaken.*

- `kIsWeb` will always be `true` if `kIsWasm` is in case of [Flutter/WASM](https://docs.flutter.dev/platform-integration/web/wasm)
- The main reason for using `defaultTargetPlatform` over `Platform.operatingSystem` from `dart: is that it can be tested and overridden, and supports the web in case we want to check if the platform regardless if it's on a web browser.
- Recently (about 7 months ago), `defaultTargetPlatform` has been updated to use `vm:platform-const-if` pragma introduced in [https://github.com/dart-lang/sdk/commit/57a1168875](https://github.com/dart-lang/sdk/commit/57a1168875) in [Flutter #141105](https://github.com/flutter/flutter/pull/141105) which allow it to be computed as if it was a constant field in non-debug builds which means platform-specific code executed conditionally based on `defaultTargetPlatform` can be tree-shaken in release builds
- Added properties such as `isIos` which will check if the operating system is iOS which will be `true` even on the web, `isIosApp` will require the platform to be an iOS app and non-web to be `true`:
```dart
bool get isIos => defaultTargetPlatform == TargetPlatform.iOS;
bool get isIosApp => !kIsWeb && isIos;
```

Also added `@pragma('vm:platform-const-if', !kDebugMode)` on each property, `kIsWeb` is a constant field, `defaultTargetPlatform` is constant only on non-debug builds, I'm uncertain if we need the pragma annotation for each field (e.g. `isAndroidApp` or `isMobileApp`) to be computed as constants and tree shaken same way if we use `defaultTargetPlatform` directly.

If we take a look at the [file`platform.dart` from `dart:io`](https://github.com/dart-lang/sdk/blob/main/sdk/lib/io/platform.dart#L82-L83), all fields that are using `operatingSystem`:

```dart
  @pragma("vm:platform-const")
  static final operatingSystem = _Platform.operatingSystem;
```

are also using `@pragma("vm:platform-const")`:

```dart
  @pragma("vm:platform-const")
  static final bool isLinux = (operatingSystem == "linux");

  @pragma("vm:platform-const")
  static final bool isIOS = (operatingSystem == "ios");

  // Etc...
```

## Related Issues

I expect some regressions from this PR, as such it's not ready for merge yet.

Also see:

- [VM-Specific Pragma Annotations](https://github.com/dart-lang/sdk/blob/main/runtime/docs/pragmas.md). Was not able to find this page due to an issue [Dart #54672](https://github.com/dart-lang/sdk/issues/54672)
- *[Flutter #141105](https://github.com/flutter/flutter/pull/141105)*
- *[Flutter #14233](https://github.com/flutter/flutter/issues/14233)*
- *[Dart #31969](https://github.com/dart-lang/sdk/issues/31969)*
- [`defaultTargetPlatform` docs](https://github.com/flutter/flutter/blob/aea84342eb861832bbf8578de432c4da5a074136/packages/flutter/lib/src/foundation/platform.dart#L17-L52)

## Type of Change

### DIFF
--- a/lib/src/editor/editor.dart
+++ b/lib/src/editor/editor.dart
@@ -3,7 +3,7 @@ import 'dart:math' as math;
 import 'package:flutter/cupertino.dart'
     show CupertinoTheme, cupertinoTextSelectionControls;
 import 'package:flutter/foundation.dart'
-    show ValueListenable, defaultTargetPlatform;
+    show ValueListenable, defaultTargetPlatform, kIsWeb;
 import 'package:flutter/gestures.dart'
     show
         PointerDeviceKind,
@@ -263,10 +263,7 @@ class QuillEditorState extends State<QuillEditor>
     Color selectionColor;
     Radius? cursorRadius;
 
-    if (isAppleOS(
-      platform: theme.platform,
-      supportWeb: true,
-    )) {
+    if (theme.isCupertino) {
       final cupertinoTheme = CupertinoTheme.of(context);
       textSelectionControls = cupertinoTextSelectionControls;
       paintCursorAboveText = true;
@@ -316,10 +313,7 @@ class QuillEditorState extends State<QuillEditor>
                   ? (configurations.contextMenuBuilder ??
                       QuillRawEditorConfigurations.defaultContextMenuBuilder)
                   : null,
-              showSelectionHandles: isMobile(
-                platform: theme.platform,
-                supportWeb: true,
-              ),
+              showSelectionHandles: isMobile,
               showCursor: configurations.showCursor ?? true,
               cursorStyle: CursorStyle(
                 color: cursorColor,
@@ -381,7 +375,7 @@ class QuillEditorState extends State<QuillEditor>
           )
         : child;
 
-    if (isWeb()) {
+    if (kIsWeb) {
       // Intercept RawKeyEvent on Web to prevent it from propagating to parents
       // that might interfere with the editor key behavior, such as
       // SingleChildScrollView. Thanks to @wliumelb for the workaround.
@@ -477,11 +471,7 @@ class _QuillEditorSelectionGestureDetectorBuilder
       return;
     }
 
-    final platform = Theme.of(_state.context).platform;
-    if (isAppleOS(
-      platform: platform,
-      supportWeb: true,
-    )) {
+    if (Theme.of(_state.context).isCupertino) {
       renderEditor!.selectPositionAt(
         from: details.globalPosition,
         cause: SelectionChangedCause.longPress,
@@ -554,9 +544,7 @@ class _QuillEditorSelectionGestureDetectorBuilder
 
     try {
       if (delegate.selectionEnabled && !_isPositionSelected(details)) {
-        final platform = Theme.of(_state.context).platform;
-        if (isAppleOS(platform: platform, supportWeb: true) ||
-            isDesktop(platform: platform, supportWeb: true)) {
+        if (isAppleOS || isDesktop) {
           // added isDesktop() to enable extend selection in Windows platform
           switch (details.kind) {
             case PointerDeviceKind.mouse:
@@ -619,11 +607,7 @@ class _QuillEditorSelectionGestureDetectorBuilder
     }
 
     if (delegate.selectionEnabled) {
-      final platform = Theme.of(_state.context).platform;
-      if (isAppleOS(
-        platform: platform,
-        supportWeb: true,
-      )) {
+      if (Theme.of(_state.context).isCupertino) {
         renderEditor!.selectPositionAt(
           from: details.globalPosition,
           cause: SelectionChangedCause.longPress,
--- a/lib/src/editor/raw_editor/raw_editor_state.dart
+++ b/lib/src/editor/raw_editor/raw_editor_state.dart
@@ -513,11 +513,7 @@ class QuillRawEditorState extends EditorState
             maxHeight: widget.configurations.maxHeight ?? double.infinity,
           );
 
-    // Please notice that this change will make the check fixed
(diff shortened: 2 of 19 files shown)


singerdmx (approved):

lib/src/editor/raw_editor/raw_editor_state.dart
```diff
@@ -1191,9 +1187,9 @@ class QuillRawEditorState extends EditorState
     _floatingCursorResetController = AnimationController(vsync: this);
     _floatingCursorResetController.addListener(onFloatingCursorResetTick);
 
-    if (isKeyboardOS(supportWeb: true)) {
+    if (isKeyboardOS) {
```
EchoEllet, line 1190: I haven't understood what this check mean by `keyboardOS`, I assume it's an operating system with a built-in physical keyboard. It checks if it's desktop or fuchsia OS and fuchsia OS doesn't come with a built-in keyboard AFAIK.

lib/src/editor/editor.dart
```diff
@@ -619,11 +607,7 @@ class _QuillEditorSelectionGestureDetectorBuilder
     }
 
     if (delegate.selectionEnabled) {
-      final platform = Theme.of(_state.context).platform;
-      if (isAppleOS(
-        platform: platform,
-        supportWeb: true,
-      )) {
+      if (Theme.of(_state.context).isCupertino) {
```
EchoEllet, line 610: Many of the checks here depend on `ThemeData` from `MaterialApp`. Usually, it should be for styling purposes so I created an extension called `isCupertino` to be a bit clear.

We need to double-check if this platform-specific logic shouldn't be used when overriding the platform with an Apple operating system on Android or other platforms.

lib/src/editor/editor.dart
```diff
@@ -316,10 +313,7 @@ class QuillEditorState extends State<QuillEditor>
                   ? (configurations.contextMenuBuilder ??
                       QuillRawEditorConfigurations.defaultContextMenuBuilder)
                   : null,
-              showSelectionHandles: isMobile(
```
EchoEllet, line 319: I'm not quite sure if this is a reason why we're overriding the platform with the one from `ThemeData`

lib/src/editor/editor.dart
```diff
@@ -263,10 +263,7 @@ class QuillEditorState extends State<QuillEditor>
     Color selectionColor;
     Radius? cursorRadius;
 
-    if (isAppleOS(
-      platform: theme.platform,
-      supportWeb: true,
-    )) {
```
EchoEllet, line 269: Same question

lib/src/editor/editor.dart
CatHood0: It was probably a temporary solution added several versions ago that was eventually dropped due to other issues.

test/common/utils/platform_test.dart
```diff
@@ -20,137 +18,43 @@ void main() {
...
-            platform: mobilePlatform,
-          ),
-          true,
+          isMobile,
+          false,
         );
       }
     });
-    test(
-      'Check supportWeb parameter when using desktop platform on web',
```
CatHood0, line 94: Why were these tests removed?

AtlasAutocode:
This PR clears up a lot of the confusion on implementing platform specific options. I have never understood `supportWeb:` and anything that removes the confusion is good.

_Comment:_
One point that may lead to confusion is the close similarity of `isIOS` and `isIos`, for example. When you use intellisense to auto enter code, it might be random which gets picked. Also, as I get involved in other programming tasks, I tend to forget the distinction and why I should pick one over the other. 

The impact on tree-shaking is something I have not considered but will be very important when there is a platform-specific option that could increase program size by multi-megabytes.

I note that text_selection.dart and delegate.dart are already using defaultTargetPlatform.

test/common/utils/platform_test.dart
EchoEllet: > Why were these tests removed?

We no longer depend on it. Previously, we used something like `isIos` with `supportWeb`, which determined whether the result should be `true` for web or not. Now, we depend on `kIsWeb`, which is not overridable.

EchoEllet:
> This PR clears up a lot of the confusion on implementing platform specific options. I have never understood supportWeb: and anything that removes the confusion is good.


If `supportWeb` is `true` and we're on iOS and check if we're on iOS, then it will return `true` on the web, otherwise will only return `true` on standlone iOS app and not on web browser, useful when you want to use platform specific features that can't be on other platforms. The name is a bit confusing.

> One point that may lead to confusion is the close similarity of isIOS and isIos, for example. When you use intellisense to auto enter code, it might be random which gets picked. Also, as I get involved in other programming tasks, I tend to forget the distinction and why I should pick one over the other.

The choice is a bit random, which leads to inconsistency since we will choose something different each time. I decided to stick to `iOS` since it's more common. 

> The impact on tree-shaking is something I have not considered but will be very important when there is a platform-specific option that could increase program size by multi-megabytes.

Usually, this will be useful to help improve the performance a bit where using platform check extensively, for example, when building platform adaptive application with a lot of check-in `build` method.

> I note that text_selection.dart and delegate.dart are already using defaultTargetPlatform.

The `platform.dart` was already using `defaultTargetPlatform`, Previously, I didn't know there was a way to override the platform for tests. Tree shaking for this property wasn't a feature.

AtlasAutocode (approved):
Thanks for your comments. Looks good.

CatHood0 (approved):

singerdmx (approved):

PR #2195, opened 2024-09-09
feat: Support for QuillTextLink in  QuillToolbarLinkStyleButtonOptions

ferdiangunawan:
This PR may not be valid if there is a better approach that is not covered in the documentation.

## Context
When implementing a link button using QuillEditor (not QuillEditor.basic), developers currently use QuillToolbarLinkStyleButton and QuillToolbarLinkStyleButtonOptions for styling. However, when users select text containing a link, the link button menu should ideally switch to an active state, which requires determining whether the selected text includes a link.

After reviewing the documentation, I found no clear API guidance on how to check if the selected text contains a link. Upon inspecting the codebase, I discovered that QuillTextLink.prepare can be used to retrieve the necessary link information.

## Proposal
This PR proposes adding QuillTextLink to QuillToolbarLinkStyleButtonExtraOptions. By passing the textLink through the childBuilder property in QuillToolbarLinkStyleButton, developers can directly determine the link presence in the text without having to explore the codebase to discover QuillTextLink.prepare().

## Changes of the usage
Before:

```dart
QuillToolbarLinkStyleButton(
  controller: _controller,
  options: QuillToolbarLinkStyleButtonOptions(
    childBuilder: (options, extraOptions) {
      final textLink = QuillTextLink.prepare(_controller);
      final hasLink = textLink.link != null;

      return GestureDetector(
        // Additional implementation
      );
    },
  ),
);
```

After:

```dart
QuillToolbarLinkStyleButton(
  controller: _controller,
  options: QuillToolbarLinkStyleButtonOptions(
    childBuilder: (options, extraOptions) {
      final hasLink = extraOptions.textLink.link != null;

      return GestureDetector(
        // Additional implementation
      );
    },
  ),
);
```

This enhancement simplifies the implementation by allowing developers to utilize the API more effectively without needing to manually check for links in the codebase. If there is a better approach or an existing API that I missed, I'm open to suggestions and adjustments.

## Type of Change

### DIFF
--- a/lib/src/toolbar/buttons/link_style_button.dart
+++ b/lib/src/toolbar/buttons/link_style_button.dart
@@ -81,11 +81,13 @@ class QuillToolbarLinkStyleButtonState
     final childBuilder =
         options.childBuilder ?? baseButtonExtraOptions?.childBuilder;
     if (childBuilder != null) {
+      final textLink = QuillTextLink.prepare(controller);
       return childBuilder(
         options,
         QuillToolbarLinkStyleButtonExtraOptions(
           context: context,
           controller: controller,
+          textLink: textLink,
           onPressed: () {
             _openLinkDialog(context);
             afterButtonPressed?.call();
--- a/lib/src/toolbar/config/buttons/link_style2_configurations.dart
+++ b/lib/src/toolbar/config/buttons/link_style2_configurations.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart';
 
+import '../../../editor/widgets/link.dart';
 import '../../theme/quill_dialog_theme.dart';
 import '../base_button_configurations.dart';
 
@@ -8,8 +9,10 @@ class QuillToolbarLinkStyleButton2ExtraOptions
   const QuillToolbarLinkStyleButton2ExtraOptions({
     required super.controller,
     required super.context,
+    required this.textLink,
     required super.onPressed,
   });
+  final QuillTextLink textLink;
 }
 
 class QuillToolbarLinkStyleButton2Options extends QuillToolbarBaseButtonOptions<
--- a/lib/src/toolbar/buttons/link_style2_button.dart
+++ b/lib/src/toolbar/buttons/link_style2_button.dart
@@ -122,11 +122,13 @@ class _QuillToolbarLinkStyleButton2State
     final childBuilder =
         options.childBuilder ?? baseButtonExtraOptions?.childBuilder;
     if (childBuilder != null) {
+      final textLink = QuillTextLink.prepare(controller);
       return childBuilder(
         options,
         QuillToolbarLinkStyleButton2ExtraOptions(
           controller: controller,
           context: context,
+          textLink: textLink,
           onPressed: () {
             _openLinkDialog();
             afterButtonPressed?.call();
--- a/lib/src/toolbar/config/buttons/link_style_configurations.dart
+++ b/lib/src/toolbar/config/buttons/link_style_configurations.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart' show Color;
 
+import '../../../editor/widgets/link.dart';
 import '../../base_toolbar.dart';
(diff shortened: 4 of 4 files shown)


lib/src/toolbar/config/buttons/link_style2_configurations.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:flutter/widgets.dart';
 
+import '../../../../flutter_quill.dart';
```
singerdmx, line 3: Please do individual import instead of importing everything

singerdmx (changes requested):

singerdmx:
info • The import of '../base_button_configurations.dart' is unnecessary because all of the used elements are also provided by the import of '../../../../flutter_quill.dart' • lib/src/toolbar/config/buttons/link_style2_configurations.dart:5:8 • unnecessary_import

lib/src/toolbar/config/buttons/link_style2_configurations.dart
ferdiangunawan: okay

EchoEllet:
The naming we have in this area is a quite confusing and some things are a bit broken, we will have to break things in order to solve this issue, we already did some breaking changes and we already have plan for more soon so I suggest we delay this for quite some time (at least 6 months or more).

The `extraOptions` contains the state properties of the original widget to use a different widget with the same state, and is not for helpers or such classes, since it's possible using the (before) code snippet you used, I would suggest creating an extension function instead in your app.

I suggest documenting this class and providing a code snippet in the docs instead.

lib/src/toolbar/buttons/link_style_button.dart
```diff
@@ -81,11 +81,13 @@ class QuillToolbarLinkStyleButtonState
     final childBuilder =
         options.childBuilder ?? baseButtonExtraOptions?.childBuilder;
     if (childBuilder != null) {
+      final textLink = QuillTextLink.prepare(controller);
```
EchoEllet, line 84: Same here

singerdmx:
@EchoEllet 6 months is a bit too long

EchoEllet:
> @EchoEllet 6 months is a bit too long

To give users more time to migrate from deprecated code before we remove it.

singerdmx:
@EchoEllet I will leave this PR for you to decide. Thanks!

singerdmx:
Is this PR not backwards-compatible? If not, maybe there is no harm merging it.

EchoEllet:
> Is this PR not backwards-compatible? If not, maybe there is no harm merging it.

It's backward compatible. The issue is that it adds something that's not quite related to the `extraOptions`, and the name of this property is quite confusing. It already has properties that are not related to it. We do have a lot of internal APIs that could use improvements for maintainability, organization, and consistency. 

So I would prefer to be a bit more careful when adding more features. 

For this PR, it's already possible to check if the text has a link with an additional line.

We can document this and make it as an example. In the `example` app and in docs, and that's what I'm suggesting.

singerdmx:
Thanks for the explanation. Sounds good to me.

ferdiangunawan:
Thank you for the review and suggestions @singerdmx @EchoEllet. that sounds good to me. So, this PR will be closed, and documen this on example instead, right?

singerdmx:
I believe so

EchoEllet:
> So, this PR will be closed, and document this on example instead, right?

You can revert the changes and then document it in the example in the same PR or submit another.

We will either document it as an example somewhere in the`doc` directory or in the `example` app or both.

You can also leave it to me if you want.

ferdiangunawan:
> > So, this PR will be closed, and document this on example instead, right?
> 
> You can revert the changes and then document it in the example in the same PR or submit another.
> 
> We will either document it as an example somewhere in the`doc` directory or in the `example` app or both.
> 
> You can also leave it to me if you want.

alright, I will submit another for documentation and close this PR. thankyou

EchoEllet:
Thank you for your contributions.

PR #2223, opened 2024-09-13
refactor: moved shortcuts and onKeyEvents to its own file

CatHood0:
## Description

Moved `Shortcut` widget and `Focus` widget that listened for keyboard events to their own file in order to reduce the size of `raw_editor_state` and to separate the logic and make it easier to read.

### DIFF
--- a/lib/src/editor/widgets/default_single_activator_actions.dart
+++ b/lib/src/editor/widgets/default_single_activator_actions.dart
@@ -1,0 +1,206 @@
+import 'package:flutter/material.dart'
+    show
+        AxisDirection,
+        Intent,
+        RedoTextIntent,
+        ScrollIncrementType,
+        ScrollIntent,
+        ScrollToDocumentBoundaryIntent,
+        SelectionChangedCause,
+        SingleActivator,
+        UndoTextIntent;
+import 'package:flutter/services.dart'
+    show LogicalKeyboardKey, SelectionChangedCause;
+
+import '../../document/attribute.dart';
+import '../raw_editor/raw_editor_actions.dart'
+    show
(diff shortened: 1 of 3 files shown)


EchoEllet:
If the `keyboard_service_widget.dart` is only used in `QuillRawEditorState`, shouldn't this file be somewhere like `editor/raw_editor/widgets` instead of `editor/widgets`?

CatHood0:
I'll move it in another PR. Thanks for report it

EchoEllet:
@CatHood0 

Just as a reminder, the `QuillKeyboardServiceWidget` class file is still not moved.

image

I also suggest maybe using a different name such as `EditorKeyboardShortcuts`?
Would like to know the reason behind choosing the current name first.

EchoEllet:
> Probably just a bad naming

Doesn't have to.

If you prefer, I can introduce this change since it's minimal.

lib/src/editor/widgets/default_single_activator_actions.dart
```diff
@@ -0,0 +1,206 @@
...
+        OpenSearchIntent,
+        QuillEditorApplyCheckListIntent,
+        QuillEditorApplyHeaderIntent,
+        QuillEditorApplyLinkIntent,
+        QuillEditorInsertEmbedIntent,
+        ToggleTextStyleIntent;
+
+Map<SingleActivator, Intent> defaultSinlgeActivatorActions(
+        bool isDesktopMacOS) =>
+    {
```
EchoEllet, line 29: It's something we have missed, this should be `defaultSinlgeActivatorIntents` since the actions are not here.

PR #2228, opened 2024-09-14
Feat: customizable character and space shortcut events

CatHood0:
## Description

A new possibility to add real-time shortcuts has been created using `CharacterShortcutEvent` and` SpaceShorcutEvent`. Both are executed when `_onKeyEvent` is called, however, checks are made to not execute everything at the same time first (this could be subject to change in order to improve the performance of the operations).

`CharacterShorcutEvent` is a class that is called in a general way every time the user types something inside the editor (these events are ignored and the key is space, tab, or a new line).

`SpaceShorcutEvent` is a class that is only called (as its name says) when the space key is pressed.

### Without shorcuts implemented
<p align="left">
  
</p>

### With shorcuts implemented

https://github.com/user-attachments/assets/106348d3-c8b8-4eee-9d55-0f734fa93490


README.md
```diff
@@ -291,6 +293,128 @@ It's implemented using the package `simple_spell_checker` in the [Example](./exa
 
 Take a look at [Spelling Checker](./doc/spell_checker.md) page for more info.
 
+## ✂️ Shortcut events
```
EchoEllet, line 296: Can you move this section into its own file?

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
```diff
@@ -0,0 +1,120 @@
...
+      break;
+    case SingleCharacterFormatStyle.strikethrough:
+      style = const StrikeThroughAttribute();
+      break;
+    case SingleCharacterFormatStyle.code:
+      style = const InlineCodeAttribute();
+      break;
+    default:
+      style = null;
+      assert(false, 'Invalid format style');
```
EchoEllet, line 94: If all cases are handled, the `default` would be optional?

lib/src/editor/raw_editor/config/events/space_shortcut_events.dart
```diff
@@ -0,0 +1,51 @@
...
+
+typedef SpaceShortcutEventHandler = bool Function(
+    QuillText node, QuillController controller);
+
+/// Defines the implementation of shortcut events for space key calls.
+class SpaceShortcutEvent {
+  SpaceShortcutEvent({
+    required this.character,
+    required this.handler,
+  }) : assert(character != '\n' && character.trim().isNotEmpty);
```
EchoEllet, line 12: Can you add a message like indicating that the character should not be empty or a new line? This would be useful when the assertion fails and we get a bug report.

lib/src/editor/raw_editor/config/events/space_shortcut_events.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  String character;
+  final SpaceShortcutEventHandler handler;
+
+  void updateCharacter(String newCharacter) {
+    assert(character != '\n' && character.trim().isNotEmpty);
```
EchoEllet, line 18: Same here

lib/src/editor/raw_editor/config/events/character_shortcuts_events.dart
```diff
@@ -0,0 +1,55 @@
...
+typedef CharacterShortcutEventHandler = bool Function(
+    QuillController controller);
+
+/// Defines the implementation of shortcut event based on character.
+class CharacterShortcutEvent {
+  CharacterShortcutEvent({
+    required this.key,
+    required this.character,
+    required this.handler,
+  }) : assert(character.length == 1 && character != '\n');
```
EchoEllet, line 12: The same for other assertions.

lib/src/editor/raw_editor/config/events/character_shortcuts_events.dart
```diff
@@ -0,0 +1,55 @@
...
+      handler: handler ?? this.handler,
+    );
+  }
+
+  @override
+  String toString() =>
+      'CharacterShortcutEvent(key: $key, character: $character, handler: $handler)';
+
+  @override
+  bool operator ==(Object other) {
```
EchoEllet, line 44: Not necessary but you can also use [`equatable`](https://pub.dev/packages/equatable) instead of implementing `==` operation and hashcode.

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
CatHood0: Yeah, probably i will remove it because never is handled (since the implementation is the default from the project and the users cannot access to these functions)

README.md
CatHood0: I'll do it

EchoEllet: Thank you.

AtlasAutocode (commented):
It looks like this is cleaning up a lot of complex code and enabling a lot of potential for customization. Without trying the code, the only thing that worries me is typing in the following:

> Multiplication example: 2 * 3 = 6, whereas the product of 3 * 4 = 12

Would that appear as:
> Multiplication example: 2 _3 = 6, whereas the product of 3_ 4 = 12

where the text is converted to italic when it should not be.
Is there a way of saying not to apply the conversion - I really want an asterisk here.
Asterisk is also used as a superscript to refer to a footnote. Would that also trigger the conversion?

On desktops, it is easier to just use control+i (windows, or mac Cmd+i? linux?) to turn on/off italic so it would never occur to me to use markdown syntax. For mobile, the reverse would be true!

CatHood0:
> where the text is converted to italic when it should not be.
Is there a way of saying not to apply the conversion - I really want an asterisk here.
Asterisk is also used as a superscript to refer to a footnote. Would that also trigger the conversion?

There will always be one condition for CharacterShortcutEvents (at least the default ones), and that is, in this case at least, the asterisk needs and must be the one on the right side. That is, if I insert an asterisk on the right and then one on the left, it will not be applied (because it is not what is wanted by default).

I know it may seem a bit confusing, because it is new and probably requires more documentation, but in this case of asterisks the handler could be bypassed if we put the asterisk on the right first and then the one on the left. However, if in your case, at least you want this type of text block to be ignored, you can always create your own handler to ignore any text that is between numbers.

It should be noted that these are NOT implemented by default in the editor. It is the user who must call standardCharacterShortcutEvents to have these implementations.

README.md
EchoEllet: Thanks.

singerdmx (approved):

lib/src/editor/raw_editor/config/raw_editor_configurations.dart
```diff
@@ -49,6 +50,8 @@ class QuillRawEditorConfigurations extends Equatable {
     required this.selectionCtrls,
     required this.embedBuilder,
     required this.autoFocus,
+    required this.characterShortcutEvents,
+    required this.spaceShortcutEvents,
```
EchoEllet, line 54: Why were those added as required? We shouldn't add required parameters in this case and shouldn't at all in non-major versions.

CatHood0: It was most likely a mistake on my part when I was creating the feature.

EchoEllet: Don't worry about it, fixed in https://github.com/singerdmx/flutter-quill/pull/2522.

PR #2234, opened 2024-09-16
Fix: QuillToolbarToggleStyleButton Switching failure

### LINKED ISSUE #2216: QuillToolbarToggleStyleButton， QuillToolbarToggleStyleButton Switching failure  on macOS

### Flutter Quill version

flutter_quill: ^10.5.16

### Steps to reproduce

First, add a task list and input the content, then press enter to create a new empty task list. 
Then, delete this new task list and return to the first list.
 Finally, switch to QuillToolbarToggleCheckListButton
At this point, clicking on the switch event will render it invalid

### Expected results

Click to switch states correctly

### Actual results

First, add a task list and input the content, then press enter to create a new empty task list. Then, delete this new task list and return to the first list. Finally, switch to QuillToolbarToggleCheckListButton
At this point, clicking on the switch event will render it invalid

### Code sample

I am using the demo from the Flutter_quill library
```dart
      QuillToolbarToggleCheckListButton(
                    controller: controller,
                  ),
                  QuillToolbarToggleStyleButton(
                    controller: controller,
                    attribute: Attribute.ol,
                  ),
                  QuillToolbarToggleStyleButton(
                    controller: controller,
                    attribute: Attribute.ul,
                  ),
```

</details>

### Additional Context

<details>

[Upload media here]

</details>

```console
[Paste your logs here]
```

</details>
[新标签页.webm](https://github.com/user-attachments/assets/4d02ff04-bfbe-4930-a07d-12c1353bdddf)

AtlasAutocode:
## Description

Fix: QuillToolbarToggleStyleButton Switching failure for block attributes

## Related Issues

- *Fix #2216*

## Type of Change

- [ X ] 🛠️ **Bug fix:** Resolves an issue without altering current behavior.


EchoEllet:
Thank you for addressing the `pasteUsingPlainOrDelta()` issue.

As a side note, we should avoid exposing new APIs to the public, when a new property or function is introduced, it should be experimental for a while before promoting to stable when something is deprecated it's probably better to create extension on the class and move the deprecated members there instead of having them in the main file, we usually want to make more members private unless there is a reason to not, or moving them into a separate file without exporting the internal file.

Also using `@visibleForTesting` is useful. It can be considered as a breaking change if that API commonly used.

lib/src/controller/quill_controller.dart
```diff
@@ -603,7 +601,7 @@ class QuillController extends ChangeNotifier {
     return false;
   }
 
-  /// Internal method to allow unit testing
+  @visibleForTesting
```
EchoEllet, line 604: Update: I haven't noticed it but it seems this function is used internally, `@visibleForTesting` is for code that will only used in tests and should not used internally in the package itself or by users. Instead should use `@internal`

PR #2235, opened 2024-09-17
feat: soft keyboard events

### LINKED ISSUE #2214: `QuillEditorConfigurations.enableMarkdownStyleConversion` not respected

### Flutter Quill version

10.5.16

### Steps to reproduce

Platform: iOS/Android

1. Open example app
2. Open empty document
3. Type `1. ` (space included) __USE SOFTWARE KEYBOARD__

### Expected results

The `1.` is turned into ordered list

### Actual results

The `1.` remains an ordinary text

### Additional Context

I've looked into the bit where the auto formatting should happen:

`_onKeyEvent` > `_handleSpaceKey`

https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/raw_editor/raw_editor_state.dart#L790

The problem is `_onKeyEvent` is never getting triggered. 

It could be https://github.com/singerdmx/flutter-quill/issues/2039 is related .

vishna:
## Description

Provides space/character event emulation on platforms that are not equipped with hardware keyboard (e.g. Android/iOS).

Emulation is based on diff between delta, so it happens after the change is committed to the document rather than on key press.

Enable it by setting true to `QuillEditorConfigurations.softKeyboardShortcutSupport` (This disables hardware events)

There might be some issues I haven't foreseen/came across yet, but it seems to work fine with all the standard  shortcuts.

https://github.com/user-attachments/assets/4c3be427-afe4-4d66-b676-78000507035a

## Related Issues

- Fix #2214

## Type of Change

- [x] ❌ **Breaking change:** Alters existing functionality and requires updates. (__MINOR__: breaking change, affects classes overriding QuillController.replaceText )


example/lib/screens/quill/quill_screen.dart
```diff
@@ -130,6 +131,9 @@ class _QuillScreenState extends State<QuillScreen> {
                       searchEmbedMode: SearchEmbedMode.plainText,
                     ),
                     sharedConfigurations: _sharedConfigurations,
+                    softKeyboardShortcutSupport:
+                        defaultTargetPlatform == TargetPlatform.android ||
+                            defaultTargetPlatform == TargetPlatform.iOS,
```
EchoEllet, line 136: Is this feature specific to Android and iOS outside of web browser (native app)? If yes, then this should be updated to `isAndroidApp` and `isIosApp` which can be imported from `platform.dart`.

lib/src/controller/quill_controller.dart
```diff
@@ -369,6 +395,7 @@ class QuillController extends ChangeNotifier {
     });
   }
 
+  // FIXME attribute should be non-nullable
```
EchoEllet, line 398: Is this a TODO? If yes, then I would suggest updating it to start with `TODO` (e.g. `TODO: Attribute should be non-nullable, fix later`)

EchoEllet (changes requested):

lib/src/controller/quill_controller.dart
vishna: FIXME is a variant of TODO (gets picked up by dart analyzer/vscode afaik) - I'll change it to TODO as requested + create issue and link it in the comment

vishna: ok, looking at issue template it might be an overkill, will just fix the comment 😅

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
```diff
@@ -0,0 +1,149 @@
...
+  if (text == null) {
+    return false;
+  }
+
+  var effectiveTextValue = text.value;
+
+  final documentOffset = text.documentOffset;
+  if (baseOffset > documentOffset + 1) {
+    try {
+      effectiveTextValue =
```
CatHood0, line 129: Why do you make this operation?  

_I really want to know why, because as far i know, Try/catch could affect to the perfomance with large docs_

lib/src/controller/quill_controller.dart
EchoEllet: I will probably simplify the issue bug/feature template soon.

lib/src/controller/quill_controller.dart
```diff
@@ -57,6 +60,25 @@ class QuillController extends ChangeNotifier {
         selection: const TextSelection.collapsed(offset: 0),
       );
 
+  void _maybeEnableSoftKeyboard() {
```
CatHood0, line 63: > I think we should need to add an assert to verify if the platform is not mobile, because some users couldn't know what makes this parameter, and just put true without know that this makes the onKeyEvent ignore all events even when this is Desktop or Web. Or just we can add a validation to check if the current device is Android or IOS (that removes the assert).
> 
> Doing this we can remove isAdroidApp or isIosApp in the example and the users can put true or false without remove any functionality on their other platforms.

Following with this, you can also add an validation here for desktop or web to avoid execute this method when is not needed. Avoiding add a listener or add instances when won't be needed .

lib/src/editor/raw_editor/config/events/format/format_double_character_handler.dart
```diff
@@ -34,9 +34,13 @@ bool handleFormatByWrappingWithDoubleCharacter({
     return false;
   }
 
+  final isSoftInsert =
+      controller.editorConfigurations.softKeyboardShortcutSupport;
```
CatHood0, line 38: If you add the conditions, we will probably have to do the same here as well (i.e. add to validate if it is web or desktop as well) to avoid any unexpected errors.

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
```diff
@@ -33,9 +33,13 @@ bool handleFormatByWrappingWithSingleCharacter({
     return false;
   }
 
+  final isSoftInsert =
+      controller.editorConfigurations.softKeyboardShortcutSupport;
```
CatHood0, line 37: > If you add the conditions, we will probably have to do the same here as well (i.e. add to validate if it is web or desktop as well) to avoid any unexpected errors.

Same for this.

CatHood0 (changes requested):
See the changes that need to be do it into the code to avoid unexpected errors and new issues created by users that may be confused with how works this new functionality.

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
```diff
@@ -0,0 +1,149 @@
...
+          return lastOpData;
+        }
+      }
+    }
+
+    return null;
+  }
+}
+
+extension _QuillControllerExt on QuillController {
```
CatHood0, line 83: I think we should need this on it's own file to improve the readability (just avoid exposing it to the public API).

EchoEllet: We can also leave a comment indicating, which will be useful for future developers working on this file.

We can't leave comments everywhere, we probably need a way to put our internal APIs separately in more organized way, for example, each file have its file that have internal members, for this having `soft_keyboard_shortcut_support_internal.dart`.

We should check how Flutter does this internally to follow conventions that most developers are familiar with. For example you usually don't call a widget that's a button `WButton` similar to Java swing (e.g. `JLabel`, `JButton`) on Flutter and other frameworks

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
vishna: - this whole method is based off `QuillKeyboardServiceWidget._handleSpaceKey`
- `effectiveTextValue` used in line 138: `if (spaceEvent.character == effectiveTextValue)`
- I do this calculation since at this point space has been added to the document versus `QuillKeyboardServiceWidget._handleSpaceKey` where you have ability to prevent adding the space
- I might need to re-investigate try/catch (I was getting out of bounds when typing in the longer line that had some formatting applied in the middle, it's probably coming from this:
    - `final text = castOrNull<leaf.QuillText>(line.first);`
    - probably should be fine to ignore anything out of bounds since the space events are only applied at the beginning of the line

I'll put some extra effort to make it prettier here

lib/src/editor/widgets/keyboard_service_widget.dart
```diff
@@ -71,6 +71,10 @@ class QuillKeyboardServiceWidget extends StatelessWidget {
   }
 
   KeyEventResult _onKeyEvent(node, KeyEvent event) {
+    if (controller.editorConfigurations.softKeyboardShortcutSupport) {
+      return KeyEventResult.ignored;
```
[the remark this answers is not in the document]
vishna: I'll go with an assert(s)

lib/src/controller/quill_controller.dart
vishna: I'll go with an assert(s)

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
CatHood0: Thanks for the explanation.

vishna:
@CatHood0 @EchoEllet thanks for the thorough review 🙇  applied all the remarks, let me know if I can make it any better

example/lib/screens/quill/quill_screen.dart
```diff
@@ -1,6 +1,7 @@
 import 'dart:convert' show jsonEncode;
 
 import 'package:flutter/material.dart';
+import 'package:flutter_quill/extensions.dart';
```
EchoEllet, line 4: Not sure why this file is imported? It's exist so the `flutter_quill_extensions` can have access for some internal APIs that are not meant to be used internally, or in `flutter_quill` or in the example either since that can be confusing to the users.

The `example` is also for users to see how to use `flutter_quill` properly, I have added a comment in this `extensions.dart` file but it's still confusing since the name doesn't have `internal` in it and the docs doesn't cover this file so we probably don't want this file in the example.

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
```diff
@@ -33,9 +34,17 @@ bool handleFormatByWrappingWithSingleCharacter({
     return false;
   }
 
+  final usesSoftKeyboardShortcut =
+      controller.editorConfigurations.softKeyboardShortcutSupport;
+  if (usesSoftKeyboardShortcut) {
+    assert(isAndroidApp || isIosApp,
+        'softKeyboardShortcutSupport should only be used on Android/iOS');
```
EchoEllet, line 41: Have noticed there are different places uses this same assertion with the same message and check, doesn't have to be changed at all but if this is fixed, it can prevent future regressions in case someone updates this check somewhere and forgets to update it somewhere else or wasn't aware about it.

I would suggest having a member like `isSoftKeybaordShortcutSupported` that's available internally, provide some docs comment explaining why it's only on Android and iOS, and use it in different places, the message doesn't have to be extracted and can be hardcoded, the important thing is that we have one place to keep track of different related checks.

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:dart_quill_delta/dart_quill_delta.dart';
```
EchoEllet, line 1: This is optional but we probably want to import this from `package:flutter_quill/quill_delta.dart`.

So we don't have to change imports from different places in case we changed something related to `dart_quill_delta`.

EchoEllet (changes requested):

example/lib/screens/quill/quill_screen.dart
vishna: It's there for these two to work:

Screenshot 2024-09-18 at 18 37 33

I could change it to:

```dart
import 'package:flutter/foundation.dart';
// ...
softKeyboardShortcutSupport: !kIsWeb && (defaultTargetPlatform == TargetPlatform.android || defaultTargetPlatform == TargetPlatform.iOS)
```

I could also not touch this file at all and have this flag disabled by default

wdyt?

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
vishna: I've extracted this assert to a single place `QuillSoftKeyboardShortcutSupport.assertSupport()`, let me know if this is what you had in mind

example/lib/screens/quill/quill_screen.dart
EchoEllet: You could use:

```
// ignore: implementation_imports
import 'package:flutter_quill/src/common/utils/platform.dart';
```

We probably want to have `package:flutter_quill/internals.dart` at some point in a separate PR.

Since the `isAndroidApp` and `isIosApp` is internal APIs.

This API shouldn't used in the example, since the example is an example for users.

Should use `defaultTargetPlatform` from Flutter `foundation.dart` instead:

```dart
import 'package:flutter/foundation.dart';
```

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
```diff
@@ -80,4 +80,10 @@ class QuillSoftKeyboardShortcutSupport {
 
     return null;
   }
+
+  /// helper method that should indicate wrong use of this class
+  static void assertSupported() {
+    assert(isAndroidApp || isIosApp,
+        'softKeyboardShortcutSupport should only be used on Android/iOS');
+  }
```
EchoEllet, line 88: What I was trying to suggest is to move the `isAndroidApp` and `isIosApp` into their own internal member with an explanation of why it's for iOS and Android as a docs comment for future developers.

To improve readability and also making it easier to keep track of usages. The message itself doesn't have to be extracted.

Also, we should probably avoid using assertion in empty function, assertation will be removed in release mode, which cause to call an empty function called `assertSupported`. Causing confusion if there is a stack trace related to this function due to tree shaking in rare cases.

EchoEllet (changes requested):
What I was trying to suggest is to move the `isAndroidApp` and `isIosApp` into it's own internal member with explanation why it's for iOS and Android

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
EchoEllet: Thank you for the update. I would suggest `isSoftKeybaordShortcutSupported`.

EchoEllet: Solving this conversation to move into the other.

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
vishna: k, redone - hopefully it's better now. the assert string is extracted to one place in case someone wants to improve on the copy.

vishna:
@EchoEllet @CatHood0  I've started wondering about one thing in my PR:

https://github.com/singerdmx/flutter-quill/blob/27343d56ba7f5b911bdc7b6d1cd2ba5f2f272dfd/lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart#L37-L38

Does delta diff gets more costly the bigger the doc is? Is there maybe a better way to detect this key press than doing a diff on the delta? Obviously key press events are not available on mobile but is there maybe some API of intercepting delta inserts? Could this maybe be more efficient this way? 🤔

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
EchoEllet: > k, redone - hopefully it's better now. the assert string is extracted to one place in case someone wants to improve on the copy.

Thank you for the update.

lib/src/editor/raw_editor/config/events/soft_keyboard_shortcut_support.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'space_shortcut_events.dart';
+
+part 'soft_keyboard_shortcut_support_internal.dart';
+
+/// Provides space/character event emulation on platforms that are not
+/// equipped with hardware keyboard (e.g. Android/iOS).
+///
+/// Emulation is based on diff between delta, so it happens after the change
+/// is committed to the document rather than on key press.
+class QuillSoftKeyboardShortcutSupport {
```
EchoEllet, line 18: To double check, is this file exported? If yes, then we probably want to move the `isSupported` and the message to the internal file instead.

vishna: This file is not exported, end developer should only see the editor configurations flag

lib/src/controller/quill_controller.dart
```diff
@@ -322,6 +315,7 @@ class QuillController extends ChangeNotifier {
     TextSelection? textSelection, {
     bool ignoreFocus = false,
     bool shouldNotifyListeners = true,
+    bool isInputClient = false,
```
EchoEllet, line 318: I'm not sure about this new API change, we're breaking a lot of the newly introduced APIs with a lot of breaking changes after a while.

In general, I would avoid introducing any new API, the project is not in a stable state and is not scalable enough to handle minor changes. At least adding new features internally requires efforts only from the maintainers & contributors side. It might seems that I unnecessarily worry too much about such changes but we're getting a lot of reports (even outside GitHub) about the stability and it's kind of a known issue if you search on Google, don't want to have more on the surface for more issues.

I would highly suggest that we focus on the existing features and public APIs instead of more.

Same for other changes here that add more to the public API.

vishna:
@EchoEllet @CatHood0 I've removed delta diffing from my PR, it should make things faster 🤞 Now to detect if we're getting a new character:

- added a flag to differentiate if the `replaceText` is called by an input client
https://github.com/singerdmx/flutter-quill/blob/e27ba2a12aef1f0f33328fafa08b91fd35a68cbe/lib/src/controller/quill_controller.dart#L311-L319

- checking if the incoming delta is single character insert
https://github.com/singerdmx/flutter-quill/blob/e27ba2a12aef1f0f33328fafa08b91fd35a68cbe/lib/src/controller/quill_controller.dart#L331-L333

- passing relevant delta to the logic I've written so far at the end of this method
https://github.com/singerdmx/flutter-quill/blob/e27ba2a12aef1f0f33328fafa08b91fd35a68cbe/lib/src/controller/quill_controller.dart#L388-L395

I probably should be able to make `_softKeyboardShortcutSupport!.onNewChar` static and get rid of the _softKeyboardShortcutSupport instance from QuillController

lib/src/controller/quill_controller.dart
vishna: Would something like this be acceptable?

```dart
  void replaceText(
    int index,
    int len,
    Object? data,
    TextSelection? textSelection, {
    bool ignoreFocus = false,
    bool shouldNotifyListeners = true,
  }) {
    replaceTextInternal(
      index,
      len,
      data,
      textSelection,
      ignoreFocus: ignoreFocus,
      shouldNotifyListeners: shouldNotifyListeners,
    );
  }

  @internal
  void replaceTextInternal(
    int index,
    int len,
    Object? data,
    TextSelection? textSelection, {
    bool ignoreFocus = false,
    bool shouldNotifyListeners = true,
    bool isInputClient = false,
  })
```

EchoEllet: It would complicate the maintainability for contributors and maintainers. Can you use the `experimental` annotation instead for everything new to the public API?

vishna: marked `QuillEditorConfigurations.softKeyboardShortcutSupport` and `isInputClient` as `experimental`.

this I guess becomes a breaking change (if you e.g. extend from QuillController and override replaceText ) 😬

EchoEllet: Is `isInputClient` only meant to be used internally to support this feature? We probably also want to annotate it with `internal` (keep the `experimental` annotation)

lib/src/controller/quill_controller.dart
```diff
@@ -301,26 +301,7 @@ class QuillController extends ChangeNotifier {
...
-  @internal
-  void replaceTextInternal(
-    int index,
-    int len,
-    Object? data,
-    TextSelection? textSelection, {
-    bool ignoreFocus = false,
-    bool shouldNotifyListeners = true,
-    bool isInputClient = false,
+    @experimental bool isInputClient = false,
```
EchoEllet, line 304: If this property is only added to support this feature and meant to be used internally or optional for using this feature, we probably also want to add `internal` annotation (instead of having `replaceTextInternal`)

vishna: done

lib/src/controller/quill_controller.dart
vishna: done - added `internal` there

lib/src/editor/config/editor_configurations.dart
```diff
@@ -95,6 +95,7 @@ class QuillEditorConfigurations extends Equatable {
     this.readOnlyMouseCursor = SystemMouseCursors.text,
     this.onPerformAction,
     this.customLeadingBlockBuilder,
+    this.softKeyboardShortcutSupport = false,
```
EchoEllet, line 98: For the analysis to work, we need to add the `@experimental` here as well, adding it to the variable itself will only work if we have an instance of that object and we're accessing it, but not when creating `QuillEditorConfigurations` using the constructor.

lib/src/editor/widgets/keyboard_service_widget.dart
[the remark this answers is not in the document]
EchoEllet: The new code seems to work the same as:

```dart
assert(QuillSoftKeyboardShortcutSupport.isSupported,
        QuillSoftKeyboardShortcutSupport.assertMessage);
```

Is this part outdated with the `QuillSoftKeyboardShortcutSupport.isSupported`?

lib/src/editor/widgets/keyboard_service_widget.dart
```diff
@@ -71,6 +71,12 @@ class QuillKeyboardServiceWidget extends StatelessWidget {
   }
 
   KeyEventResult _onKeyEvent(node, KeyEvent event) {
+    if (controller.editorConfigurations.softKeyboardShortcutSupport) {
+      assert(isAndroidApp || isIosApp,
+          'softKeyboardShortcutSupport should only be used on Android/iOS');
```
EchoEllet, line 76: Same question as [previous comment](https://github.com/singerdmx/flutter-quill/pull/2235/files#r1767045164).

AtlasAutocode (commented):
I do not like these types of markdown conversions and soft key expansions.
However, I rarely use mobile and will not use markdown conversion.

I agree with @EchoEllet that there are enough questions about this topic, and its impact on our users, that this should at least be marked as experimental until it can be established that there are no unwanted side effects.

singerdmx:
@EchoEllet I will leave this PR to you regarding whether to merge or when to merge

lib/src/controller/quill_controller.dart
```diff
@@ -65,8 +71,14 @@ class QuillController extends ChangeNotifier {
...
-  set editorConfigurations(QuillEditorConfigurations? value) =>
-      _editorConfigurations = document.editorConfigurations = value;
+  set editorConfigurations(QuillEditorConfigurations? value) {
+    _editorConfigurations = document.editorConfigurations = value;
+
+    if (editorConfigurations.softKeyboardShortcutSupport) {
+      assert(QuillSoftKeyboardShortcutSupport.isSupported,
+          QuillSoftKeyboardShortcutSupport.assertMessage);
+    }
+  }
```
EchoEllet, line 81: This check should be only in debug mode, the assert will be removed but the check itself will still run and do nothing, don't know how exactly the tree shaking will handle that kind of checks but it seems that the compiler can guarantee to strips out the code using the following way:

```dart
if (kDebugMode) {
// Your debug code
}
```

Or:

```dart
assert((){
// Your debug code here
return true;
}());
```

I do suggest:

```dart
assert((){
if (editorConfigurations.softKeyboardShortcutSupport) {
      assert(QuillSoftKeyboardShortcutSupport.isSupported,
          QuillSoftKeyboardShortcutSupport.assertMessage);
    }
return true;
}());
```

Since it's less error-prone and can be more clear. You could modify the code slightly for improvements.

lib/src/controller/quill_controller.dart
```diff
@@ -298,8 +311,12 @@ class QuillController extends ChangeNotifier {
 
     Delta? delta;
     Style? style;
+    var isNewCharDelta = false;
     if (len > 0 || data is! String || data.isNotEmpty) {
       delta = document.replace(index, len, data);
+      if (isInputClient && len == 0 && data is String && data.length == 1) {
+        isNewCharDelta = true;
+      }
```
EchoEllet, line 319: Can you explain why this change was introduced? If it's specific to this feature then it probably should be in a separate file including the new variable `isNewCharDelta`.

This is an example:

```dart
/// Paste event for the web.
///
/// Will be `null` for non-web platforms.
///
/// See: https://developer.mozilla.org/en-US/docs/Web/API/Element/paste_event
StreamSubscription? _webPasteEventSubscription;

extension QuillControllerWeb on QuillController {
  void initializeWebPasteEvent() {
    _webPasteEventSubscription =
        EventStreamProviders.pasteEvent.forTarget(window.document).listen((e) {
      // TODO: See if we can support markdown paste
      final html = e.clipboardData?.getData('text/html');
      if (html == null) {
        return;
      }
      // TODO: Temporarily disable the rich text pasting feature as a workaround
      //    due to issue https://github.com/singerdmx/flutter-quill/issues/2220
      // pasteHTML(html: html);
    });
  }

  void closeWebPasteEvent() {
    _webPasteEventSubscription?.cancel();
    _webPasteEventSubscription = null;
  }
}

```

We have introduced `_webPasteEventSubscription` in a separate file to not `quill_controller.dart` harder to read and understand.

lib/src/controller/quill_controller.dart
```diff
@@ -353,6 +370,14 @@ class QuillController extends ChangeNotifier {
...
     }
     ignoreFocusOnTextChange = false;
+
+    if (delta != null &&
+        isNewCharDelta &&
+        editorConfigurations.softKeyboardShortcutSupport) {
+      assert(QuillSoftKeyboardShortcutSupport.isSupported,
+          QuillSoftKeyboardShortcutSupport.assertMessage);
+      QuillSoftKeyboardShortcutSupport.onNewChar(delta, this);
+    }
```
EchoEllet, line 380: Same here, should be in debug mode only (the check itself and not what is inside it).

Can you move the check into it's own file with a name explaining the change?

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
```diff
@@ -33,9 +34,17 @@ bool handleFormatByWrappingWithSingleCharacter({
     return false;
   }
 
+  final usesSoftKeyboardShortcut =
+      controller.editorConfigurations.softKeyboardShortcutSupport;
+  if (usesSoftKeyboardShortcut) {
+    assert(QuillSoftKeyboardShortcutSupport.isSupported,
+        QuillSoftKeyboardShortcutSupport.assertMessage);
+  }
```
EchoEllet, line 42: Same as before

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
```diff
@@ -77,6 +86,12 @@ bool handleFormatByWrappingWithSingleCharacter({
     return false;
   }
 
+  // this gets triggered for e.g. '**', meaning we would try to format empty string
+  // therefore wrongly removing '**'
+  if (usesSoftKeyboardShortcut && lastCharIndex == caretPosition) {
+    return false;
+  }
+
```
EchoEllet, line 94: This code seems to be related to a specific feature, shouldn't be here, if we add all requested features into the same file for a few years then it will be much harder to read and maintain.

EchoEllet (changes requested):
Apologize for requesting too many changes many times but we will take code reviews more seriously than before from now on (at least try to). If you want, you can leave it to me and I will address those changes but will probably take longer than expected and I might ask you a few questions.

Thank you for your efforts on this PR.

lib/src/controller/quill_controller.dart
CatHood0: This is why i also suggested the option to ignore this conditional if the platform is desktop or web. Of course, i say this because this way, the user can or cannot disable the feature, without having to disable the others (and without having to verify the platform themselves). Of course, this is just my opinion.

lib/src/editor/config/editor_configurations.dart
vishna: done

lib/src/controller/quill_controller.dart
vishna: applied this suggestion:

```dart
assert((){
if (editorConfigurations.softKeyboardShortcutSupport) {
      assert(QuillSoftKeyboardShortcutSupport.isSupported,
          QuillSoftKeyboardShortcutSupport.assertMessage);
    }
return true;
}());
```

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
vishna: It's specific for the post-insertion detection of the shortcuts which have the caret at a different position. I don't think there will be much more features added here over years. If this line can't be here, I'll need to copy paste the whole set of shortcut classes to achieve desired effect ...which is fine, I just wanted to avoid doing this.

lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
```diff
@@ -218,6 +218,7 @@ mixin RawEditorStateTextInputClientMixin on EditorState
         diff.deleted.length,
         diff.inserted,
         value.selection,
+        isInputClient: true,
```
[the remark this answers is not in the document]
vishna: I use `isInputClient` to differentiate between `replaceText` called by `RawEditorStateTextInputClientMixin.updateEditingValue` and other actors. This allows me to avoid triggering shortcut code if e.g. space is inserted as a result of some other operation.

Once I know the data is coming from the input client, I check what is inserted. If it's a single character, I mark the flag and at the end of `replaceText` method I can trigger the shortcut code (e.g. apply ordered list formatting)

This is more efficient than what I was doing initially (which was diffing every delta + controller listener - the bigger the delta, the longer it takes to calculate the diff).

Maybe there could be some callback API for this 🤔 

```
void onReplacedText(Delta diff, CallSource source); // called at the end of replaceText
```

This alone would probably allow me to implement this feature without any other major modifications to the package core.

vishna:
> Apologize for requesting too many changes many times but we will take code reviews more seriously than before from now on (at least try to). If you want, you can leave it to me and I will address those changes but will probably take longer than expected and I might ask you a few questions.
> 
> Thank you for your efforts on this PR.

No worries, my knowledge of the library internals is limited so I learn as I poke things around. I appreciate all the feedback!

lib/src/editor/raw_editor/config/events/format/format_single_character_handler.dart
EchoEllet: > more features added here over years

Until now we're having more changes to different files when adding more features.

If someone removed all those files of this feature they will have no idea this change made here is related so should be in its own file with a clear name and docs.

lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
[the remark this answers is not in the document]
EchoEllet: > I use `isInputClient` to differentiate between `replaceText` called by `RawEditorStateTextInputClientMixin.updateEditingValue` and other actors. This allows me to avoid triggering shortcut code if e.g. space is inserted as a result of some other operation.
> 
> Once I know the data is coming from the input client, I check what is inserted. If it's a single character, I mark the flag and at the end of `replaceText` method I can trigger the shortcut code (e.g. apply ordered list formatting)
> 
> This is more efficient than what I was doing initially (which was diffing every delta + controller listener - the bigger the delta, the longer it takes to calculate the diff).

The main issue is that we're adding features that are highly specific. For example, we can't provide toolbar buttons to close the app, switch to full-screen mode, or share the document. Instead, we offer the option to allow users to add their own toolbar buttons and even customize the toolbar completely. Other buttons, like bold, italic, and search, are more general and commonly requested features.

Same thing for embedding widgets into the editor.

High-level features should not be included in low-level classes. To be more specific, such features should not be a direct part of the core package (at least in my opinion and that of others). Separation of concerns is important.

I might be wrong, which is why I suggested to mark those new APIs as `@experimental`, so we can correct them at any time without breaking changes but still not how I prefer to do it.

> Maybe there could be some callback API for this

If you can provide a code example of how this will work, it would help clarify the change.

vishna:
@EchoEllet 

- refactored everything soft keyboard related to be in a separate folder -> `soft_keyboard_shortcut_support` (some unavoidable code duplication)
-  added the aforementioned `ReplacedTextCallback` support to `QuillController` (this is minor breaking change, unsure if this can be done any smarter)
-  adapted `quill_screen.dart` to the refactored code

Let me know what you think and where it would be fine to store these `soft_keyboard_shortcut_support` classes. I might be the only person needing these so might not make sense to keep them in this library 😅

EchoEllet:
To be honest, I'm hesitant to merge or reject this change:

> [!NOTE]
> To be clear, we're discussing the project issues that are not related to the issues of this change, but merging
> more changes we're not sure about can make the maintenance process harder and can contribute to them. It's a issue
> We have for a long time. We do appreciate your time and efforts though this is not the main issue.

- All files related to a specific feature should be feature-based and not based on the file type.
We previously added directories such as `models`, `config`, and `widgets`, and files of each feature will be separated across those directories, the config directory of the raw editor and the editor should be the config for the editor itself and not for the newly added feature, this change has quite a lot of changes.
- We do have other tasks to focus on and prioritize, I couldn't understood quite a lot of changes without contacting the contributors who made them. This will be a long process and we will probably need to put the library in maintenance mode, no longer accepting changes, only bug fixes and code cleanup, big projects are doing something very similar, see [gson](https://github.com/google/gson)
- We don't want to introduce something to the public as stable and then remove it after very short while, to clean up a lot of design issues we need to break things instead of introducing new changes that we're not sure about, making it more overwhelmed to maintain. This is not an issue I have, @CatHood0, @AtlasAutocode, and the others are also unable to fix some issues due to the codebase we have even though they put efforts more than what I do since they joined at the start of 2024, take a search at Google (`I fork flutter quill`) and you will find threads and article discussing this, users are moving away from Flutter Quill and I have seen some posts around from there and here, so it's not necessarily a personal issue. You will see quite a lot of forks and other projects that handle things very differently, we do plan on improving this.
- I don't plan on rejecting this PR, if I'm unsure will leave the decision to @singerdmx, @CatHood0, @AtlasAutocode. If this PR was not merged, then it's not necessarily an issue with the PR or the changes, or efforts you made, we do appreciate your help with the project but the issue is not mainly with your changes.
- You did put efforts to abstract this change but introducing low-level code that gives you more control (closer to the core) is not something we can make, our next plan is to review all PRs that have been introduced since 2020, understanding the changes and cleaning them up, since not all contributors will be available to contact, it will be more difficult task, you will find that many of the previous contributors have left the project and decided to move into another project and one of the main reasons are the issues we have, I have tried to avoid saying this for a year but the first step of solving an issue is to identify it. It's very likely we will be unable to solve those issues but we have to at least try, so we can't give promises we can't keep up to. I have tried (didn't put much effort due to the like of time) to fix as many issues as I can in 2023 and many of the users were disappointed since they were expecting too much.
- I think we're being too rush when merging things and make decisions, every decision need to be discussed first before making it for long enough, to be clear this is not an issue with this PR, but when I asked in this [comment](https://github.com/singerdmx/flutter-quill/pull/2235#discussion_r1768464748) is to give me an example of how this low-level API will abstract your implementation, it seems that you introduced changes very quickly.
- Focusing on something that is simple doesn't mean it's simple to execute, you will find that the Dart team still hasn't decided on the `plugin_platform_interface ` regarding it should be removed or maintaining this package, the package itself is very simple, they are more than capable of managing it, they also know the differences if they should remove or maintain it ([Flutter #127396](https://www.github.com/flutter/flutter/issues/127396)). Similar issues in [Compose Multiplatform](https://www.jetbrains.com/lp/compose-multiplatform/)

EchoEllet:
I can't reject this change since you made efforts while working on this change and I didn't make all of this clear from the start, we should have open an issue and then put a warning in `README.md` and `CONTRIBUTING.md` about the maintenance mode or at least indicate that making new features or changes this big should be discussed somewhere in Github discussion or Slack before attempting to make them, otherwise there will be a high chance of rejecting changes.

So I will have to leave the final decision to the other maintainers.

For now, I'm waiting for your response regarding the first point mentioned in the comment above, by "file type", I meant the type of class (whether it's a widget, a model, a data class, etc.).

CatHood0:
I have seen the changes made recently and I am also in doubt whether we should merge them or not. It's quite complex. 

Obviously we thank you for not only taking the time to do it, but also explaining why many parts were added and the changes you made when we requested them. However, a lot of low-level code is involved to make a feature that it is quite likely that mobile users will not use very frequently. 

Above all, it would be even better to create a native package that allows us to listen to keyboard events on mobile devices (which would completely eliminate the need to add all this code just to do the same thing) and create a widget similar to the `Focus` widget it does. The latter is already quite complicated in itself. I tried to do it but I still need time to be able to do it (I'm not very good with Kotlin code and that has taken its toll on me in not being able to create this package).

AtlasAutocode:
> I agree with @EchoEllet that this discussion is not a reflection on the OP and the quality of their PR.
We have the highest regard for his contributions and quality of work.

However, this issue has sparked an important milestone in the future of the Flutter Quill project.

> What is the future of the Flutter Quill project?

We have an opportunity to decide the future of this project:
1. Is this an amateur/hobbyist project where things get added without planning for future maintenance and functions get lost or broken?
2. Or, is this a serious project where programmers can rely on this editor to create projects that are going to be used by millions (hopefully) of end users?

I think this project is a serious project and is worth putting in the effort to make it stable, reliable and dependable.

I agree with @EchoEllet that we need to have a plan. There must be a stable version that people can rely on. It must not have experimental or incomplete functions. There will always be bugs and edge cases where the behavior is incorrect and these need to be fixed on a priority basis. There will always be implementation differences between the platforms supported.

There is also a need to add functionality (because: why are we programmers? - we always want to improve the capabilities of whatever we are working on and it is fun to challenge ourselves the solve impossible problems). But this must not affect the end users who use our programs to do their critical work.

I agree with @EchoEllet 
1. We must make this project 'Stable'
2. We must fix, or address, all outstanding issues - there are 200 unresolved issues.
3. We must add unit tests and even integration tests. (Code coverage would be nice but implementing that is a nightmare IMO.)
4. We must plan, and agree, how new features will be integrated into the project. New features must not be added because one person needs it. Proposed new features must be important to a major proportion of our users. Hooks and callbacks and API are time-tested methods of allowing individual programmers to implement custom actions because, if anything goes wrong, they have the means to fix it without updating the project's code.

I have no idea if this is possible, or even desirable. This community needs to decide.

@EchoEllet has done an admirable job of trying to keep the project organized and understandable. But we all know that we only understand 10% of the current code and many functions are incomprehensible. I have seen many functions that look totally wrong, but I do not know what their purpose is or why the code is used. So, I cannot fix it because I would probably break something.

I understand the complexity and commitment required to achieve this. Implementing this will elevate the project to a whole new level.

> Are we up to the challenge?

EchoEllet:
> I'm not very good with Kotlin code and that has taken its toll on me in not being able to create this package

If this issue is only on mobile, I will probably be able to solve it since I'm already familiar with Kotlin, Java, and the JVM platform. 

iOS implementation (using Swift) is likely very easy to implement, most of the time I take on Android implementation (regarding my recent PR), I'm not familiar with native iOS and it only take me a few minutes from searching to implementing the feature, see the code difference in #2230 for more details. 

I will only need time. Being able to do something doesn't mean we should do it. There are priorities. Updates for Clipboard and Spell checking issues will probably arrive soon.

Regarding the keyboard events, it's also a missing feature.

**This comment will be updated soon.**

vishna:
I made my team aware of the problem - we decided we can live without this soft keyboard feature. It would have been a nice UX addition to our app but it doesn't make sense just to fork everything to have it. That said, I'm closing this issue since there is no point in keeping it open and distracting authors of this package from solving more important issues.

EchoEllet:
I appreciate your understanding, it's unfortunate that we couldn't merge this feature built-in or at least provide a way for users to implement their own custom changes seamlessly, let's assume we have access to native mobile keyboard events, would it require fewer changes to implement?

Looking forward to your future contributions.

EchoEllet:
I have minimized our comments that are unrelated to the PR. My comments were unnecessarily longer than they should and I wasn't aware of this.

EchoEllet:
I have minimized our comments that are unrelated to the PR. My comments recently were unnecessarily longer than they should and I wasn't aware of this.

PR #2239, opened 2024-09-18
Fix: Resolved issue with broken IME composing rect in Windows desktop

agata:
## Description

When using the Japanese IME in the Windows desktop app, the composing region often appears at the beginning of the document instead of at the selection position, which leads to a poor user experience. The issue was caused by assigning a fixed position of 0 when the position calculated from the conversion information was invalid. This issue can be resolved by using the current selection position as the offset instead of fixing it to 0. Additionally, the function to calculate the ComposingRange has been refactored into a separate function.

Before:
https://github.com/user-attachments/assets/3afd158f-73f1-45ba-b56e-56ff4c9b965a

After:
https://github.com/user-attachments/assets/e9246da5-6ed0-416d-b358-36bc83f74c62

## Type of Change

### DIFF
--- a/lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
+++ b/lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
@@ -97,9 +97,19 @@ mixin RawEditorStateTextInputClientMixin on EditorState
     _textInputConnection!.show();
   }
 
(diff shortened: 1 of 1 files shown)


singerdmx:
It seems unable to pass the tests

agata:
I've fixed it.

Issue #2193, opened 2024-09-07
Insert Table don't work in flutter web and flutter web-server


Paiolus:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill version

10.0.0 on all flutter_quill packages

### Steps to reproduce

When you use the "insertTable" button, it throws an error of rendering...

A RenderEditor expected a child of type RenderEditableBox but received a child of type RenderErrorBox.
RenderObjects expect specific types of children because they coordinate with their children during layout and paint. For example, a RenderSliver cannot be the child of a RenderBox because a RenderSliver does not understand the RenderBox layout protocol.
The RenderEditor that expected a RenderEditableBox child was created by:
QuillRawEditorMultiChildRenderObject-GlobalKey#fa320] - MouseRegion - Composited TransformTarget -_SingleChildViewport - IgnorePointer-[GlobalKey#a466d] -
Semantics < Listener +
_GestureSemantics + RawGestureDetector-LabeledGlobalKey<RawGestureDetectorState>#09d2) - Listener - _ScrollableScope +
_ScrollSemantics [GlobalKey#c09c5| - ••
The RenderErrorBox that did not match the expected child type was created by:
ErrorWidget #08827 - Directionality - QuillRawEditorMultiChildRenderObject-[GlobalKey#fa320] - MouseRegion + Composited TransformTarget +
_SingleChildViewport + IgnorePointer-[GlobalKey#a466d] - Semantics + Listener - _GestureSemantics + RawGestureDetector-[LabeledGlobalKey<RawGestureDetectorState>#09d2] - Listener + •••
See also: https://docs.flutter.dev/testing/errors

<img width="1496" alt="Captura de pantalla 2024-09-07 a las 16 57 40" src="https://github.com/user-attachments/assets/18986a43-65e3-4273-b596-74b85e4195f8">
<img width="1504" alt="Captura de pantalla 2024-09-07 a las 16 57 33" src="https://github.com/user-attachments/assets/6201fab6-baf2-401a-a31e-b963138fcce1">
<img width="1499" alt="Captura de pantalla 2024-09-07 a las 16 57 15" src="https://github.com/user-attachments/assets/c1332215-f99e-40d1-ba52-6596a9697fd2">

### Expected results

You expect to insert a table, such as it happens in flutter web:
[...]

Issue #2130, opened 2024-08-21
Trying to update table cells throws exception


CatHood0:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### The question

While I know I'm the one who created this feature, I did it in a rush and with quite a few bugs in the code. I'm currently fixing most of these so that the tables can be usable by all users.

However, one thing I haven't been able to figure out is how to consistently update the tables once they're already embedded. I tried using the example code in [custom_embed_blocks](https://github.com/singerdmx/flutter-quill/blob/master/doc/custom_embed_blocks.md) but to no avail. The `getEmbedNode` method never finds the table.

So far, this is the code that is responsible for updating the tables (it's a bit different as I wanted it to be easier to understand):
```dart
void _updateTable() {
final delta = Delta()
   ..insert({'table': _tableModel.toMap()})
   ..insert('\n');
   final block = BlockEmbed.custom( CustomTableEmbed.fromDocument(Document.fromDelta(delta)), );
   final offset = getEmbedNode(widget.controller, widget.controller.selection.start).offset;
       widget.controller.replaceText(offset, 1, block, TextSelection.collapsed(offset: offset));
}
```

PR #2254, opened 2024-09-20
Chore: deprecate embed table feature

CatHood0:
## Description

It was decided to disable embed tables for the moment.

Why was this decision made? It's not hard to see. Not only are there notable visual errors, but also in the logic itself of adding columns, rows, and even where the popup that allows us to add new tables to the document should appear.

This feature was implemented at a very early stage and should not have been allowed to be implemented as if it were a totally stable feature without bugs (or at least not many). The best thing for the moment, instead of removing it, is to disable it so that the end user does NOT have access to it and allows us to improve it to a point where we can verify that it is stable and can be implemented along with the other features.

Captura de pantalla_2024-09-20_19-01-07

Captura de pantalla_2024-09-20_19-01-25

A PR will be coming soon to try to improve this implementation. However, there are priorities and we have to solve the problems that most affect the editor (such as the problem with the indent attribute and even the performance, which is quite bad on web and mobile).

## Related Issues

- *Related  #2193*
- *Related #2130*

### DIFF
--- a/flutter_quill_extensions/lib/src/editor/table/table_embed.dart
+++ b/flutter_quill_extensions/lib/src/editor/table/table_embed.dart
@@ -9,7 +9,8 @@ import 'table_cell_embed.dart';
 import 'table_models.dart';
 
 @experimental
-@Deprecated('CustomTableEmbed will no longer used and it will be removed in future releases')
+@Deprecated(
+    'CustomTableEmbed will no longer used and it will be removed in future releases')
 class CustomTableEmbed extends CustomBlockEmbed {
   const CustomTableEmbed(String value) : super(tableType, value);
 
@@ -223,6 +224,7 @@ class _TableWidgetState extends State<TableWidget> {
         if (key != 'id') {
           final columnId = key;
           final data = value;
+          // ignore: deprecated_member_use_from_same_package
           rowCells.add(TableCellWidget(
             cellId: rowKey,
             onTap: (node) {
--- a/flutter_quill_extensions/lib/src/editor/table/table_embed.dart
+++ b/flutter_quill_extensions/lib/src/editor/table/table_embed.dart
@@ -9,6 +9,7 @@ import 'table_cell_embed.dart';
 import 'table_models.dart';
 
 @experimental
+@Deprecated('CustomTableEmbed will no longer used and it will be removed in future releases')
 class CustomTableEmbed extends CustomBlockEmbed {
   const CustomTableEmbed(String value) : super(tableType, value);
 
@@ -45,7 +46,9 @@ class QuillEditorTableEmbedBuilder extends EmbedBuilder {
   }
 }
 
-@Deprecated('TableWidget will no longer used and it will be removed in future releases')
+@experimental
+@Deprecated(
+    'TableWidget will no longer used and it will be removed in future releases')
 class TableWidget extends StatefulWidget {
   const TableWidget({
     required this.tableData,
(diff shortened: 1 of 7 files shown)


EchoEllet:
Can you mark it as experimental with the annotation and deprecated as well? We do want to give users more time. I'm not saying we should remove it anytime soon but since we don't recommend users use it in production applications, we should

CatHood0:
> Can you mark it as experimental with the annotation and deprecated as well? We do want to give users more time. I'm not saying we should remove it anytime soon but since we don't recommend users use it in production applications, we should

That's exactly what I just did (since my PC is very slow it takes more than 10 minutes just to run before_push.dart script)

flutter_quill_extensions/lib/src/editor/table/table_embed.dart
```diff
@@ -34,13 +37,15 @@ class QuillEditorTableEmbedBuilder extends EmbedBuilder {
...
     final tableData = node.value.data;
+    // ignore: deprecated_member_use_from_same_package
     return TableWidget(
       tableData: tableData,
       controller: controller,
     );
   }
 }
 
+@Deprecated('TableWidget will no longer used and it will be removed in future releases')
```
CatHood0, line 48: I'll add experimental mark here

flutter_quill_extensions/lib/src/flutter_quill_embeds.dart
```diff
@@ -119,6 +120,8 @@ class FlutterQuillEmbeds {
     QuillToolbarVideoButtonOptions? videoButtonOptions =
         const QuillToolbarVideoButtonOptions(),
     QuillToolbarCameraButtonOptions? cameraButtonOptions,
+    @Deprecated(
+        'tableButtonOptions will no longer used by now, and probably will be removed in future releases.')
     QuillToolbarTableButtonOptions? tableButtonOptions,
```
CatHood0, line 125: Same for this.

flutter_quill_extensions/lib/src/editor/table/table_embed.dart
```diff
@@ -3,10 +3,12 @@ import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:flutter_quill/flutter_quill.dart';
 import 'package:flutter_quill/quill_delta.dart';
+import 'package:meta/meta.dart';
 import '../../common/utils/quill_table_utils.dart';
 import 'table_cell_embed.dart';
 import 'table_models.dart';
 
+@experimental
 class CustomTableEmbed extends CustomBlockEmbed {
```
CatHood0, line 12: Same for this

CatHood0:
> Update: Can you mark it as experimental to make it more clear why it's deprecated?

I already do it. Can you check if i missing something?

EchoEllet:
> I already do it. Can you check if i missing something?

LGTM.

EchoEllet (approved):

singerdmx (approved):

EchoEllet:
> The support for tables is currently limited and underdevelopment, more changes are expected to arrive. We are actively working on enhancing its functionality and usability. We appreciate your feedback as it is invaluable in helping us refine and expand this feature.

This part in `README.md` is outdated.

CatHood0:
I'll take a look as soon as possible.

EchoEllet:
Updated.

CatHood0:
> Updated.

Thanks for taking the efforts and the time.

Issue #2246, opened 2024-09-19
Rewrite the SpellCheckerService for simplicity and extensibility and support for Android, iOS and macOS


EchoEllet:
We plan on introducing a **breaking change** for `SpellCheckerService` (from `flutter_quill`) and using the Flutter [`SpellCheckerService`](https://github.com/flutter/flutter/blob/568ed33da9c2e44c7c46d944771a5afc58dd2241/packages/flutter/lib/src/services/spell_check.dart#L101C16-L101C33) or extending it instead of having our own custom `SpellCheckerService` which we discovered that it has some issues:

- **Extensibility**: The interface should be more generic and less specific to client-side spell-checking service, it should be focused as a source of checking the spelling itself and not specific to Flutter `TextSpan` or having some methods such as `dispose`, `setNewLanguageState`, `updateCustomLanguageIfExist`, `addCustomLanguage`, those methods seems to be helpers for managing a client-side service that is on the app itself, the user may want to use a web API that provides this service [Grammarly API](https://developer.grammarly.com/) or use method channel, it's also not a `Future` which is expected to be by the user. The service itself should not be responsible for customizing the UI. Take a look at `TextField` to see how it's done in Flutter (See [SpellCheckConfiguration](https://api.flutter.dev/flutter/widgets/SpellCheckConfiguration-class.html)). The service will be activated in the `QuillEditorConfiguration` instead of having the `SpellCheckerServiceProvider`.
- **Bundle size**: We have introduced a package that has supports for too many languages all at once, we did initially wanted to separate each language into its own package but this would complicate the process and require more efforts when using the package since the user needs the core package and a package for each language which is a common solution is used by VS Code and JetBrains IDEs, in general even when using one language (English) it still increase the bundle size quite significantly, which require us to make this feature completely optional, having our own package that integrates with `flutter_quill`, the project is currently in the process of a changes more frequently than before and there are priorities, we need to fix existing issues before introducing something that leads to more issue reports. Usually, this is a bigger issue on web.
- **Web Support**: Even though it's likely this issue will not be fixed, it's currently experiencing issues on some platforms so we probably want to make it experimental. See (#2196)
- **Not using the native API as default**: This will be fixed with platform-specific API using `quill_native_bridge` (or at least attempt to). We do plan on supporting it on:

  **Android**: Using [`android.view.textservice.TextServicesManager`](https://developer.android.com/develop/ui/views/touch-and-input/spell-checker-framework)
[...]

PR #2255, opened 2024-09-20
chore: deprecate spell checker service

EchoEllet:
## Description

*Deprecate `SpellCheckerService` as a breaking change will arrive soon.*

Make it clear to the users that the `SpellCheckerService` should not be used as breaking change will be arriving soon, we provided an implementation in the example only (with the usage of [simple_spell_checker](https://pub.dev/packages/simple_spell_checker)) and mentioned that this is not a feature implemented in the project, which doesn't mean we can remove the `SpellCheckerService` right away. We should at least give the users a few weeks (which is not enough) if we're going to introduce this change in a minor breaking change release (`10.7.0`).

Should also be mentioned that, will cause issues for most projects that use it since [pub.dev](https://pub.dev/) ([pub_semver](https://pub.dev/packages/pub_semver)) expects the version to follow the same format and guidelines as [semver](https://semver.org/). Some projects use tools to automate the process of updating the dependencies to the latest non-breaking release and then publishing a new version (sometimes).

`flutter pub upgrade --major-versions` will update them even if breaking changes are introduced. `flutter pub upgrade` will update to the latest non-breaking release. It assumes that `10.7.0` is not a breaking change.

Temporarily removed the docs as it can be used to indicate that this is a stable feature. We do want to limit the damage that can be caused by fixing #2246.

## Related Issues

- *Related #2246*
- *Not directly related #2235*

## Type of Change

## Suggestions

I highly suggest marking new features (including #2230) as experimental, marking them as experimental doesn't necessarily mean the feature is likely to be broken or incomplete but more like this is not stable and can be changed in a way that users will also have to update their code in order to use newer versions. This gives us the opportunity to fix issues without having to worry about releasing a new major breaking release.

We should be more direct and straight with the users to explain the breaking changes if having minor breaking changes such as updating the name of something (or correct misspellings, [example without breaking change for internal use only class](https://github.com/singerdmx/flutter-quill/blob/f3089c58ce61b7a62c00e7c1ff8da3bb9ac810bd/lib/src/editor_toolbar_controller_shared/clipboard/clipboard_service_provider.dart#L10-L14)) in minor versions is acceptable (while mentioning this in `README`)?

In all cases we should be more cautious about adding new APIs and should be at least marked as experimental, I prefer that we don't add any new features for the next year (or even more) once we fix all issues and code quality issues. This will be a long way and we don't want to give promises we can't keep up to. So no guarantee that this will be worth it.

But having more incomplete features (which is very often) with more issues, and even more code quality issues doesn't seem to be a solution at all, if you disagree, you might see example projects are moving away from the project. Fixing this is simple, which doesn't mean it's simple to execute.

We had more discussion about this in #2235.


CatHood0 (approved):
Thank you for taking the time to deprecate this feature. For now these changes are more than enough because `SpellCheckerService` is in a very early stage. And as you mentioned, it's best to release these new features only when they are stable (which in this case is the least likely, since I couldn't provide a more compact solution that would fit the needs of the developers).

For now I'll take the time to start fixing problems instead of adding new features, since so many new things really add more complexity to the project.

EchoEllet:
Thanks. Users should still be able to use [simple_spell_checker](https://pub.dev/packages/simple_spell_checker) even after the breaking change.

CatHood0:
> Users should still be able to use [simple_spell_checker](https://pub.dev/packages/simple_spell_checker) even after the breaking change.

I think that's fine, but I'm planning to break changes anyway because i need to improve the current implementation of it, which is not only slow, but cumbersome and sometimes doesn't even fulfill its basic function. It will take some time anyway. Thanks for taking the time and effort.

singerdmx (approved):

EchoEllet:
> I think that's fine, but I'm planning to break changes anyway because i need to improve the current implementation of it, which is not only slow, but cumbersome and sometimes doesn't even fulfill its basic function. It will take some time anyway. Thanks for taking the time and effort.

If you're going to introduce breaking changes and in case you have the time, you probably want to separate support of each language into its own package, use [melos](https://invertase.docs.page/melos) to automate the publishing since it's can be time-consuming.

There are other things to consider such as web support, how will these files will be supported? especially for the web, don't know how Flutter Web loads files but shouldn't be more memory performance efficient when loading them if they were file assets instead? Do they slow down the app loading on the web if deployed?

Of course, it might not be a priority until now or you might not have the time at the moment for those changes.
But those kinds of issues you probably want to address all at once when introducing a major breaking change in a major version, instead of multiple major versions like what we have here.

CatHood0:
> If you're going to introduce breaking changes and in case you have the time, you probably want to separate support of each language into its own package, use [melos](https://invertase.docs.page/melos) to automate the publishing since it's can be time-consuming.

Thanks a lot for the suggestion, I'll probably use it. And I guess since you mentioned making a package for each language, it gave me an idea of ​​how I'll do this.

> There are other things to consider such as web support, how will these files will be supported? especially for the web, don't know how Flutter Web loads files but shouldn't be more memory performance efficient when loading them if they were file assets instead? Do they slow down the app loading on the web if deployed?

Apparently the website is able to load the dictionary files, but since they are large (normally, 80 MB is too much for such a small feature) it takes longer for a page to load.

> Of course, it might not be a priority until now or you might not have the time at the moment for those changes.
> But those kinds of issues you probably want to address all at once when introducing a major breaking change in a major version, instead of multiple major versions like what we have here.

Yes, for now I will keep the package the same while I create an implementation that is more in line with what is required and that is used by developers.

I say this because I have noticed that some people make forks to disable all other languages ​​and leave only the one they will use. This gives me the impression that, as you mention, having languages ​​in separate packages would allow all developers to have only the languages ​​they are interested in instead of having everything (which they will probably only use one or two of those dictionaries)

Thank you very much for your recommendations. I really appreciate the help.

singerdmx:
Can we please merge some of PRs before making more?

PR #2258, opened 2024-09-21
chore: deprecate flutter_quill/extensions.dart

EchoEllet:
## Description

*Deprecate usages of `extensions.dart` to indicate that it's for internal usage only. Add `flutter_quill_internal.dart` as a replacement.*

I have seen that some users are using the `extensions.dart` directly in their project which is a file for `flutter_quill_extensions` only and contains internal APIs that will change and update for the extensions package without any notice, which introduce breaking changes, users are often don't see the file itself from source and use it directly. Adding a comment is not clear enough. The `flutter_quill_internal` makes sure to make it more clear that it's for internal usage only.

Should increase the minimum version of `flutter_quill_extensions` by `flutter_quill`.

## Type of Change

## Suggestions

For now, avoid introducing APIs as part of the public API, make it more clear it's for internal usage only using docs comment and `internal` annotation.


EchoEllet:
It seems that we're using the `extensions.dart` that exposes the internal API to `flutter_quill_extensions` inside of `flutter_quill`. The new name should avoid confusion and be more specific.

Which is:

```dart
import 'package:flutter_quill/flutter_quill_internal.dart
```

Can also be:

```dart
import 'package:flutter_quill/internal.dart
```

Issue #2211, opened 2024-09-11
Improve quality and format of CHANGELOG.md


EchoEllet:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Use case

The `CHANGELOG.md` is getting bigger and we do automate the process of generating release notes, it might not be very clear what's added, removed, or changed.

Having a readable `CHANGELOG.md` while still automating the process of generating release notes for each release
require to enforce the use of conventional commits in the PR title and using [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)

### Proposal

I do suggest that we:

- Update the format of`CHANGELOG.md` to apply to [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) first by automating this task.
- Add a new check in the main GitHub workflow to check if the title confirms conventional commits.
- When publishing a new version, extract and generate the changes in a new version by using a tool that gets them from the PRs titles and integrates with **Keep a Changelog** format and [Conventional Commits
](https://www.conventionalcommits.org/en/v1.0.0/), we can use GitHub action (e.g. [extract-release-notes](https://github.com/marketplace/actions/extract-release-notes), [changelog-version-extractor](https://github.com/marketplace/actions/changelog-version-extractor)) or dart package (for dev only) such as [`change`](https://pub.dev/packages/change).
- To allow adding notes, the workflow will get the notes of the GitHub release created by the maintainer manually and put them at the top and will be optional, the changes will be automated

The GitHub generate release notes button will be no longer required as it doesn't use a specific format and it's what we have been using for automating the changes.

EchoEllet:
I will start working on this soon, meanwhile, we can discuss this change before process further.

PR #2262, opened 2024-09-21
chore: remove pubspec_overrides.yaml and pubspec_overrides.yaml.disabled

EchoEllet:
## Description

*Remove and disable the local development mode completely.*

The file `pubspec_overrides.yaml` exist so we can override the dependencies to use the local ones instead from pub.dev

We usually don't want that in production, so it's only during development to allow contributors to focus on the changes instead having to update the minimum change of `flutter_quill` by first introducing the internal breaking changes, upload a version so we can use it in the version after it as a minimum for `flutter_quill_extensions` (See [10.7.2](https://github.com/singerdmx/flutter-quill/releases/tag/v10.7.2)) as an example.

Contributors had to include the `dependency_overrides` and then remove it before pushing a commit, sometimes they push a commit to test the changes in CI. I introduced the `pubspec_overrides.yaml.disabled` which will be ignored by pub.dev (as it uses `pubspec_overrides.yaml` instead) and two scripts which will create `pubspec_overrides.yaml` files and delete them (which is ignored in `.gitignore`).

I don't see a strong reason for this feature, if we need to fix something we should do it manually and ensure a compatible minimum version requirement, we also usually test using the example app so it's usually enough to have it there.

Also, I see some other issues of having only one version for all packages, `dart_quill_delta` should have its own version separated from `flutter_quill`. Same to `flutter_quill_test` as it didn't introduce any changes. The `flutter_quill` and `flutter_quill_extensions` might or might not require only one version.

- When we want to introduce breaking changes to one package, we have to break them all
- When a minor version made to a package, we have to publish them all over again
- They will all share the same `CHANGELOG.md` which is not very organized (see #2211).
- It gives us some limitations when introducing internal breaking changes. We will always have to make some internal changes backward compatible between `flutter_quill` and `flutter_quill_extensions`. When a new API introduced in `flutter_quill` and it's needed in `flutter_quill_extensions`, with our current workflow, we have to release it in backward compatible way even if it's an internal API first, once we have the new version of `flutter_quill`, we need another release that increases the minimum version of `flutter_quill` in `flutter_quill_extensions`
- The current workflow doesn't give us the flexibility when publishing the packages in the correct order, there is no specific order that we must stick to.
- We don't want to fix this issue between `flutter_quill` and `flutter_quill_extensions`, when a breaking change is introduced, users expect improvements and new features to the final product, so we should enhance `flutter_quill_extensions` and once there is a strong reason to introduce a breaking change by fixing issues, introducing features and more tests then we will probably introduce it, fixing it requires a separate package of `flutter_quill_extensions` and that's not a priority at the moment.
- This whole workflow is not scalable enough to have `quill_native_integration` with its own packages (`quill_native_integration_platform_interface`, `quill_native_integration_web`, `quill_native_integration_windows` etc...). See #2230 for more details.
- We don't want complexity that doesn't fix a issue, especially with our current situation.

As for `quill_native_bridge`, I will also separate the version, I need to publish a new package and deprecate this, luckily it's already experimental and for internal use so shouldn't be a breaking change as we didn't expose the APIs of it yet (see #2230). The new package will probably be `quill_native_integration`.

> [!TIP]
> This PR only removes thee local development mode and those two files in addition to their scripts and doesn't fix or change all of the points mentioned above. The publishing will still be the same.

## Type of Change

- [x] ✅ **Build configuration change:** Changes to build or deploy processes.


EchoEllet:
@CatHood0 , @AtlasAutocode , @singerdmx 
If you also think having this feature of local development mode is unnecessary and can be confusing, will merge this PR to remove it.

CatHood0 (approved):
I have found it useful when adding new changes to, for example, `flutter_quill_extensions` and `flutter_quill` and ensuring that any mutual changes are not ignored. However, it is confusing in itself since it is almost magical to run a script and be able to have the local changes reflected without having to update one part and then the other part that needs to have that new change.

In any case, it seems to me to be quite useful, but only in some cases. Most of the time it doesn't help much as we are more focused on fixing bug or stability issues than adding new features that require local mutual changes to be reflected.

EchoEllet:
At some point, we may consider using tools like [melos](https://invertase.docs.page/melos) to manage multiple packages, until know shouldn't be an issue.

EchoEllet:
See an example issue caused after merging this change to [#2230](https://github.com/singerdmx/flutter-quill/actions/runs/10973600775/job/30471167349). We need to publish changes of `quill_natibe_bridge` before proceeding further.

This will ensure such issues will not be encountered by users, as we may forget to update the minimum required version.

EchoEllet:
An alternative solution is to add `pubspec_overrides.yaml` temporarily only when needed and revert later once is no longer required. See [this commit](https://github.com/singerdmx/flutter-quill/pull/2230/commits/d89073a3c8b7a63e731aec590d005532e69fab78) as an example.

This will also help ensure that we do publish the package that needs to be published first before updating the other and bump the minimum version of that package in the dependencies.

CatHood0:
> An alternative solution is to add pubspec_overrides.yaml temporarily only when needed and revert later once is no longer required. See this commit as an example.

Thanks for mentioning this. I'll keep it in mind for when I need it.

EchoEllet:
> I'm able to reproduce the issue on my local machine

It seems that [melos](https://pub.dev/packages/melos) does address that kind of issue more effectively. See [Why do I need to bootstrap](https://invertase.docs.page/melos~melos-latest/getting-started#why-do-i-need-to-bootstrap)

CatHood0:
> It seems that [melos](https://pub.dev/packages/melos) does address that kind of issue more effectively.

For now I'll first look at how to use [melos](https://pub.dev/packages/melos) and slowly implement it in my packages. Once I'm confident that I can configure it without messing anything up, I'll open an issue or discussion to implement melos in the package.

EchoEllet:
Sure, before introducing this solution, I planned on trying to use `melos` but didn't since it required contributors to install it, which is an additional step that I wanted to avoid.

PR #2264, opened 2024-09-21
revert: "Resolved issue with broken IME composing rect in Windows desktop"

### LINKED ISSUE #2253: Applying indent attribute throws `container.containsOffset(position.offset)` fail assertion

### Flutter Quill version

10.6.5

### Steps to reproduce

* Insert a header at the beginning of the document (write whatever you want).
* Create a new line.
* ​​Below the new line you created in the previous step, write some text.
* Apply the `indent` attribute to this last line (any that is not the first one) and you will see how this happens.

### Expected results

Apply indent to any line without exception calls.

### Actual results

When apply indent to any line that is not the first one, `text_line` throws an exception by the container is not registering the current offset position.

### _This is outdated. By now we found now the cause of the issue. Please scroll to the last comments to see which is the reason of this issue existence._

It seems that when we try to indent a line `text_line` throws the typical exception "`container.containsOffset(position.offset)`" which seems to tell us that the node we are selecting is not at the offset that the position is telling the editor.

But, I was wrong, I thought that the `Offset` was reloaded, instead of keeping the values ​​it had before applying the indentation attribute.

it seems that there is a problem with the `childAtPosition` function of `RenderEditableContainerBox` which compares in a while if the target we currently have corresponds to the node found through the `queryChild` method. But, is this not right? Yes, but it seems that there is an error at the level of `equals` and `hashcode` (this last one is still a theory that I am testing).

This would be the function in question: 

```dart
RenderEditableBox childAtPosition(TextPosition position) { 
  assert(firstChild != null);
  final targetNode = container.queryChild(position.offset, false).node;

  var targetChild = firstChild;
  while (targetChild != null) { 
    /// The issue is here with the equality
    if (targetChild.container == targetNode) { 
       break;
    } 
    final newChild = childAfter(targetChild);
    if (newChild == null) { 
       // At start of document fails to find the position 
       // here is where the issue start
       targetChild = childAtOffset(const Offset(0, 0));
       break;
    }
    targetChild = newChild;
  } 

  if (targetChild == null) { 
     throw 'targetChild should not be null';
  } 
  return targetChild;
}
```

####  What happens with this? 

Let's take an example, let's assume that the editor has a text such that (I'll use `markdown` syntax to not complicate the example using `Deltas`):

```markdown
# First header

any content
```

Supposing that we are right at the beginning of the third line (i.e. where "any content" starts), the caret should be with a `TextPosition` with an offset of 13. Having this value now, if we apply indent to the third line, the error will now occur. Why? Obviously the position was not updated neither in `QuillController's` `updateSelection` nor much less from our `raw_editor_state_text_input_client_mixin` since the latter is the one that maintained this last value.

Since we mentioned our `text_input_client` we have to mention that when we apply the indent, it goes and calls `_updateComposingRectIfNeeded` which our editor calls `getLocalRectForCaret` to get the current `Rect` of the caret. Why do I mention `getLocalRectForCaret`? This is because when it is called, then at the same time, in order to have which node is selected it calls `childAtPosition`. Do you see the problem? Every time we update the position of the caret, the node is searched, at this point there is no problem.

But, it seems that when we apply indent, something changes inside the editor, which is not able to determine if the node we obtained is really the one found through the `targetChild` of the code sample I showed before. It's strange, isn't it? Even though `queryChild` DOES find the node, when the loop is executed and we compare the `RenderEditableBox` container, it is not able to know if it is really equal or not. This only happens if indent is applied. With any other attribute it doesn't seem to be happening (although I've noticed that link sometimes suffers from this problem as well).

Concluding with this, it seems that since it is not able to properly compare the node obtained (which is the one we are actually selecting) with the node that `childAtPosition` returns, then it throws an error. This is mostly because since it is not able to obtain any new value that satisfies the condition, then it takes the first node it finds from the document and returns it, which causes this problem, since it is clear that it is more than likely that the first node is not or does not contain the offset in which we are.

I have a theory that the `RenderEditableBox` isn't updated fast enough so that when the editor gets it, it can compare the new value to the old one.

Having said that, it looks like we'll have to figure out if this is just an equality issue or something like one part being updated way before the other.

https://github.com/user-attachments/assets/f00d5df8-f035-4469-ac38-cfb9a665d991

CatHood0:
## Description

These changes were reverted because they are causing issues when using the indentation attribute. I can't say if they are causing other parts to fail, but they are definitely not allowing the editor to work stably.

@agata  if you are going to redo your PR fixing this IME issue, please make sure to test if the IME works along with everything else in the editor (selection, typing, applying block and inline attributes, etc.) to make sure the change is really stable. Because i spent 3 days looking for the cause of this, and yes, it is a bit exhausting.

_No related: I've noticed that on the web this fails randomly too._

## Related Issues

- *Fix #2253*

## Suggestions

Before merging any new changes into the main release, we should consider whether these changes work completely without damaging the other existing functionality within the package.

We should establish some step-by-step guide before merging which PR to first check whether the changes really do not alter the correct functioning of the editor.

### DIFF
--- a/lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
+++ b/lib/src/editor/raw_editor/raw_editor_state_text_input_client_mixin.dart
@@ -112,19 +112,9 @@ mixin RawEditorStateTextInputClientMixin on EditorState
     _textInputConnection!.show();
   }
 
-  TextRange _getComposingRange() {
-    if (_lastKnownRemoteTextEditingValue != null &&
-        _lastKnownRemoteTextEditingValue?.composing.isValid == true) {
-      return _lastKnownRemoteTextEditingValue!.composing;
-    } else if (textEditingValue.composing.isValid == true) {
-      return textEditingValue.composing;
-    } else {
-      return widget.controller.selection;
-    }
-  }
-
   void _updateComposingRectIfNeeded() {
-    final composingRange = _getComposingRange();
+    final composingRange = _lastKnownRemoteTextEditingValue?.composing ??
+        textEditingValue.composing;
     if (hasConnection) {
       assert(mounted);
       final offset = composingRange.isValid ? composingRange.start : 0;


EchoEllet (approved):

EchoEllet:
The failure has references to methods added that are only available in the branch [feat/default-clipboard-native-impl](https://github.com/singerdmx/flutter-quill/tree/feat/default-clipboard-native-impl)?

CatHood0:
> The failure has references to methods added that are only available in the branch [feat/default-clipboard-native-impl](https://github.com/singerdmx/flutter-quill/tree/feat/default-clipboard-native-impl)? See [QuillNativeBridgePlatform](https://github.com/singerdmx/flutter-quill/blob/master/quill_native_bridge/lib/src/quill_native_bridge_platform_interface.dart), it doesn't have getClipboardImage and other methods introduced in https://github.com/singerdmx/flutter-quill/pull/2230, which is still not merged yet.

I don't really know. why this happens. Should i close this and make the changes in another branch? It's weird. I create this branch from the master branch.

EchoEllet:
I'm able to reproduce the issue on my local machine on the `master` branch, which is still strange, I haven't merged anything yet.

CatHood0:
> I'm able to reproduce the issue on my local machine on the master branch, which is still strange, I haven't merged anything yet.

I don't know what to tell you. This is actually the first time I've seen an error like this.

EchoEllet:
Issue fixed after running `flutter pub get` inside `flutter_quill` (in the main repo directory) and inside `quill_native_bridge`.

image

This keeps getting more interesting.

EchoEllet:
We will move `quill_native_bridge` outside of this repository at some point, integration tests are expensive to run and can take longer than expected since they will run on real devices and will take too much time, also issues of this plugin should be separated.

EchoEllet:
Why are all those commits added? #2265 fixed the issue so once I merge it, you can pass this issue by merging the `master` branch or rebase.

EchoEllet:
Can you revert to this [commit](https://github.com/singerdmx/flutter-quill/pull/2264/commits/5e09f362cea4c86a4986d80c2ff5351b0715d46d)

EchoEllet (approved):
LGTM. Still don't know what causes this pub failure and why #2265 is a fix.

singerdmx:
Are we supposed to merge this despite CI failure?

singerdmx (approved):

EchoEllet:
I thought the issue was fixed when I commented above.

Still don't know why the CI doesn't fail when I sent this PR above. It was already the case before.

@CatHood0 Can you try manually reverting this change or using GitHub web? Also, send a PR to your fork and see if this is still an issue. I will check it into more details tomorrow.

agata:
@CatHood0 I'm sorry for the trouble I caused, and I appreciate your response. Thank you very much. Moving forward, I will conduct more extensive testing before submitting pull requests.

CatHood0:
> Thank you very much. Moving forward, I will conduct more extensive testing before submitting pull requests.

No problem. We are also at fault for not taking the time to test the changes that the PRs make and making sure to document any changes and make any changes without going through any filter. If you can fix the problem with the IME without breaking the normal functioning of the editor, you are more than welcome.

**Not directly related:** try to try a different way of getting the `TextRange`, because if you use the `QuillController` selection this error will probably continue.

EchoEllet:
> I'm sorry for the trouble I caused,

When a contributor submits their changes, the maintainers are fully responsible for any issues caused by their changes since they have the choice to merge or reject, request changes, or edit them directly. We should understand the changes and will request more details in future PRs. Which is why I will be more strict in future PRs code reviews.

EchoEllet:
> I don't know what to tell you. This is actually the first time I've seen an error like this.

FYI: The issue happened because `pubspec.lock` isn’t checked in for the example app, but it should be (see the [.gitignore of the root project](https://github.com/singerdmx/flutter-quill/blob/master/.gitignore#L78)). While packages exclude `pubspec.lock`, apps should include it to lock dependency versions. Running `flutter pub upgrade` updates to the latest compatible versions, but I usually update `pubspec.yaml` manually for apps. Without `pubspec.lock`, everyone uses the latest versions, which is the expected behavior for packages but not apps (https://dart.dev/guides/libraries/private-files#pubspeclock).

Solution: Move `flutter_quill` package into its own directory with a separate `.gitignore`, sharing some rules with the root `.gitignore`.

PR #1937, opened 2024-06-20
fix(editor): implement editor shortcut action for home and end keys to fix exception about unimplemented ScrollToDocumentBoundaryIntent

ellet0:
## Description

*Implement home and end keyboard shortcuts.*

Before implementing this feature, I clicked `Home` and `End` keys in **Linux** and **macOS** you will get this exception:

```console
════════ Exception caught by services library ══════════════════════════════════
Unable to find an action for an Intent with type ScrollToDocumentBoundaryIntent in an Actions widget in the given context.
```

<details>
<summary>Stacktrace</summary>

```console

════════ Exception caught by services library ══════════════════════════════════
The following assertion was thrown during method call TextInputClient.performSelectors:
Unable to find an action for an Intent with type ScrollToDocumentBoundaryIntent in an Actions widget in the given context.
Actions.invoke() was unable to find an Actions widget that contained a mapping for the given intent, or the intent type isn't the same as the type argument to invoke (which is Intent - try supplying a type argument to invoke if one was not given)
The context used was:
Focus(focusNode: FocusNode#f5fe7([PRIMARY FOCUS]), dependencies: [_ActionsScope, _FocusInheritedScope, _InheritedTheme, _LocalizationsScope-[GlobalKey#fa1ef]], state: _FocusState#e49c4)
The intent type requested was:
ScrollToDocumentBoundaryIntent

When the exception was thrown, this was the stack:
#0      Actions.invoke.<anonymous closure> (package:flutter/src/widgets/actions.dart:973:9)
#1      Actions.invoke (package:flutter/src/widgets/actions.dart:987:6)
#2      QuillRawEditorState.performSelector (package:flutter_quill/src/widgets/raw_editor/raw_editor_state.dart:1707:17)
#3      ListBase.forEach (dart:collection/list.dart:51:13)
#4      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1916:19)
#5      TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1794:20)
#6      MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:571:55)
#7      MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:564:34)
#8      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:581:35)
#9      _invoke2 (dart:ui/hooks.dart:344:13)
[...]

PR #2279, opened 2024-09-24
fix(macos): Implement actions for ExpandSelectionToDocumentBoundaryIntent and ExpandSelectionToLineBreakIntent to use keyboard shortcuts, unrelated cleanup to the bug fix.

### LINKED ISSUE #1192: [Desktop] Error when using selection shortcut on Mac
Hello,
I'm getting the following 2 errors when using ```command + shift + arrow``` shortcuts on MacOS:

Error 1:
```Unable to find an action for an Intent with type ExpandSelectionToDocumentBoundaryIntent in an Actions widget in the given context.```
Triggered by ```command + shift + arrow up``` and ```command + shift + arrow down``` (Platform shortcut for moving selection from current position to start or end of text)

Error 2:
```Unable to find an action for an Intent with type ExpandSelectionToLineBreakIntent in an Actions widget in the given context.```
Triggered by ```command + shift + arrow right``` and ```command + shift + arrow left``` (Platform shortcut for moving selection from current position to start of end of line.)

### PR DESCRIPTION
**Pull request by EchoEllet**
## Description

*Implement actions for [ExpandSelectionToDocumentBoundaryIntent](https://github.com/flutter/flutter/blob/2663184aa79047d0a33a14a3b607954f8fdd8730/packages/flutter/lib/src/widgets/text_editing_intents.dart#L147-L162) and [ExpandSelectionToLineBreakIntent](https://github.com/flutter/flutter/blob/2663184aa79047d0a33a14a3b607954f8fdd8730/packages/flutter/lib/src/widgets/text_editing_intents.dart#L164-L181) in addition to unrelated cleanup that is not required to have the bug fix.*

This PR will be updated soon, see #1192 for more details about the issue.

## Related Issues

- *Fix #1192*
- *Related #1937*
- *Related #2288*

`ExtendSelectionToNextParagraphBoundaryOrCaretLocationIntent` is also missing and will throw when pressing (option + shift + arrow up/down).

## Type of Change

## Suggestions

Not for merge yet.

## TODO

- [ ] Explain the changes
- [ ] Add tests, improve docs, and clear the Todos in code.

### CODE REVIEW

**Comment by EchoEllet** [comment]
@CatHood0 I will address the change of
`QuillKeyboardServiceWidget` directly in this PR, worrying about having a clean history, separation, and organization of changes is not our first priority at the moment, I keep forgetting about adding this to the Todo and already forgot many things since I delay them later, my TODO is already full.

The support for shortcuts is not clean enough, some of them are overridable:

```dart
  HideSelectionToolbarIntent:
        _makeOverridable(QuillEditorHideSelectionToolbarAction(this)),
    UndoTextIntent: _makeOverridable(QuillEditorUndoKeyboardAction(this)),
    RedoTextIntent: _makeOverridable(QuillEditorRedoKeyboardAction(this)),
```

Some of them do start with `QuillEditor` prefix and some of them do not:

```dart
    ToggleTextStyleIntednt: _formatSelectionAction,
    IndentSelectionIntent: _indentSelectionAction,
    QuillEditorApplyHeaderIntent: _applyHeaderAction,
    QuillEditorApplyCheckListIntent: _applyCheckListAction,
    // TODO: Double check and see if those and related should be overridable
    QuillEditorApplyLinkIntent: QuillEditorApplyLinkAction(this),
    ScrollToDocumentBoundaryIntent: NavigateToDocumentBoundaryAction(this),
```

Others are not instantiated directly and have private member and the only usage is in the shortcuts:

```
    ToggleTextStyleIntent: _formatSelectionAction,
    IndentSelectionIntent: _indentSelectionAction,
    QuillEditorApplyHeaderIntent: _applyHeaderAction,
    QuillEditorApplyCheckListIntent: _applyCheckListAction,
```

It's not clear how should we name things, we need to agree on something, each class is named very differently without conventions or guidelines, it's also not clear that if we should add the `SingleActivator` (see [#1937 comment](https://github.com/singerdmx/flutter-quill/pull/1937/files#r1773712588)) or if they are already registered with Flutter, or if we're using something somewhere that do register those, why they those actions are missing in the first place? It's causing issues on the desktop with keyboard shortcuts.

In addition to other questions and issues, I didn't register them in the TODOs.

The file `default_single_activator_actions.dart` should also be somewhere inside `raw_editor`.

EchoEllet:
It seems that there are even more important things to clean up, I have also noticed some duplicated logic, and a new update will be soon however it's not related to this PR so it might be confusing to review. I have tried to separate it into a new PR though decided not to.

EchoEllet:
> or if we're using something somewhere that do register those, why they those actions are missing in the first place

It seems that we're using [`intentForMacOSSelector()`](https://github.com/flutter/flutter/blob/stable/packages/flutter/lib/src/widgets/default_text_editing_shortcuts.dart#L587-L648) from Flutter inside the `raw_editor_state.dart` without implementing the actions for each intent.

image

image

We need to file an issue for this and fix them all in a clean way so we don't lost track of this.

lib/src/editor/raw_editor/keyboard_shortcuts/default_single_activator_intents.dart
```diff
@@ -0,0 +1,191 @@
...
+    SingleActivator(
+      LogicalKeyboardKey.home,
+      control: !_isDesktopMacOS,
+      meta: _isDesktopMacOS,
+    ): const ScrollToDocumentBoundaryIntent(forward: false),
+    SingleActivator(
+      LogicalKeyboardKey.end,
+      control: !_isDesktopMacOS,
+      meta: _isDesktopMacOS,
+    ): const ScrollToDocumentBoundaryIntent(forward: true),
```
EchoEllet, line 165: Those two included in #1937. I wasn't and still not sure if this is needed, seems to be redudent though.

lib/src/editor/raw_editor/raw_editor_actions.dart
```diff
@@ -268,6 +268,113 @@ class QuillEditorExtendSelectionOrCaretPositionAction extends ContextAction<
...
+      extentOffset: targetLineBreak,
+    );
+
+    return Actions.invoke(
+      context ?? (throw StateError('BuildContext should not be null.')),
+      UpdateSelectionIntent(
+        state.textEditingValue,
+        newSelection,
+        SelectionChangedCause.keyboard,
+      ),
```
EchoEllet, line 373: Will need to know if this functionality already exist as a function to call, prefer to not repeat it twice, it can be confusing for other developers and for me in the future.

EchoEllet: Related [_linebreak()](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/raw_editor/keyboard_shortcuts/editor_keyboard_shortcut_actions.dart#L56-L80)

AtlasAutocode (commented):
> The support for shortcuts is not clean enough, some of them are overridable:

Just a comment: I have had a brief encounter with _makeOverridable and was surprised by how it works:

Overridable means it can be overridden by **_prior_** implementations. That is exactly the opposite of what I understand by the term.

For example: Widget build Stack:
Widget 1 overridable action
Widget 2 overridable action
Widget 3 uses action from Widget 1! Because the prior implementation takes priority.

In order for the implementation in Widget 2 to correct a defect in Widget 1's implementation, it has to not use _makeOverridable.

I hope I am wrong in this. 

In my case, I seem to recollect that there was an error in how Flutter handled a macOS shortcut key and since Flutter was at the top of the rendering stack it always took priority so my fix would never work if I used _makeOverridable!

EchoEllet:
Our current code is quite confusing, even after the cleanup we have `editor_keyboard_shortcut_actions.dart` and `raw_editor_actions.dart`, `ScrollToDocumentBoundaryIntent` is already registered intent by default so including it in:

```dart
  SingleActivator(
      LogicalKeyboardKey.home,
      control: !_isDesktopMacOS,
      meta: _isDesktopMacOS,
    ): const ScrollToDocumentBoundaryIntent(forward: false),
    SingleActivator(
      LogicalKeyboardKey.end,
      control: !_isDesktopMacOS,
      meta: _isDesktopMacOS,
    ): const ScrollToDocumentBoundaryIntent(forward: true),
```

Doesn't make any difference, I still want to know if this macOS specific and why we don't need to register a activator for it before moving further.

Was added in #1937.

singerdmx (approved):

EchoEllet:
See #2288, while cleaning up, I'm having even more questions and the answers are not clear, I have spent almost one or two days trying to fix a minor bug, the bug has been fixed without fixing the related code quality issues, I prefer to completely rewrite the Keyboard shortcuts with their actions and intent instead and might do in a separate PR or might improve it later.

Hopefully was able to move around 170 lines of code from `QuillRawEditorState`.

@CatHood0 @AtlasAutocode If you can review the PR, once addressing all the changes, will merge this PR for now.

CatHood0:
Sorry I haven't been able to fix a few issues yet. I've had a busy time these past few days. I'll take the time tonight to do some code reviews and fix issues that were already on my list.

Issue #1243, opened 2023-05-30
Is it possible to add request headers for the display of network pictures


zhtes:
Is it possible to add request headers for the display of network pictures?
like `Image.network("url",headers:{})`

singerdmx:
You can add this feature by submitting a PR

PR #2317, opened 2024-10-15
custom HTTP headers for network images

charlesxsh:
## Description
Allow users to pass custom HTTP headers when accessing network images.

## Related Issues

- *Related #1243*

## Type of Change

### DIFF
--- a/flutter_quill_extensions/lib/src/editor/image/image_embed.dart
+++ b/flutter_quill_extensions/lib/src/editor/image/image_embed.dart
@@ -50,6 +50,7 @@ class QuillEditorImageEmbedBuilder extends EmbedBuilder {
       width: width,
       assetsPrefix: QuillSharedExtensionsConfigurations.get(context: context)
           .assetsPrefix,
+      networkImageHeaders: configurations.networkImageHeaders,
     );
 
     final imageSaverService =
--- a/flutter_quill_extensions/lib/src/editor/image/models/image_configurations.dart
+++ b/flutter_quill_extensions/lib/src/editor/image/models/image_configurations.dart
@@ -17,6 +17,7 @@ class QuillEditorImageEmbedConfigurations {
     this.imageProviderBuilder,
     this.imageErrorWidgetBuilder,
     this.onImageClicked,
+    this.networkImageHeaders,
   }) : _onImageRemovedCallback = onImageRemovedCallback;
 
   /// [onImageRemovedCallback] is called when an image is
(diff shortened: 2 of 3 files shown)


charlesxsh:
> Thank you for your contribution.
> 
> You can use `imageProviderBuilder` instead to check if it's a network image and then pass the headers from there.
> 
> This new property will be ignored if users are using `imageProviderBuilder`.
> 
> I prefer not to add too many properties to pass them somewhere else.
I agree that not adding too many properties is the right way. But in this quite common & important usecase, this is impratical.
That is my initial try. And then I noticed that, I have to cover all the other situations in `imageProviderBuilder` (Memory, Asset, File, etc..) if I use it, which is quite inconvenient (All the other handlers/functions for different image are not accessible from outside pacakge).

```
if (imageProviderBuilder != null) {
    return imageProviderBuilder(context, imageSource);
}
// rest of memory, asset, file will be ignored.
```
But what I want is simply add headers to network image.

EchoEllet:
> But in this quite common & important usecase, this is relatively impratical.

Currently, we do have more important issues and should prioritize them over new features.

> That was my initial thinking, using imageProviderBuilder. And then I noticed that, I have to cover all the other situations in imageProviderBuilder (Memory, Asset, File, etc..) if I use it, which is quite inconvenient (All the other handlers/functions for different image are not accessible from outside pacakge).

There is another way around this. We could make the return type of `imageProviderBuilder` nullable, and if it's null, then we will fallback to our default handing. There are other ways as well.

Similar to `customVideoBuilder` (see [10.8.0](https://github.com/singerdmx/flutter-quill/releases/tag/v10.8.0)).

charlesxsh:
If that case, I agree and this PR can be dropped.

EchoEllet:
This issue has been fixed in [release/v11](https://github.com/singerdmx/flutter-quill/blob/release/v11/flutter_quill_extensions/lib/src/editor/image/config/image_config.dart#L76-L93) branch.

See the [related migration change](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#7-the-flutter_quill_extensions) or [image assets within `release/v11` branch](https://github.com/singerdmx/flutter-quill/blob/release/v11/flutter_quill_extensions/README.md#%EF%B8%8F-image-assets).

EchoEllet:
Published a pre-release fixing this issue (#2338):

```yaml
flutter_quill: ^11.0.0-dev.1
flutter_quill_extensions: ^11.0.0-dev.1
```

Issue #2326, opened 2024-10-18
Prevent inserting illegal characters into a link


mbenci:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill version

10.8.4

### Steps to reproduce

Focus the quill editor
Type https://google.com/
Type , or )

### Expected results

The char ","  or ") "should not be part of the link.

### Actual results

The char ","  or ") " is part of the link and the link does not work

orevial:
`,` and `)` (and `*`, `(`, `!`...) are all valid characters as soon as they don't appear in the host part.

To me quill editor works as expected when it comes to the handling of these characters: 

* Type `https://google.com` (without the end slash) and then type `,` or `)` : the last char won't be recognized as part of the link because it wouldn't make for a valid host in the link.
* Type `https://google.com/` (this time with the end slash) and then type `,` or `)` : the last char will be recognized as part of the link because it is a valid path character.

Take this Google Maps position URL as an example : 

```https://www.google.fr/maps/place/Paris/@48.8588475,2.3058358,13z/data=!3m1!4b1!4m6!3m5!1s0x47e66e1f06e2b70f:0x40b82c3688c9460!8m2!3d48.8575475!4d2.3513765!16zL20vMDVxdGo?entry=ttu&g_ep=EgoyMDI0MTAyMi4wIKXMDSoASAFQAw%3D%3D```

As you can see, they are using `@`, `!` or even `,` as part of the path, and IMO we should definitely be able to support that in Quill Editor

mbenci:
Thanks for the reply and the choice is understandable.
However, there remains a problem which in my opinion is a bug:
if I write https://www.google.it, the ',' is part of the (correct) link, 
but if I delete the ',' it continues to be part of the (wrong) link, same thing with other characters like '.' ')' etc

CatHood0:
I'm working on a way to prevent events using the `QuillController` and a new parameter for `QuillEditorConfigurations`.

CatHood0:
The issue persists in the current version of the package?

PR #2328, opened 2024-10-21
fix: allow all correct URLs to be formatted

orevial:
## Description

Quill editor has a rule named `AutoFormatMultipleLinksRule` that auto format HTTP(s) links, however there was a small caveat where path was mandatory if characters others than A-Z were used. 

The problem is that the URL can contain fragments and query params without requiring an extra `/` as per RFC 3986 on URL structure:

```
scheme://host[:port][/path][?query][#fragment]
``` 

Examples of what the rule was handling before the fix: 

- ✅ http://beginner.example.edu/#act => {link: http://beginner.example.edu/#act}
- ❌ http://beginner.example.edu#act => {link: http://beginner.example.edu}
- ✅ https://brass.example.com/?anger=branch => {link: https://brass.example.com/?anger=branch}
- ❌ https://brass.example.com?anger=branch&actor=amusement#adjustment => {link: https://brass.example.com}

Proposed PR allows query params and fragments to be used even without a `/` character beforehand, thus respecting what the RFC is stating.

Also, I removed the non-prefixed URL from the comments because as opposed to what what was said, it didn't work: 

```dart
/// It works for the following testing URLs:
// www.google.com 
```

### DIFF
--- a/lib/src/rules/insert.dart
+++ b/lib/src/rules/insert.dart
@@ -343,7 +343,6 @@ class AutoFormatMultipleLinksRule extends InsertRule {
   /// This pattern is used to match a links within a text segment.
   ///
   /// It works for the following testing URLs:
-  // www.google.com
   // http://google.com
   // https://www.google.com
   // http://beginner.example.edu/#act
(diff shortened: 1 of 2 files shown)


EchoEllet:
LGTM.

Does this PR fix #2326 or any related issues in the repo?

EchoEllet (approved):
LGTM.

Does this PR fix #2326 or any related issues in the repo?

orevial:
> LGTM.
> 
> Does this PR fix #2326 or any related issues in the repo?

Nope, but to me issue #2326 is not an actual problem we want to solve, see [my comment](https://github.com/singerdmx/flutter-quill/issues/2326#issuecomment-2437384951).

Issue #2260, opened 2024-09-21
Breaking changes of version 11.0.0


EchoEllet:
This issue will document the breaking changes that are planned for 11.0.0

The new major version should not be released without discussing it first, introducing all of our breaking changes at once, major versions give us the opportunity to clean a lot of issues so it should be used to clean up as much as we can, versions such as 7.0.0 and 10.0.0 almost didn't introduce much of breaking changes.

7.0.0 only cleaned some minor code that shouldn't be exposed directly and 10.0.0 introduced due to the project structure refactoring which wasn't a breaking change for many users.

This issue will be updated as we plan. All breaking changes should be planned here instead of discussing them in different issues.

CatHood0:
I propose that we should and need to refactor files that contain several classes that are only used internally.

If you want to see what the problem is with this you can take a look at: [editor](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/editor.dart), [text_line](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/widgets/text/text_line.dart), [text_block](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/widgets/text/text_block.dart), [proxy](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/widgets/proxy.dart), etc. There are many files of this style that we can perfectly divide into different files (and thus avoid creating files with more than 1000 lines that are not always easy to debug)

EchoEllet:
**Update**: I have started working on [release/v11](https://github.com/singerdmx/flutter-quill/tree/release/v11), if anything is missing, something that shouldn't be done, improvements or an issue, or suggestions for some breaking changes, I'm open for all suggestions.

This major release aims to address issues requiring breaking changes, along with cleanup and refactoring for better organization. See the issue details.

akshaybabloo:
@EchoEllet Would you be planning for tables in v11?

EchoEllet:
> Would you be planning for tables in v11?

For now, we don't have any plans to, we have deprecated it in #2254 and has been removed in v11 (#2338). We might create a highly experimental package (in a separate repo) with this feature though with our current state **it's not production-ready**.

We're trying to prioritize existing issues before adding more features to provide more on the surface for more issues and we have agreed on that for now.

@CatHood0 

[...]

PR #1876, opened 2024-05-18
Fix block multi-line selection style

AtlasAutocode:
## Description

Resolve problems with style selection for multi-line selections involving block styles.
Improved maintainability by refactoring toolbar buttons to derive from common base class.

## Related Issues

- *Fix #1862*
- *Related #1861, #1863, #1846*

## Improvements
Refactored toolbar buttons (DRY methodology) to improve maintainability and resolve Issue 1862 for multiple use cases.
Class QuillToolbarCommonButtonState holds common access to: controller, options, tooltip, iconSize, iconFactor, iconTheme and afterButtonPressed. Toolbar buttons derive from common base class using generic type definitions.
Enabled focus to be restored to editor after toolbar button action.
## Checklist

- [ X ] I read the [Contributor Guide](../CONTRIBUTING.md) and followed the process outlined there for submitting PRs.
- [ X ] I titled the PR using [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0).
- [ X ] I did not modify the `CHANGELOG.md` nor the plugin version in `pubspec.yaml` files.
- [ X ] All existing and new tests are passing.
- [ X ] I have run the commands in `./scripts/before_push.sh` and it all passed successfully

## Breaking Change

Does your PR require plugin users to manually update their apps to accommodate your change?

- [ ] Yes, this is a breaking change (please indicate that with a `!` in the title as explained in [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0)).
- [ X ] No, this is *not* a breaking change.

PR #2338, opened 2024-10-27
feat!: release version 11.0.0 with breaking changes

### LINKED ISSUE #2211: Improve quality and format of CHANGELOG.md

### LINKED ISSUE #2249: Simplify the example
The example currently has different things that are not related to the project such as using `bloc` and `hydrated_bloc` only to save the state of the settings in the app storage. The settings itself should not be a thing, and the same thing for the home screen, we should be directly in the editor screen and then have the option to open different templates instead of having to choose a default template/sample to open, to save all of us time when writing step to reproduce a bug, avoid confusing the user, better hot reload experience on the web since it will reset the state of the app completely on each change, it should be focused on using the editor itself as it's in a real application, not related but the UI of the toolbar is not exactly the user experience we're looking for, especially on mobile, too many buttons that are not grouped and it take half of the screen, with our current state I don't plan on improving the simple toolbar provided with the package, instead will only provide it in the example app.

Android Example</img>

We should not do anything in the example that we don't recommend the user to do, we already mentioned to not importing the file from the src directly especially if not exported but we're doing it in the example, currently, we're in the middle of other tasks, once done will fix this issue.

The only reason I remade this example in 2023 since the previous example implemented some features such as the image embed builder of the web on the example while mentioned in README it's a supported feature for the web, also some bug fixes were introduced only in the example and is not documented.

Also, we want to simplify the bug report template, we should not accept the code sample input and should be only tested on the example itself to save time for both developer and user.

### LINKED ISSUE #2301: Remove the QuillToolbarProvider, QuillEditorProvider and related classes and members
**A feature allows children of `QuillEditor` and `QuillToolbar` and related widgets to access their configuration and properties so that we have one place to change some things and access them across different places inside those widgets.**

This feature was added initially so that we can access the configuration and properties in the widget tree using the `BuildContext` without passing the property across multiple constructors, unfortunately, this change wasn't introduced with enough planning and I don't see any valid use case, or how it's useful, it's problematic and caused bugs that we had to manually fix in a non-clean way that's not type-safe, sometimes we had to write some minor tests for bug fixes, making it inefficient, instead we should have one data class that hold all related parameters and passing them all around and requiring them if necessary.

I also expect minor memory leak issues from this feature since not being disposed of properly. It doesn't impact the performance much if that's our main concern.

It's not documented properly, poorly written and tested and this was a mistake I made last year, we don't have a strong reason to keep using it, as such, in the long term we will remove it in either backward compatible change or release it as a major release. Luckily it doesn't expose APIs that are difficult to remove or change to the public.

This is one of the things to fix in `11.0.0`, though we should have more planning about how it will be done and why this is an issue in the first place.

Why might want to know why would want this feature? Is customization options to different things all from one place should be a feature in the first place?

We should also discuss whether we want to fix more issues in `11.0.0` or simply introduce it with fewer breaking changes or fixes and then introduce `12.0.0` after a while or have them all in one version. IMO That will depend on the user feedback.

After planning, it will be a while until I start to fix this issue though it could also done soon.

### LINKED ISSUE #2320: The option to disable the Rich clipboard feature or customize the clipboard behavior
With the rich clipboard feature implemented in the project, we don't have a clean way to provide so users can customize the behavior or disable some of the features, and change how they work.

See [related comment](https://github.com/singerdmx/flutter-quill/pull/2230#issuecomment-2386277135).

A robust way to override the behavior of all clipboard operations without changing much or having to fork the project.

- Part of #2230
- Related #2156

### LINKED ISSUE #2321: Update the exposed libraries
- The `flutter_quill_internal.dart` should be renamed to `internal.dart`, so the import will be `import 'package:flutter_quill/internal.dart';` instead of `import 'package:flutter_quill/flutter_quill_internal.dart';`. I introduced it as `flutter_quill_internal` mainly because it's a common convention in some packages, now I think it creates inconsistency issues and makes the import unnecessarily longer. We do have other libraries such as `quill_delta.dart` or `translations.dart`, both don't start with `flutter_quill` and I prefer not to hardcode the project name everywhere where we don't need to. Also, add a clear dart doc [in here](https://github.com/singerdmx/flutter-quill/blob/409d2d1736c78b904f495b2fcebfa27e89cc6131/lib/flutter_quill_internal.dart#L1-L6).
- Might add new library `experimental.dart` for exporting new experimental APIs only, though needs more consideration before introducing it.
- Remove the deprecated library `markdown_quill.dart`. Removing it is a breaking change, so I deprecated it and it has been deprecated for a month.

### LINKED ISSUE #2335: Localization issue with Image picker menu popup

### Flutter Quill version

10.8.4

### Steps to reproduce

1. Create a Flutter toolbar editor with a French Locale:
QuillToolbar.simple(
              controller: _controller,
              configurations: QuillSimpleToolbarConfigurations(
                showCodeBlock: false,
                showSearchButton: false,
                showFontFamily: false,
                embedButtons: FlutterQuillEmbeds.toolbarButtons(),
                sharedConfigurations: QuillSharedConfigurations(
                  locale: const Locale('fr'),
                ),
              ),
            )
2. Start your quill editor with toolbar 

### Expected results

The image and video caption button should be both in French.
The imagePicker menu should be in French.

### Actual results

Only the image button caption is in French.
The imagePicker menu is in English.
Screenshot_20241025_220952
Screenshot_20241025_220907
Screenshot_20241025_220819

### Additional Context

<details>
<summary>Debug analysis</summary>

The context when the class SelectImageSourceDialog is build is in English. There seems to be a disruption between the context given in the toolbar button and the context given to the popup.
builder: (_) => const FlutterQuillLocalizationsWidget(
      child: SelectImageSourceDialog(),
    ),

Not sure exactly why...

</details>

### LINKED ISSUE #2350: Add paste options (bool plainTextOnly and a callback to view/modify a delta prior to pasting)

### Use case

Problem: On iOS when pasting HTML, it is converted into a quill delta object and then performs a controller `replaceText()`.  The conversion assumes the full functionality of the editor and keeps all converted attributes.  If the QuillEditor implementation isn't using some of the Quill features, it can be inserted into the document and not be editable.  i.e. fonts, colors, images, etc.

I propose adding two changes to the controller.

1) Whenever HTML is pasted, pass the converted delta object to a callback that would allow the delta to be edited. The callback would return the edited delta object and that would be used with the replaceText controller method.

2) Provide a bool that would convert any pasted HTML into plain text prior to calling `replaceText()`.

Alternative solutions: Currently in the onReplaceText controller callback, I'm examining the delta, removing any unsupported attributes, and then calling a 2nd replace text with the modified delta.
      

#### `quill_controller_rich_paste.dart`

```dart
final htmlText = await getHTML();
    if (htmlText != null) {
      final htmlBody = html_parser.parse(htmlText).body?.outerHtml;
      // ignore: deprecated_member_use_from_same_package
      var deltaFromClipboard = DeltaX.fromHtml(htmlBody ?? htmlText);
      
      //Add an optional callback...
      if (verifyPasteDelta != null) {
          deltaFromClipboard = verifyPasteDelta!(deltaFromClipboard);
       }
       // end of add

      _pasteUsingDelta(deltaFromClipboard);

      return true;
    }
```

EchoEllet:
## Description

*Prepare version `11.0.0` and introduce breaking changes, and cleanup with no new features.*

To try the new version before it is released:

```yaml
dependency_overrides:
  flutter_quill:
    git:
      url: https://github.com/singerdmx/flutter-quill.git
      ref: release/v11
  flutter_quill_extensions:
    git:
      url: https://github.com/singerdmx/flutter-quill.git
      ref: release/v11
      path: flutter_quill_extensions
```

See the [migration guide](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md) to migrate from `10.0.0` to `11.0.0`.

## Related Issues

- *Related #2260*
- *Fix #2321*
- *Fix #2301*
- *Fix #2320*
- *Fix #2335*
- *Fix #2211*
- *Fix #2249*
- *Partial Fix #2350*
- *Related #2097*

## Type of Change

- [x] ❌ **Breaking change:** Alters existing functionality and requires updates.


EchoEllet:
@CatHood0 @AtlasAutocode @singerdmx 

If you have the time, can you review this PR? See [the migration guide](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md) if you want the short list of changes.

singerdmx (approved):

CatHood0 (approved):
I agree with these changes.

I can't really give a more detailed opinion, my inactivity due to lack of time made me miss many changes that were made, and until now I feel a bit lost among so many changes.

Thank you very much for taking the time to put all these changes together and create this PR.

AtlasAutocode (changes requested):
To evaluate the transition, I attempted to upgrade a serious application using the flutter quill editor from version 10 to version 11.

### Major Breaking Bugs:

#### Quill Toolbar no longer works
After clicking a toolbar button, focus is not restored to the editor window.
Example: Type in some text. Click the Bold toolbar button. Focus is not restored to the editor window. User must click in the editor window to continue typing.
**Reason**: editorFocusNode was removed from QuillController
This is a Major Bug and must be fixed.
Please restore the ability to set the focus to the editor when a toolbar action is invoked.

#### QuillToolbar class no longer exists - need documentation on what to use and what configuration options to use.
You are forcing users to use the QuillSimpleToolbar.
Documentation states it was removed as not used which is not true.
Anyone who used the original class is totally dead - users will expect a suggested replacement.
I used the original class in my app to create a custom toolbar specifically for desktop.
I do not use the QuillSimpleToolbar as it is not suitable for my desktop application.
The loss of the base implementation means I have to resurrect the old flutter quill code and create my own class.
This is a major inconvenience!

#### Conclusion:
I am sure you have considered these breaking changes and have an answer.
I am giving a quick response that mimics users upgrading and they face a major amount of work in this upgrade.

When I get time, I will try to look at other aspects.

EchoEllet:
Thank you all for the review.

Feel free to drop suggestions for further changes.

> After clicking a toolbar button, focus is not restored to the editor window.
Example: Type in some text. Click the Bold toolbar button. Focus is not restored to the editor window. User must click in the editor window to continue typing.
Reason: editorFocusNode was removed from QuillController
This is a Major Bug and must be fixed.

Adding the editor focus request was already a breaking change introduced in a minor version, I removed it since it's possible to implement in the dev project. Adding properties in the controller for the `QuillSimpleToolbar` is not ideal IMO, especially if it's the simple toolbar config.

> User must click in the editor window to continue typing.

We shouldn't set it as the default since the behavior is different on mobile than on desktop. I agree that this should be in the [breaking behavior](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#-breaking-behavior) section and should provide code snippets to revert to the old behavior.

> QuillToolbar class no longer exists - need documentation on what to use and what configuration options to use.
You are forcing users to use the QuillSimpleToolbar.
Documentation states it was removed as not used which is not true.
Anyone who used the original class is totally dead - users will expect a suggested replacement.

I agree that the current docs change is not documented enough and should clarify this, and provide more code snippets. The custom toolbar page has been updated.

I'm trying not to force users to use the `QuillToolbar` to write their toolbar, a minor reason why the provider and `FlutterQuillLocalizationsWidget` were removed.

<details><summary>Introduce alternative for `QuillSimpleToolbar`?</summary>
<p>

> I do not use the QuillSimpleToolbar as it is not suitable for my desktop application.

I wanted to rewrite the toolbar, to provide something that's opinioned and works for most use cases, looks and functions better since now we have too many buttons, they take more than half of the screen on mobile, and are not platform adaptive on desktop. However, that doesn't solve the priorities issue.

</p>
</details> 

> The loss of the base implementation means I have to resurrect the old flutter quill code and create my own class.
This is a major inconvenience!

I'm aware of the how inconvenience and it's unfortunate.

I didn't want to touch the `QuillSimpleToolbar` to keep it backward-compatible, and instead, provide a new toolbar widget, so I didn't have plans to remove this feature.

Restoring it without the toolbar provider will introduce us to a lot of boilerplates, we have to pass them around the constructors and create a method that's the opposite of `copyWith()` which will provide its default value if `null` for all the button options, it was already a buggy feature and I shouldn't introduce it, I didn't prioritize`QuillSimpleToolbar` since on the long-term it's better to replace it. The toolbar is a small part of the library and shouldn't be a priority with our current issues.

Also, it's confusing to document that they need to wrap their toolbar with `QuillToolbar`, providing them the button options param and it only works if they're using the buttons of the simple toolbar.

> I am sure you have considered these breaking changes and have an answer.
I am giving a quick response that mimics users upgrading and they face a major amount of work in this upgrade.

I published this PR as a dev release since I'm still considering things and it might not be final, I appreciate the feedback and I'm looking for more feedback before releasing this as stable.

EchoEllet:
> I am giving a quick response that mimics users upgrading and they face a major amount of work in this upgrade.

What took the most time during the migration? IMO removing the base options will require time the most if the user depends on them and the simple toolbar buttons in their custom toolbar or `QuillSimpleToolbar`.

Some common changes such as the config class can be backward compatible:

```dart
@Deprecated('Use QuillEditorConfig instead.')
typedef QuillEditorConfigurations = QuillEditorConfig;

class QuillEditorConfig {}
```

Deprecating them rather than removing them can help users migrate quickly without reading the migration guide.

I would appreciate more feedback, as I want to understand the user's perspective on a production project.

EchoEllet:
I think there is some confusion understanding the removal of `QuillToolbar` due to [this invalid commit](https://github.com/singerdmx/flutter-quill/commit/5fc6d34d1078539777b6748841f0d022e750442d#diff-0c4f7fedfe20d7730b4a276ad040506420791a192ba5387fb963cdcde0d8b415L10-R9).

In the previous code, it uses `QuillToolbar.simple` which is incorrect for a custom toolbar, instead, you would use `QuillToolbar`. The name is confusing and is not documented properly.

The widget `QuillToolbar` is not a toolbar on its own like the `QuillSimpleToolbar`. It's a widget that only 
ensures to provide the **localization delegate** and the **toolbar provider** and that's no longer required with `11.0.0`.

I have [updated the migration guide](https://github.com/singerdmx/flutter-quill/pull/2338/commits/176e36e6e96507383323a6b9e0ccbe0184007f96#diff-e1c7bc0cca56f09fd900aa3c19d5b9a1c8a0f2a8474af024eea5ca32c30215abR255-R266) once again to clarify this.

AtlasAutocode (changes requested):
I am disappointed by your answers.

Desktop users expect the toolbar to be a part of the editor and the focus to remain within the editor so they can continue to type after clicking a toolbar button.

If you feel that this is incompatible with mobile, then you could have simply made this an option rather than removing the functionality completely.
If you have a code snippet to achieve this, please provide it.

At present I will not be able to use version 11.

EchoEllet:
> Desktop users expect the toolbar to be a part of the editor and the focus to remain within the editor so they can continue to type after clicking a toolbar button.

Introducing this feature across all platforms, including mobile, in a minor version with behavior changes and without clear documentation is problematic. In #1876, testing was done only on the desktop, which led to added properties in the controller that complicate usage and make it hard to revert or fully understand.

Example code snippet:

```dart
  final controller = QuillController.basic();
  final editorFocusNode = FocusNode();
  QuillSimpleToolbar(
    controller: controller,
    config: QuillSimpleToolbarConfig(
      buttonOptions: QuillSimpleToolbarButtonOptions(
        clearFormat: QuillToolbarClearFormatButtonOptions(
          afterButtonPressed: editorFocusNode.requestFocus,
        ),
       // For all buttons
      ),
    ),
  );
  QuillEditor.basic(
    controller: controller,
    config: const QuillEditorConfig(),
    focusNode: editorFocusNode,
  );
```

Adding this to every button is inefficient. So I will try to restore the `base` option since I don't plan on improving the `QuillSimpleToolbar`. Might restore the `QuillSimpleToolbarProvider` and restrict it for internal usage this time.

Thank you for bringing this up.

EchoEllet:
> If you have a code snippet to achieve this, please provide it.

The base button options [have been restored](https://github.com/singerdmx/flutter-quill/pull/2338/commits/1f51935f1eaa229f01c4d14398708ab2d3bd05b0), and it supports the buttons of `flutter_quill_extensions` too:

```dart
  final editorFocusNode = FocusNode();
  QuillSimpleToolbar(
    config: QuillSimpleToolbarConfig(
      buttonOptions: QuillSimpleToolbarButtonOptions(
        base: QuillToolbarClearFormatButtonOptions(
          afterButtonPressed: editorFocusNode.requestFocus,
        ),
      ),
    ),
  );
  QuillEditor.basic(
    focusNode: editorFocusNode,
  );
```

See [this section](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#no-longer-request-editor-focus-by-default-after-pressing-a-quillsimpletoolbars-button) for more details with code snippet using a **custom toolbar**. The migration guide has been updated to include more details with code snippets to revert to the old behavior in the [breaking behavior](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#-breaking-behavior) section.

> then you could have simply made this an option rather than removing the functionality completely.

<details><summary>Why I made this change?</summary>
<p>

**Note: This is not related to the PR directly but to explain a reason for a change.**

**I'm grateful for all of your work over the last few months.**

Removing the `base` option only requires more code to implement and doesn't make it impossible to implement, the main goal of this major release is to clean up some of the changes that were introduced since `7.5.0` and until now. Now that it has been restored, it requires very few changes, the migration guide provides code snippets for changes that break the behavior of the latest stable version (`10.8.5`), you have added this feature recently and I didn't see any reports about the previous behavior.

My main issue is the following:

### The addition of the editor focus node was the `QuillController`.

You also added the [`QuillSimpleToolbarConfig` and `QuillEditorConfig` in the `QuillController`](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/controller/quill_controller.dart#L63-L70), in the [`Document`](https://github.com/singerdmx/flutter-quill/blob/master/lib/src/document/document.dart#L251-L261) too as public API.

We have discussed previously that we don't properly understand the current architecture and we're making changes to the current one which is not enough documented. Having a good structure is currently a priority.

### The default editor focus shouldn't added without a major update and clear docs. 

<details><summary>Reason with details</summary>
<p>

When users run `flutter pub upgrade` all dependencies will be updated to their latest compatible version, and this shouldn't change the current behavior unless there is a strong reason to. With major releases, users will need to manually update the `pubspec.yaml` and read the `CHANGELOG.md` which includes the migration guide. So that they are aware of the breaking changes, especially the ones that change behavior without requiring them to update their existing code. Refer to the [updating package dependencies](https://docs.flutter.dev/packages-and-plugins/using-packages#updating-package-dependencies) page for more details.

</p>
</details> 


### It hasn't been tested on mobile and is more common on desktop than mobile

Depending on the use case and requirements of the developer, it can be unsuitable. They will have to override the default `afterButtonPressed` to achieve this functionality with `() {}` to revert the behavior, as [you stated before](https://www.github.com/singerdmx/flutter-quill/pull/2235#issuecomment-2364588955), features should not be added if one person needs them.


### I haven't done enough code reviews

<details><summary>Details to an old issue</summary>
<p>

I haven't done enough to review PRs recently, merging things and then fixing them later is not ideal either especially since we don't have good code coverage. I have asked questions regarding merged PRs including yours and I couldn't get any answers after a month. I understand you don't have enough time which is why I stopped mentioning you in issue comments and decided to take action but release it as a dev release, and will wait more for reviews. **No need to rush, take your time.**

</p>
</details> 

### Not a commonly requested feature

I do follow all issues on GitHub, Slack, and previously on Telegram and I didn't see anyone requiring this feature before. A feature that's frequently requested doesn’t mean it should be always implemented. For instance, embedding tables and mentions is often requested though we should prioritize features with caution as you [pointed out](https://www.github.com/singerdmx/flutter-quill/pull/2238#pullrequestreview-2312706901).

There are no reasons to make this a default, especially since this minor feature is not commonly requested and it's possible to implement with the minimal code snippet above.

</p>
</details>

EchoEllet:
It's time to review and merge this PR into `master`, but not yet publish it as stable. I want more user feedback and I'm considering changes like disabling rich text paste by default (can be enabled with one line, see #2350) and reintroducing spell check as experimental (removed in v11).

To summary our discussion:

> QuillToolbar class no longer exists

The `QuillToolbar` was used to create a custom toolbar while being able to use the buttons of `flutter_quill`, now it's no longer required and users can use the buttons directly, it has been removed. The [docs has been updated to clarify this](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#%EF%B8%8F-8-removal-of-the-quilltoolbar).

> Quill Toolbar no longer works

It work as before, no changes were introduced, the only minor breaking change that introduced is not requesting the keyboard focus by default, and [that has been already documented before](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#-breaking-behavior), the migration guide has been updated to provide [code snippets to revert on both custom and simple toolbar](https://github.com/singerdmx/flutter-quill/blob/release/v11/doc/migration/10_to_11.md#3-no-longer-request-editor-focus-by-default-after-pressing-a-quillsimpletoolbars-button).

> I used the original class in my app to create a custom toolbar specifically for desktop.
I do not use the QuillSimpleToolbar as it is not suitable for my desktop application.
The loss of the base implementation means I have to resurrect the old flutter quill code and create my own class.
This is a major inconvenience!

[It has been restored](https://github.com/singerdmx/flutter-quill/pull/2338#issuecomment-2449968853) to make the migration smoother and it supports the custom button embed too (integrate with `flutter_quill_extensions`'s buttons).

The pre-release has been published again on pub.dev

singerdmx (approved):

lib/src/editor/config/editor_config.dart
```diff
@@ -458,34 +425,26 @@ class QuillEditorConfigurations extends Equatable {
...
   final void Function(TextInputAction action)? onPerformAction;
 
-  @override
-  List<Object?> get props => [
-        placeholder,
-        // ignore: deprecated_member_use_from_same_package
-        controller?.readOnly,
-      ];
+  // IMPORTANT For project authors: The copyWith()
+  // should be manually updated each time we add or remove a property
```
CatHood0, line 429: Thanks for posting this warning. At the time I really didn't know this and I was racking my brains over why the new parameters I added to the class didn't work.

CatHood0 (approved):

lib/src/editor/config/editor_config.dart
EchoEllet: I was thinking about removing the `copyWith`, using code generators (add dependency of the annotation to production), or adding tests to verify them.

This will ve no longer an issue once [Dart macros](https://dart.dev/language/macros) is complete, and it will require  bumping the min SDK.

EchoEllet:
This PR focuses on breaking changes and prioritizes issues needed in a major release.

Other changes will be addressed separately before publishing the stable release of v11.

Most changes are minor, primarily removals (e.g., `CHANGELOG.md` files, examples). It's not final though I don't expect new bugs.

Two minor things to address:

* If [`QuillEditorConfig.disableClipboard`](https://github.com/singerdmx/flutter-quill/blob/76b55640a659c9074e85fcc3df9b2c9ff90d99a2/lib/src/editor/raw_editor/config/raw_editor_config.dart#L156-L163) should be moved to [QuillClipboardConfig](https://github.com/singerdmx/flutter-quill/blob/7ccd6929ab7c83fb4a55b808130be06f6ca8190c/lib/src/controller/clipboard/quill_clipboard_config.dart#L8) which is used within the controller's config.
* If [`setState` should be always called to update the `_dirty` state](https://github.com/singerdmx/flutter-quill/pull/2338/commits/a127628214c23bb4a7a3b0cdc644fefb21eee738) or only if it's `false`, I reverted this change (made in `8.0.0`) since I don't expect any performance difference and it can cause unexpected bugs.

PR #2345, opened 2024-11-02
Feat: support for customize `onKey` events

### LINKED ISSUE #2308: Flutter Quill Editor Custom Embed: Select Custom Dropdown on Backspace Key

### The question

# **Flutter Quill Editor Custom Embed: Select Dropdown on Backspace Key**

## **Description**
I am working on a Flutter project where I use the **Quill editor** and have created a custom embedded dropdown using the `DropdownEmbed` class. My goal is to detect when the cursor reaches this embedded dropdown and handle the **backspace** key in such a way that instead of deleting the dropdown immediately, it gets selected or highlighted first.

## **Expected Behavior**
- When the user presses the **backspace** key and the cursor reaches the position of the embedded dropdown, the dropdown should be **selected or highlighted**, giving the user a visual cue before it is deleted.

## **Actual Behavior**
- Currently, when the backspace key is pressed and the cursor reaches the embedded dropdown, the dropdown is deleted immediately without selection or highlight.

## **Code Example**
Here is my `DropdownEmbed` class used for embedding a dropdown within the Quill editor:

```dart
const String dropDownType = 'dropdown';

class DropdownEmbed extends Embeddable {
  final DropDownTitleItem title;
  final String options;

  DropdownEmbed({
    required this.title,
    required this.options,
  }) : super(
          dropDownType,
          jsonEncode(
            {
              'title': title.toJson(),
              'options': options,
            },
          ),
        );

  @override
  Map<String, dynamic> toJson() {
    return {
      dropDownType: {
        'title': title.toJson(),
        'options': options,
      },
    };
  }

  static DropdownEmbed fromJson(Map<String, dynamic> json) {
    final data = Map<String, dynamic>.from(json);
    return DropdownEmbed(
      title: DropDownTitleItem.fromJson(data['title']),
      options: data['options'],
    );
  }
}

class DropdownBlockEmbed extends CustomBlockEmbed {
  DropdownBlockEmbed(DropdownEmbed embed) : super(dropDownType, embed.data);

  static DropdownBlockEmbed fromEmbed(DropdownEmbed embed) {
    return DropdownBlockEmbed(embed);
  }

  static DropdownEmbed fromCustomBlockEmbed(CustomBlockEmbed blockEmbed) {
    final data = blockEmbed.data as Map<String, dynamic>;
    return DropdownEmbed.fromJson(data);
  }
}
```

## **Code Example**
Here is my `Quill Editor Code`

```dart
QuillEditor(
  scrollController: scrollController,
  focusNode: focusNode,
  controller: controller,
  configurations: configurations.copyWith(
    customStyles: DefaultStyles(
      h1: DefaultTextBlockStyle(
        defaultTextStyle.style.copyWith(
          fontSize: 32,
          height: 1.15,
          fontWeight: FontWeight.w300,
        ),
        HorizontalSpacing.zero,
        const VerticalSpacing(16, 0),
        VerticalSpacing.zero,
        null,
      ),
    ),
    placeholder: 'Start writing your notes...',
    embedBuilders: [
      DropdownEmbedBuilder(),
      // Other embed builders
    ],
  ),
)
```
## **Question**
- How can I modify the behavior so that when the backspace key is pressed, and the cursor reaches the embedded dropdown, the dropdown gets selected or highlighted instead of being immediately deleted?

CatHood0:
## Description

A new parameter was added with the name `onKeyPressed`, which as its name indicates, calls this function every time a key on the keyboard is pressed.

#### Why was this done?

This is because previously the user never had the opportunity to prevent a key from executing its default function, because we handled all this internally.

It was decided to pass two parameters in this function, the `KeyEvent` and a `Node` that corresponds to the current node that is in selection.

### Example

If we want to prevent all `Embed Objects` from being deleted, we would only have to create a function such that:

```dart
QuillEditor(
  controller: _controller,
     onKeyPressed: (event, node) {
       if (event.logicalKey == LogicalKeyboardKey.backspace && (node is Line || node is Block)) {
          final iterator = DeltaIterator(_controller.document.toDelta())
               ..skip(_controller.selection.baseOffset - 1);
          final cur = iterator.next();
          if (cur.data is! String && cur.data != null) {
             return KeyEventResult.handled;
          }
       }
       return null;
     },
  ),
)
```

## Related Issues

- *Fix #2308*

### DIFF
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -7,6 +7,7 @@ import 'package:flutter/widgets.dart';
 import 'package:meta/meta.dart' show experimental;
 
 import '../../controller/quill_controller.dart';
+import '../../document/nodes/node.dart';
 import '../../editor_toolbar_shared/config/quill_shared_configurations.dart';
 import '../../toolbar/theme/quill_dialog_theme.dart';
 import '../editor_builder.dart';
@@ -60,6 +61,7 @@ class QuillEditorConfigurations extends Equatable {
     this.onSingleLongTapStart,
     this.onSingleLongTapMoveUpdate,
     this.onSingleLongTapEnd,
+    this.onKeyPressed,
     @Deprecated(
         'Use space/char shortcut events instead - enableMarkdownStyleConversion will be removed in future releases.')
     this.enableMarkdownStyleConversion = true,
@@ -152,6 +154,28 @@ class QuillEditorConfigurations extends Equatable {
   ///);
   ///```
   final List<SpaceShortcutEvent> spaceShortcutEvents;
+
+  /// A handler for keys that are pressed when the editor is focused.
+  ///
+  /// # Example:
+  /// assume that you want to avoid of the editor removes any character, then
+  /// you can try this:
+  ///
+  ///```dart
+  ///
+  ///configurations: QuillEditorConfigurations(
+  ///   onKeyPressed: (event, node) {
+  ///     if (event.logicalKey == LogicalKeyboardKey.backspace) {
+  ///         debugPrint('Hello there');
+  ///         return KeyEventResult.handled;
+  ///     }
+  ///     // if you already don't need add more conditions, you can return
+  ///     // null to let to the editor execute the other internal events
+  ///     return null;
+  ///   },
+  ///)
+  ///```
+  final KeyEventResult? Function(KeyEvent event, Node? node)? onKeyPressed;
 
   /// Whether the text can be changed.
   ///
@@ -484,6 +508,7 @@ class QuillEditorConfigurations extends Equatable {
     bool? autoFocus,
     bool? isOnTapOutsideEnabled,
     Function(PointerDownEvent event, FocusNode focusNode)? onTapOutside,
+    KeyEventResult? Function(KeyEvent event, Node? node)? onKeyPressed,
     bool? showCursor,
     bool? paintCursorAboveText,
     bool? enableInteractiveSelection,
@@ -539,6 +564,7 @@ class QuillEditorConfigurations extends Equatable {
       checkBoxReadOnly: checkBoxReadOnly ?? this.checkBoxReadOnly,
       disableClipboard: disableClipboard ?? this.disableClipboard,
       scrollable: scrollable ?? this.scrollable,
+      onKeyPressed: onKeyPressed ?? this.onKeyPressed,
       scrollBottomInset: scrollBottomInset ?? this.scrollBottomInset,
       characterShortcutEvents:
           characterShortcutEvents ?? this.characterShortcutEvents,
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -176,8 +176,8 @@ class QuillEditorConfigurations extends Equatable {
   ///           ..skip(_controller.selection.baseOffset - 1);
   ///     // Get the current [Operation] where the caret is on
   ///     final cur = iterator.next();
-  ///     // Check if the caret is on a Operation that is not text
-  ///     if (cur.data is! String && cur.data != null) {
+  ///     final isOperationWithEmbed = cur.data is! String && cur.data != null;  
+  ///     if (isOperationWithEmbed) {
   ///         // If it is, then ignore this [KeyEvent] to avoid the user
   ///         // removes that [Embed Object]
   ///         return KeyEventResult.handled;
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -61,6 +61,7 @@ class QuillEditorConfigurations extends Equatable {
     this.onSingleLongTapStart,
     this.onSingleLongTapMoveUpdate,
     this.onSingleLongTapEnd,
+    @experimental
     this.onKeyPressed,
     @Deprecated(
         'Use space/char shortcut events instead - enableMarkdownStyleConversion will be removed in future releases.')
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -163,7 +163,7 @@ class QuillEditorConfigurations extends Equatable {
   ///     - Desktop
   ///
   /// # Example:
-  /// assume that we want to avoid the user removes any **Embed Object**, then we can try:
+  /// To prevent the user from removing any **Embed Object**, try:
   ///
   ///```dart
   ///onKeyPressed: (event, node) {
@@ -174,12 +174,11 @@ class QuillEditorConfigurations extends Equatable {
   ///     // and jump directly before the position
   ///     final iterator = DeltaIterator(_controller.document.toDelta())
   ///           ..skip(_controller.selection.baseOffset - 1);
-  ///     // Get the current [Operation] where the caret is on
+  ///     // Get the [Operation] where the caret is on
   ///     final cur = iterator.next();
   ///     final isOperationWithEmbed = cur.data is! String && cur.data != null;  
   ///     if (isOperationWithEmbed) {
-  ///         // If it is, then ignore this [KeyEvent] to avoid the user
-  ///         // removes that [Embed Object]
+  ///         // Ignore this [KeyEvent] to prevent the user from removing the [Embed Object].
   ///         return KeyEventResult.handled;
   ///     }
   ///   }
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -167,11 +167,9 @@ class QuillEditorConfigurations extends Equatable {
   ///
   ///```dart
   ///onKeyPressed: (event, node) {
-  ///   // Check if the node is a line or a block (you can add your custom type checking)
   ///   if (event.logicalKey == LogicalKeyboardKey.backspace &&
   ///       (node is Line || node is Block)) {
-  ///     // We need to use [DeltaIterator] to go to the current position
-  ///     // and jump directly before the position
+  ///     // Use [DeltaIterator] to jump directly to the position before the current.
   ///     final iterator = DeltaIterator(_controller.document.toDelta())
   ///           ..skip(_controller.selection.baseOffset - 1);
   ///     // Get the [Operation] where the caret is on
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -159,7 +159,7 @@ class QuillEditorConfigurations extends Equatable {
   ///
   /// ### Supported by:
   ///
-  ///     - Web
+  ///     - Web (Does not work on mobile devices)
   ///     - Desktop
   ///
   /// # Example:
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -175,6 +175,7 @@ class QuillEditorConfigurations extends Equatable {
   ///   },
   ///)
   ///```
+  @experimental
   final KeyEventResult? Function(KeyEvent event, Node? node)? onKeyPressed;
 
   /// Whether the text can be changed.
--- a/lib/src/editor/config/editor_configurations.dart
+++ b/lib/src/editor/config/editor_configurations.dart
@@ -176,7 +176,7 @@ class QuillEditorConfigurations extends Equatable {
   ///           ..skip(_controller.selection.baseOffset - 1);
   ///     // Get the [Operation] where the caret is on
   ///     final cur = iterator.next();
-  ///     final isOperationWithEmbed = cur.data is! String && cur.data != null;  
+  ///     final isOperationWithEmbed = cur.data is! String && cur.data != null;
   ///     if (isOperationWithEmbed) {
   ///         // Ignore this [KeyEvent] to prevent the user from removing the [Embed Object].
   ///         return KeyEventResult.handled;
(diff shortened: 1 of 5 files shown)


lib/src/editor/raw_editor/keyboard_shortcuts/editor_keyboard_shortcuts.dart
```diff
@@ -74,6 +78,13 @@ class EditorKeyboardShortcuts extends StatelessWidget {
   }
 
   KeyEventResult _onKeyEvent(node, KeyEvent event) {
+    if (onKeyPressed != null) {
+      // we need to found the current node what the user is on
+      final node =
+          controller.document.queryChild(controller.selection.baseOffset).node;
+      final result = onKeyPressed!.call(event, node);
```
EchoEllet, line 85: Let's avoid the `!` operator and instead use a local variable, check it, and then access it, this is the preferred way in Flutter and Dart code ([example code](https://github.com/flutter/packages/blob/674179f97bda7eab1aa5454a0be185ed7312af51/packages/image_picker/image_picker_platform_interface/lib/src/platform_interface/image_picker_platform.dart#L358-L363)).

See the [dart working with nullable fields](https://dart.dev/null-safety/understanding-null-safety#working-with-nullable-fields) for details.

CatHood0: Thanks for the suggestion. I'll change it.

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,29 @@ class QuillEditorConfigurations extends Equatable {
...
+  /// # Example:
+  /// assume that you want to avoid of the editor removes any character, then
+  /// you can try this:
+  ///
+  ///```dart
+  ///
+  ///configurations: QuillEditorConfigurations(
+  ///   onKeyPressed: (event, node) {
+  ///     if (event.logicalKey == LogicalKeyboardKey.backspace) {
+  ///         debugPrint('Hello there');
```
EchoEllet, line 169: What is the use of this print statement? The `debugPrint()` will not be [removed in release mode](https://github.com/flutter/flutter/blob/455be19d928fa8e0e5bb26904f2d0f57312794c6/packages/flutter/lib/src/foundation/print.dart#L37-L43).

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,29 @@ class QuillEditorConfigurations extends Equatable {
...
+  ///         return KeyEventResult.handled;
+  ///     }
+  ///     // if you already don't need add more conditions, you can return
+  ///     // null to let to the editor execute the other internal events
```
EchoEllet, line 173: > let to

I'm not quite sure but shouldn't it `let the editor` instead of `let to the editor`?

EchoEllet, line 173: To simplify it, I suggest:

```dart
// Apply custom logic or return null to use default events
return null;
```

lib/src/editor/config/editor_configurations.dart
CatHood0: That's an example. But i still working to create another usecase that can be useful for devs.

EchoEllet: > That's an example.

I would remove it:

* To simplify it, it doesn't have to be simple, only as simple as possible.
* To avoid confusion, users often copy paste code without understanding some parts, I have added some code in an example, and I have seen it become common and used without a reason.
* It's only used for debugging and confirming if this check was evaluated.

If you will keep it, can you update the message to be more specific and clear?

> But i still working to create another usecase that can be useful for devs.

The other example is to ensure embedded widgets can't be deleted?

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,43 @@ class QuillEditorConfigurations extends Equatable {
   ///```
   final List<SpaceShortcutEvent> spaceShortcutEvents;
 
+  /// A handler for keys that are pressed when the editor is focused.
+  ///
+  /// ### Supported by:
+  ///
+  ///     - Web
+  ///     - Desktop
```
EchoEllet, line 163: Is this feature desktop specific? Is it supported on mobile when running on a browser (web)?

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,43 @@ class QuillEditorConfigurations extends Equatable {
   ///```
   final List<SpaceShortcutEvent> spaceShortcutEvents;
 
+  /// A handler for keys that are pressed when the editor is focused.
+  ///
+  /// ### Supported by:
```
EchoEllet, line 160: Nit: At some point, we need a style for docs to ensure consistency since it's different from docs comment to another. Though not an issue of this PR.

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,43 @@ class QuillEditorConfigurations extends Equatable {
...
+  ///
+  ///     - Web
+  ///     - Desktop
+  ///
+  /// # Example:
+  /// assume that we want to avoid the user removes any **Embed Object**, then we can try:
+  ///
+  ///```dart
+  ///configurations: QuillEditorConfigurations(
```
EchoEllet, line 169: Maybe we can remove the configuration part here since who sees the docs is usually already in the context of the config class.

If not, then we will need to update this to `Config` when merging the `release/v11` branch.

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,43 @@ class QuillEditorConfigurations extends Equatable {
...
+  ///   onKeyPressed: (event, node) {
+  ///     // Check if the node is a line or a block (you can add your custom type checking)
+  ///     if (event.logicalKey == LogicalKeyboardKey.backspace &&
+  ///         (node is Line || node is Block)) {
+  ///       // We need to use [DeltaIterator] to go to the current position
+  ///       // and jump directly before the position
+  ///       final iterator = DeltaIterator(_controller.document.toDelta())
+  ///             ..skip(_controller.selection.baseOffset - 1);
+  ///       // Get the current [Operation] where the caret is on
+  ///       final cur = iterator.next();
```
EchoEllet, line 179: Nit: Similar to the previous comment, we need style guidelines at somepoint, some part of the code would prefer `current`, and others use `current`. Even if it's an example.

lib/src/editor/raw_editor/config/raw_editor_configurations.dart
```diff
@@ -159,6 +137,43 @@ class QuillRawEditorConfigurations extends Equatable {
...
 
+  /// A handler for keys that are pressed when the editor is focused.
+  ///
+  /// ### Supported by:
+  ///
+  ///     - Web
+  ///     - Desktop
+  ///
+  /// # Example:
+  /// assume that we want to avoid the user removes any **Embed Object**, then we can try:
```
EchoEllet, line 148: It's quite inconvenient that we have to document the same parameter in two places, I'm actually unsure about this without knowing the core differences between the `QuillRawEditor` and `QuillEditor`.

lib/src/editor/raw_editor/config/raw_editor_configurations.dart
```diff
@@ -159,6 +137,43 @@ class QuillRawEditorConfigurations extends Equatable {
...
+  ///           // If it is, then ignore this [KeyEvent] to avoid the user
+  ///           // removes that [Embed Object]
+  ///           return KeyEventResult.handled;
+  ///       }
+  ///     }
+  ///     // Apply custom logic or return null to use default events
+  ///     return null;
+  ///   },
+  ///)
+  ///```
```
EchoEllet, line 173: Nit: Maybe we could add `See also` that reference the other config parameter, or maybe adding tests to ensure that's in sync.

So when we update, we won't forget the other. This is not part of this PR.

lib/src/editor/raw_editor/keyboard_shortcuts/editor_keyboard_shortcuts.dart
```diff
@@ -74,6 +78,14 @@ class EditorKeyboardShortcuts extends StatelessWidget {
   }
 
   KeyEventResult _onKeyEvent(node, KeyEvent event) {
+    final onKey = onKeyPressed;
+    if (onKey != null) {
+      // we need to found the current node what the user is on
```
EchoEllet, line 83: How about:

```dart
// Find the current node the user is on.
```

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,43 @@ class QuillEditorConfigurations extends Equatable {
...
+  ///       // Check if the caret is on a Operation that is not text
+  ///       if (cur.data is! String && cur.data != null) {
```
EchoEllet, line 181: Nit: Maybe extractring this check into a local variable with a descriptive name?

EchoEllet (approved):
Thanks for putting this together.

lib/src/editor/config/editor_configurations.dart
CatHood0: Yeah, it's only supported by desktop platform (mobile platform is a different case. We will need to work on this)

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,41 @@ class QuillEditorConfigurations extends Equatable {
...
 
+  /// A handler for keys that are pressed when the editor is focused.
+  ///
+  /// ### Supported by:
+  ///
+  ///     - Web
+  ///     - Desktop
+  ///
+  /// # Example:
+  /// assume that we want to avoid the user removes any **Embed Object**, then we can try:
```
EchoEllet, line 166: Nit: I suggest:

```dart
/// To prevent the user from removing any **Embed Object**, try:
```

lib/src/editor/config/editor_configurations.dart
EchoEllet: What is the reason or the logic that's desktop specific and prevent mobile support?

lib/src/editor/config/editor_configurations.dart
```diff
@@ -153,6 +155,41 @@ class QuillEditorConfigurations extends Equatable {
...
+  ///     // We need to use [DeltaIterator] to go to the current position
+  ///     // and jump directly before the position
+  ///     final iterator = DeltaIterator(_controller.document.toDelta())
+  ///           ..skip(_controller.selection.baseOffset - 1);
+  ///     // Get the current [Operation] where the caret is on
+  ///     final cur = iterator.next();
+  ///     final isOperationWithEmbed = cur.data is! String && cur.data != null;  
+  ///     if (isOperationWithEmbed) {
+  ///         // If it is, then ignore this [KeyEvent] to avoid the user
+  ///         // removes that [Embed Object]
```
EchoEllet, line 182: Nit:

Suggestion:

```dart
/// Ignore this [KeyEvent] to prevent the user from removing the [Embed Object].
```

lib/src/editor/config/editor_configurations.dart
CatHood0: Like #2235 the `onKeyEvent` from `Focus` widget never is called on mobile devices. We will need to create a Widget that makes the same, but, detecting also the key events from virtual keyboards.

EchoEllet: Is the example what is not supported on mobile or the property itself?

CatHood0: In itself, it is the property. Why? This is because on mobile devices, where the soft keyboard is more common, they do not have native support in Flutter to propagate soft keyboard events to the `Focus` widgets, since Flutter does not have a native mechanism to directly intercept or handle these soft keyboard events on all platforms (such as Android and iOS)

EchoEllet: Should we briefly document why this isn't supported on mobile?

CatHood0: Probably. This is related also with why the `CharacterShortcutEvent` and `SpaceShortcutEvent` cannot be executed using soft keyboard.

lib/src/editor/raw_editor/config/raw_editor_configurations.dart
```diff
@@ -159,6 +138,40 @@ class QuillRawEditorConfigurations extends Equatable {
...
+  /// # Example:
+  /// To prevent the user from removing any **Embed Object**, try:
+  ///
+  ///```dart
+  ///onKeyPressed: (event, node) {
+  ///   // Check if the node is a line or a block (you can add your custom type checking)
+  ///   if (event.logicalKey == LogicalKeyboardKey.backspace &&
+  ///       (node is Line || node is Block)) {
+  ///     // We need to use [DeltaIterator] to go to the current position
+  ///     // and jump directly before the position
```
EchoEllet, line 157: Nit:

How about:

```dart
// Use [DeltaIterator] to jump directly to the position before the current.
```

Or:

```dart
// Use [DeltaIterator] to navigate to the current position, then jump directly before the position.
```

lib/src/editor/raw_editor/config/raw_editor_configurations.dart
```diff
@@ -122,6 +100,7 @@ class QuillRawEditorConfigurations extends Equatable {
   ///
   ///    - Web
```
EchoEllet, line 101: Nit: This is unrelated though do we need to make it clear that it doesn't work on mobile devices when running on the browser (web) to ensure consistency?

CatHood0: When can this later to make more clear for devs.

EchoEllet:
I will solve the conflicts of this PR and the others soon.

PR #2385, opened 2024-11-16
Update example to support image pasting on web

Maksim-Nikolaev:
## Description

Update the example main.dart to demonstrate the support of image pasting on web by using in-built dart:html

https://github.com/user-attachments/assets/acc0d8b6-755b-4697-a9d6-1bc7fe7a8098

```dart
  // Create a Blob from the image bytes
  final blob = html.Blob([imageBytes], 'image/png');

  // Create a temporary URL for the blob
  final url = html.Url.createObjectUrlFromBlob(blob);

  return url;
```

## Type of Change

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -6,9 +6,9 @@ import 'package:flutter/material.dart';
 import 'package:flutter_quill/flutter_quill.dart';
 import 'package:flutter_localizations/flutter_localizations.dart';
 import 'package:flutter_quill_example/quill_delta_sample.dart';
+import 'package:flutter_quill_example/stabs/web_utils.dart';
 import 'package:flutter_quill_extensions/flutter_quill_extensions.dart';
 import 'package:path/path.dart' as path;
-import 'dart:html' as html;
 
 void main() => runApp(const MainApp());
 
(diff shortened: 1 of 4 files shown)


Maksim-Nikolaev:
I went back on the phone and realized that dart:html has to be imported conditionally which adds code complexity and requires some decisions, so it's not good for updating example.

For my project I will use universal_html to get around possible issues

EchoEllet:
> dart:html has to be imported conditionally

How about [`package:web`](http://pub.dev/packages/web) instead?

Maksim-Nikolaev:
My main goal was to avoid usage of external dependencies, to make the example as light-weight as possible.

I reopened the PR to show how the conditional imports would work. I checked Android & Web (Chrome) and both are building now while supporting the image pasting.

Feel free to close it if needed.

EchoEllet:
> My main goal was to avoid usage of external dependencies, to make the example as light-weight as possible.

FYI: The `web` package is Dart first party dependency. They didn't make it part of the SDK to keep it more open to contributons, and clients can use newer versions without updating their Dart SDK. The bundle size is the same.

Though I agree this is out of scope for this example, it might be a good idea to move it into an example somewhere else.

Issue #2290, opened 2024-09-27
Replace or remove flutter_keyboard_visibility plugin from flutter_quill


EchoEllet:
It seems that `flutter_keyboard_visibility` hasn't been updated for a while, having PRs on hold for one or two years that already fixed some issues:

- **Support for [Flutter/WASM](https://docs.flutter.dev/platform-integration/web/wasm)**: it seems that this plugin doesn't support web platform for mobile, instead it returns `false` and uses `dart:html` which is being replaced by [web](https://pub.dev/packages/web) package, this is a critical issue for users to migrate their projects to new technologies, while we don't use Flutter/WASM, we don't want to have any blockers or obstacles for the users of this project, a minimal [PR #164]. Also, see [PR #155](https://github.com/MisterJimson/flutter_keyboard_visibility/pull/155) which adds experimental support for the web.
(https://www.github.com/MisterJimson/flutter_keyboard_visibility/pull/164) is already made
- **Always using `31` for `compileSdkVersion` on Android**: Due to an internal bug in Flutter (see [Flutter #63533](https://www.github.com/flutter/flutter/issues/63533)), plugins that hardcode the `compileSdkVersion` will force the app project to not being able to target newer versions of `compileSdkVersion`, the latest is `34`, while this may sound not important for some users, Google play may require targeting newer versions of Android for publishing the app. Better solve this issue sooner than later. A minimal [PR #159](https://github.com/MisterJimson/flutter_keyboard_visibility/pull/159) is already sent.
- **Not supporting recent Android API**: This may sound not an issue at all for most users, however usually native Android apps do use newer APIs when available, and fall to the older API (even if not deprecated) for older versions, see [#120](https://github.com/MisterJimson/flutter_keyboard_visibility/pull/120).
- **Not being updated for 2 years**: It seems only minor changes has been introduced for the last 2 years, in general, we prefer something that's more stable and published less often, however, it seems that new issues and PRs are not being replied on, which might cause issues for the users.

We do want to remove more dependencies, especially heavy or outdated plugins, allowing the users to use `flutter_quill` as a dependency without conflicting or depending on other dependencies so they can choose the dependencies they want to have.

Removing a feature to fix a bug is not the solution will be used here, instead we will find another alternative, if not, we will write our own solution in `quill_native_bridge` or completely have a separate plugin for this functionality, this feature is usually only required on mobile platforms, which make it less time expensive to implement.

Issue #2276, opened 2024-09-23
Rewrite flutter_quill_extensions, separate version of each package


EchoEllet:
**This change will not be done anytime soon** but will be submitted to keep track of our plans and have related discussions in one place.

I propose that we stop supporting `flutter_quill_extensions` at some point (**currently it's not a priority**) to separate it into different packages instead of having one package that's full of dependencies:

- `quill_image_embed`: Provide support for displaying images
- `quill_video_embed`: Provide support for displaying videos
- More embeds

Each one will be separated and more extensible, tested, and documented since we're getting many questions about how to save images, also fix most of the common bugs.

Each embed package has `editor.dart` and `toolbar.dart` files and users can import what they need, support for YouTube will be discounted, instead provide a way to handle all kind of services and provide a solution that requires users to have dependencies on their own project to implement for more flexibility.

When we want to introduce a feature that's full of bugs and is not ready, we can have a separate package with its own pre-release, when a breaking change will be made, it will be made some of them instead of all affecting more users, having fewer dependencies mean fewer issues when new features come out and old packages don't support them yet and causing complication issue, such as support for [Flutter/WASM](https://docs.flutter.dev/platform-integration/web/wasm), smaller bundle size, fitting more requirements of users

Our current project structure can't handle many packages:

- `quill_native_bridge` and related packages will be moved into a separate repo
- The version of the new embed packages will be separated from `flutter_quill`.
- `flutter_quill_test` and `dart_quill_delta` (#2259) will be also separated, `dart_quill_delta` is being adopted by more projects even editors that is not related to `flutter_quill` so breaking changes should be separated for better backward compatibility, as for `flutter_quill_test` we haven't made a single change since first release, releasing changes will be confusing and unorganized. I suggest having `delta_quill` (new published version of `dart_quill_delta`) and `quill_test`. See #2259 for more details about `delta_quill`, as for `quill_test`, it provides limited functionality for now, it depends on `flutter_test` and we wanted to not introduce non-dev dependencies for tests similarly to most popular packages.

I suggest that we use [melos](https://pub.dev/packages/melos) to manage all packages instead of our manual solution (#2262) and current workflow.

We still need to improve the quality of `CHANGELOG.md` since it's important for developers. See #2211.

**This issue will likely take a long time to fix as we don't have plans to work on it anytime soon.**

EchoEllet:
> YouTube will be discounted

[...]

Issue #1529, opened 2023-11-14
How to save the new inserted image in custom directory


jasamin:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Steps to reproduce

I ran the sample code, after selecting a photo from the gallery and returning a new string in the method '`_onImagePickCallback`', However, the URL of the image in '`controller.document`' is still with 'file.path', May I ask how the current version should be replaced in document ?

### Expected results

[{"insert":{"Change to costom url string"}},{"insert":"\n"}]

### Actual results

[{"insert":{"image":"/data/user/0/com.example.example/cache/image_cropper_1699932556689.jpg"}},{"insert":"\n"}]

### Code sample

```dart
embedButtons: FlutterQuillEmbeds.toolbarButtons(

            formulaButtonOptions: const QuillToolbarFormulaButtonOptions(),

            cameraButtonOptions: const QuillToolbarCameraButtonOptions(),

            imageButtonOptions: QuillToolbarImageButtonOptions(

              imageButtonConfigurations: QuillToolbarImageConfigurations(
                onImageInsertedCallback: (image) async {
                  _onImagePickCallback(File(image));
                },

                onImageInsertCallback: onImageInsert,
              ),
            ),
          )
```

```dart
// Renders the image picked by imagePicker from local file storage
  // You can also upload the picked image to any server (eg : AWS s3
  // or Firebase) and then return the uploaded image URL.

  Future<String> _onImagePickCallback(File file) async {
    return 'Change to costom url string';
  }
```

### Screenshots or Video

<details>
<summary>Screenshots / Video demonstration</summary>

[Upload media here]

</details>

### Logs

<details><summary>Logs</summary>

```console
[Paste your logs here]
```

</details>

freshtechtips:
Hi, thank you for the report, I will check it out in a few moments

freshtechtips:
I see what you mean but this is just an example, image cropper is not used by default in `flutter_quill_extensions`
And the example is a little bit outdated, thank you for mentioning that I will update the `example`

The new way of inserting an image is by using `onImageInsertCallback` which is a code that should be executed to insert the image

The default is:

```dart
OnImageInsertCallback defaultOnImageInsertCallback() {
  return (imageUrl, controller) async {
    controller
      ..skipRequestKeyboard = true
      ..insertImageBlock(imageSource: imageUrl);
  };
}

```

[...]

Issue #1200, opened 2023-05-05
[Web / Mobile / Desktop] Weird behavior about image in flutter_quill_extensions.


Cierra-Runis:
My issue is about [Web / Mobile / Desktop]

# My question

Due to there are some strange issues with the built-in image embed in the the flutter_quill_extensions, so I have decided to write my own image embed in my product.

Here are some strange behaviors that I roughly remember:

1. When there is no text before or after the image in the same line of text, some small images will be filled with pure padding to expand the entire screen width.
<details><summary>Screenshot here</summary>

![image](https://user-images.githubusercontent.com/29329988/236413489-62335b91-7fe9-4970-a36f-e41eb3411de8.png)

</details>

2. Regardless of whether there is padding in behavior 1, once there is no text before or after the image in the same text line, if you adjust the width and height of the image, the width will not change, but the height of the padding will change.
3. When you click on the delete button of an image, it may throw an error (I don't quite remember, but it should be like this... I am too lazy to reproduce the error...).
4. After inserting an image, the cursor following it will be blocked and the user will not be able to see it.
5. After the user clears the application's cache, the image will disappear (I don't quite remember too).

# My solution

<details><summary>Code here</summary>

`Embeddable` and `EmbedBuilder`

```dart

import 'package:mercurius/index.dart';

class DiaryImageBlockEmbed extends Embeddable {
  const DiaryImageBlockEmbed(
    String value,
  ) : super(mercuriusImageType, value);

  static const String mercuriusImageType = 'mercuriusImage';

  static DiaryImageBlockEmbed fromDocument(Document document) =>
      DiaryImageBlockEmbed(jsonEncode(document.toDelta().toJson()));

  Document get document => Document.fromJson(jsonDecode(data));
}

class DiaryImageEmbedBuilderWidget extends EmbedBuilder {
  @override
  String get key => 'mercuriusImage';

  @override
  Widget build(
    BuildContext context,
    QuillController controller,
    Embed node,
    bool readOnly,
    bool inline,
  ) {
    Widget getInkWellChild(File file) {
      if (file.existsSync()) {
        return ClipRRect(
          borderRadius: BorderRadius.circular(16.0),
          child: Image.file(file, alignment: Alignment.center),
        );
      } else {
        return MercuriusModifiedFadeShimmerWidget(
          baseColor: Theme.of(context).colorScheme.outline.withOpacity(0.1),
          highlightColor:
[...]

PR #2403, opened 2024-12-02
feat(flutter_quill_extensions): improve image save

### LINKED ISSUE #1759: Saving image file should say where it has been saved on desktop

Follow:
1. Add image
2. Click on the added image, image menu popup will appear
3. From the image menu click on save 
4. A result snackbar will appear

On the result snackbar it should say where the image file has been saved, otherwise on different platforms it hard to find where it. The result snackbar should say something like: `"The image has been saved at: path/to/location"`

The current message is `"Saved using the network"`. 

I was using flutter quill on linux, after saving the image file I have searched it on all folders to find it. After some effort I found it was [`gal`](https://pub.dev/packages/gal) or [`gal_linux`](https://pub.dev/packages/gal_linux) is saving it on `/tmp/gal/image.png'.

### LINKED ISSUE #2394: Support for Swift Package Manager
Flutter now supports [Swift Package Manager](https://docs.flutter.dev/packages-and-plugins/swift-package-manager), a [CocoaPods](https://cocoapods.org/) alternative for managing iOS and macOS native dependencies that come built-in with [Xcode](https://developer.apple.com/xcode/).

* **Integration with Xcode**: With new versions of Xcode that break the CocoaPods compatibility, this is not an issue with SPM—fewer files to manage (no need to `Podfile`, `Podfile.lock`).
* **Simplifies Flutter installation**: No need to install Ruby and Cocoapods or sometimes [Homebrew](https://brew.sh/) and [RVM](https://rvm.io/).

CocoaPods should still be supported in case SPM is unsupported or disabled, currently, it's only available with the `main` branch of Flutter. Flutter plugins have already migrated and support SPM including other community plugins. SPM support is subject to change for app developers but not plugin authors.

The `flutter_quill` itself is not a plugin but it uses plugins that don't support SPM:

* ~~[`quill_native_bridge`](https://pub.dev/packages/quill_native_bridge)~~
* ~~[`flutter_keyboard_visibility_temp_fork`](https://pub.dev/packages/flutter_keyboard_visibility_temp_fork)~~
* ~~[`gal`](https://pub.dev/packages/gal) (a dependency of [`flutter_quill_extensions`](https://pub.dev/packages/flutter_quill_extensions))~~ Replaced by [quill_native_bridge](https://pub.dev/packages/quill_native_bridge) in https://github.com/singerdmx/flutter-quill/pull/2403

EchoEllet:
## Description

**Fix broken image save functionality in [`flutter_quill_extensions`](https://pub.dev/packages/flutter_quill_extensions), improve the default behavior, support web and desktop platforms (platforms that don't have gallery app), add unit tests, support Swift package manager, remove extra dependencies.**

The current behavior of image save functionality is broken on non-mobile platforms, and the way that the extensions package handles the errors and results (error and success messages) is confusing. The goal of this package is to implement an opinioned solution for loading images and videos but it seems most users avoid this package, I plan on replacing it with new packages (`quill_image` and `quill_video`, see #2276), and that will be later.

This PR addresses:

- Added web and Linux support. The [`gal_linux`](https://pub.dev/packages/gal_linux) implementation was confusing due to the way it was handled. It's not officially supported and was implemented by myself. There is no way to implement it correctly without some changes to the `gal` platform interface.
- Added support for Swift package manager for iOS and macOS (#2394).
- Updated desktop image saving behavior. Saving to the gallery isn't common and can be confusing. The desktop should open a system save dialog instead. [`quill_native_bridge`](https://pub.dev/packages/quill_native_bridge) implements both `saveImageToGallery` and `saveImage`.
- Replaced [`gal`](https://pub.dev/packages/gal) and `gal_linux` with `quill_native_bridge`. It supports:
  * [`saveImageToGallery`](https://pub.dev/documentation/quill_native_bridge/latest/quill_native_bridge/QuillNativeBridge/saveImageToGallery.html): Saves to system gallery (Android, iOS, macOS). Not supported on web/Linux; Windows support is possible but not useful; It's not common behavior for Windows users and saving images using the system file save dialog will integrate with the Windows photos app in both cases.
  * [`saveImage`](https://pub.dev/documentation/quill_native_bridge/latest/quill_native_bridge/QuillNativeBridge/saveImage.html): Prompts for a save location on the desktop and downloads image on the web.
  * [`openGalleryApp`](https://pub.dev/documentation/quill_native_bridge/latest/quill_native_bridge/QuillNativeBridge/openGalleryApp.html): Opens the system gallery app. Not supported on web/Linux.
- Improved success/error messages for all platforms; fixed broken open functionality on the desktop.
- Added unit tests to improve code coverage.
- Removed the following dependencies from `flutter_quill_extensions`:
  * ~~[path](https://pub.dev/packages/path)~~ Added back since it's needed once again and there is no reason to re-implement it.
  * [http](https://pub.dev/packages/http): Not needed anymore (legacy code).
  * [cross_file](https://pub.dev/packages/cross_file) Not needed anymore (legacy code).
- Improved mocking support of [QuillNativeBridge](https://pub.dev/packages/quill_native_bridge) (workaround is not needed anymore) for unit tests by converting instance methods to static methods (similar to `ImagePicker` from [image_picker](https://pub.dev/packages/image_picker) and other Flutter plugins).

Added `prefersGallerySave` to `ImageOptionsMenu` which determines if it should prefer saving the image to the system gallery (when supported) instead of using the system file save dialog, currently only macOS is applicable since that's the only platform that supports both `saveImage` and `saveImageToGallery`. If Windows or Linux were supported, they would be applicable too.

Part of https://github.com/FlutterQuill/quill-native-bridge/pull/9. Tests were added there too though native unit tests are still missing (will be fixed soon).

Soon, will replace [flutter_keyboard_visibility_temp_fork](https://pub.dev/packages/flutter_keyboard_visibility_temp_fork) (related #2290).

## Related Issues

- *Fix #2394*
- *Related #1200*
- *Fix #1759* (was not fixed previously)
- *Related #1529*

## Demostration

This feature is platform-specific, expand each section for more details with videos:

<details><summary>Android</summary>
<p>

Saving using `saveImageToGallery` (backed by Android [`MediaStore`](https://developer.android.com/reference/android/provider/MediaStore)) on Android API 29 and later:

https://github.com/user-attachments/assets/2d91015d-ebfa-449a-a4f7-1a99b5ac5022

The system gallery handles name conflicts.

Saving using `saveImageToGallery` (backed by [`java.io.File`](https://docs.oracle.com/javase/8/docs/api/java/io/File.html) and [`MediaScannerConnection`](https://developer.android.com/reference/android/media/MediaScannerConnection)) on Android API 28 and earlier (requires to write to external storage runtime permission):

https://github.com/user-attachments/assets/b34399ea-acde-42a8-8285-00c2bcb77671

On Android API 28 and earlier, the [quill_native_bridge](https://pub.dev/packages/quill_native_bridge) creates a unique image file name to avoid platform errors.

</p>
</details> 

<details><summary>iOS</summary>
<p>

Saving using `saveImageToGallery` (backed by iOS [`PHPhotoLibrary`](https://developer.apple.com/documentation/photokit/phphotolibrary), requires runtime permission):

https://github.com/user-attachments/assets/df26e34a-3044-470c-a561-fabc8069f903

The system gallery handles name conflicts.

</p>
</details> 

<details><summary>macOS</summary>
<p>

Saving using `saveImage` (backed by macOS [`NSSavePanel`](https://developer.apple.com/documentation/appkit/nssavepanel)):

https://github.com/user-attachments/assets/8543f1d0-cf72-4d43-b399-71193046f9c3

The system save dialog confirms overwrite if the file already exists.

Saving using `saveImageToGallery` (backed by macOS [`PHPhotoLibrary`](https://developer.apple.com/documentation/photokit/phphotolibrary), requires runtime permission):

https://github.com/user-attachments/assets/9169798f-ad4c-442c-b0cc-f8a73ade84e8

The system gallery handles name conflicts.

> **Note**
> 
> During development only, permission is always denied when running the app using sources other than Xcode or macOS terminal (such as Android Studio or VS Code). This is a known issue and the `quill_native_bridge` log to the debug console in debug-builds. See https://github.com/flutter/flutter/issues/134191#issuecomment-2506248266

</p>
</details> 

<details><summary>Web</summary>
<p>

Saving using `saveImage`:

https://github.com/user-attachments/assets/7e97892f-55d1-4a5a-b2bc-4057c674381b

It creates [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) object URL, creates `a` element, sets the `href` and `download` attributes, clicks the `a` element programmatically to download the image file, and then revokes the blob object URL.

The browser handles name conflicts.
Whether the browser will ask the user where they want to save the image will depend on the user's preferences in the settings.

</p>
</details> 

<details><summary>Linux</summary>
<p>

Saving using `saveImage` (backed by [`file_selector_linux`](https://pub.dev/packages/file_selector_linux)):

https://github.com/user-attachments/assets/89c7a7c1-b461-4c00-b0a8-ab10bec146b2

Depending on the Linux desktop environment, the system save dialog might not confirm name conflicts.

</p>
</details> 

<details><summary>Windows</summary>
<p>

Saving using `saveImage` (backed by [`file_selector_windows`](https://pub.dev/packages/file_selector_windows)):

https://github.com/user-attachments/assets/75b93e54-2341-4e27-ac90-f1909bddf849

The system save dialog confirms overwrite if the file already exists.

</p>
</details> 

## Messages

The messages have been improved and changed in an optioned way so most users don't need to override the default. It no longer shows `Saved using the network` or `Saved using the local storage`.

* **The unknown error message**: `An unexpected error occurred while saving the image. Please try again.`

* **The permission denied message**: `Couldn’t save the image due to missing permission`

* **The success snack bar message will depend on how the image is saved (platform-specific)**:

  * **Web**: Always `Image downloaded successfully.` with no action button.
  * **Mobile or macOS** (in case of saving to the gallery): `Image saved to your gallery.` with action labeled `Open 
   Gallery`.
   * **Desktop**: `Image saved successfully.` with action labeled `Open File Location` on Windows and Linux or `Open File` 
   on macOS.

## Type of Change

- [x] 🧪 **Tests:** New or modified tests


flutter_quill_extensions/lib/src/common/utils/file_path_utils.dart
```diff
@@ -0,0 +1,44 @@
+String? extractFileNameFromFilePath(String path) {
+  if (path.isEmpty) {
+    return null;
+  }
+
+  final withoutTrailings = !path.contains('/');
```
EchoEllet, line 6: I missed that, this wouldn't work on Windows.

Either we use `dart:io` (will give us a score on pub.dev, false positive indicating WASM is not supported, `dart:io` is not supported on the web with JS target either) or write a custom implementation.

EchoEllet: Will also need to update the tests so they don't fail on Windows.

flutter_quill_extensions/lib/src/editor/image/image_menu.dart
```diff
@@ -126,48 +141,69 @@ class ImageOptionsMenu extends StatelessWidget {
...
+                  return;
                 }
 
                 messenger.showSnackBar(
                   SnackBar(
-                    content: Text(message),
+                    content: Text(localizations.successImageSaved),
+                    action: SnackBarAction(
+                      label: localizations.openFileLocation,
+                      onPressed: () => launchUrl(Uri.file(imageFilePath)),
```
EchoEllet, line 195: This will open the file itself, not the directory where the file is stored. Mainly because on macOS we need an entitlement in sandboxed apps (Flutter macOS apps are sandboxed by default). Also, we will need a custom implementation to get the directory of that file or use [path](https://pub.dev/packages/path).

Importing `dart:io` always indicates the package doesn't support Flutter/WASM on pub.dev (see the [scores](https://pub.dev/packages/flutter_quill/score)) even it's not used on the web. I find this confusing, `dart:io` is not supported on the web either, why specifically Flutter/WASM, and why does it warn usage of some files only instead of all, even if `kIsWeb` is used.

pubspec.yaml
```diff
@@ -52,7 +52,7 @@ dependencies:
   # Plugins
   url_launcher: ^6.2.1
   flutter_keyboard_visibility_temp_fork: ^0.1.1
-  quill_native_bridge: ^10.7.11
+  quill_native_bridge: ^11.0.0
```
EchoEllet, line 55: I have found an issue in [quill_native_bridge_android](https://pub.dev/packages/quill_native_bridge_android), it appears on **Android API 28 and earlier versions only** when saving the image and then deny, it will crash the app with `java.util.ConcurrentModificationException`:

```console
Exception "E/AndroidRuntime(26687): 	at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:49)
E/AndroidRuntime(26687): 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
E/AndroidRuntime(26687): 	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
E/AndroidRuntime(26687): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1808)
E/AndroidRuntime(26687): 	at android.os.Handler.dispatchMessage(Handler.java:106)
E/AndroidRuntime(26687): 	at android.os.Looper.loop(Looper.java:193)
E/AndroidRuntime(26687): 	at android.app.ActivityThread.main(ActivityThread.java:6669)
E/AndroidRuntime(26687): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(26687): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
E/AndroidRuntime(26687): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
E/AndroidRuntime(26687): Caused by: java.util.ConcurrentModificationException
E/AndroidRuntime(26687): 	at java.util.HashMap$HashIterator.nextNode(HashMap.java:1441)
E/AndroidRuntime(26687): 	at java.util.HashMap$KeyIterator.next(HashMap.java:1465)
E/AndroidRuntime(26687): 	at io.flutter.embedding.engine.FlutterEngineConnectionRegistry$FlutterEngineActivityPluginBinding.onRequestPermissionsResult(FlutterEngineConnectionRegistry.java:744)
E/AndroidRuntime(26687): 	at io.flutter.embedding.engine.FlutterEngineConnectionRegistry.onRequestPermissionsResult(FlutterEngineConnectionRegistry.java:405)
E/AndroidRuntime(26687): 	at io.flutter.embedding.android.FlutterActivityAndFragmentDelegate.onRequestPermissionsResult(FlutterActivityAndFragmentDelegate.java:912)
E/AndroidRuntime(26687): 	at io.flutter.embedding.android.FlutterActivity.onRequestPermissionsResult(FlutterActivity.java:967)
```

This issue appears only in the Flutter Quill example, but not within the [quill_native_bridge example](https://pub.dev/packages/quill_native_bridge/example) where it throws `PERMISSION_DENIED` as expected instead of crashing the app.

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
```diff
@@ -0,0 +1,221 @@
...
+          // For more details, see https://github.com/flutter/flutter/issues/134191#issuecomment-2506248266
+          // and https://pub.dev/packages/quill_native_bridge#-saving-images-to-the-gallery
+
+          final possiblePermissionIssueDuringDevelopmentOnMacOS =
+              kDebugMode && defaultTargetPlatform == TargetPlatform.macOS;
+          if (possiblePermissionIssueDuringDevelopmentOnMacOS) {
+            rethrow;
+          }
+
+          return null;
```
EchoEllet, line 195: With this current code, denying the permission will always result in the message:

```
An unexpected error occurred while saving the image. Please try again.
```

Runtime permission is required on Android 9 (API 28) and earlier versions, iOS, and macOS (for macOS, only when saving the file to the gallery). Will need a message.

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
```diff
@@ -0,0 +1,221 @@
...
+}
+
+ImageSaver _imageSaver = ImageSaver();
+
+ImageSaver get imageSaver => _imageSaver;
+
+@visibleForTesting
+set imageSaver(ImageSaver value) => _imageSaver = value;
+
+class ImageSaver {
```
EchoEllet, line 117: Might move this class into a separate class (update the tests too).

pubspec.yaml
EchoEllet: Fixed though I'm not sure why I couldn't reproduce with the `quill_native_bridge` example. This issue doesn't affect production applications since changes are not merged yet. https://github.com/singerdmx/flutter-quill/pull/2403/commits/ef4968d8ab487fde4ac547bd17cee4e89f0dd573

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
EchoEllet: Fixed https://github.com/singerdmx/flutter-quill/pull/2403/commits/230d8f9163f7f6a35ed131565c7f4d760c5654b5.

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
```diff
@@ -0,0 +1,242 @@
...
+    required bool prefersGallerySave,
+  }) async {
+    assert(() {
+      if (imageUrl.isEmpty) {
+        throw ArgumentError.value(imageUrl, 'imageUrl', 'cannot be empty');
+      }
+      return true;
+    }());
+
+    final imageFileName = extractFileNameFromUrl(imageUrl);
```
EchoEllet, line 152: This is a bug introduced during refactoring and doesn't support file path URLs. I suspected a regression since `extractFileNameFromFilePath` is only used in tests.

flutter_quill_extensions/lib/src/editor/image/image_menu.dart
EchoEllet: > * Custom implementation (require unit tests on all platforms too).

I end up with:

```dart
group('pathSeparator', () {
    test('returns \\ on Windows', () {
      debugDefaultTargetPlatformOverride = TargetPlatform.windows;
      expect(pathSeparator, '\\');
      debugDefaultTargetPlatformOverride = null;
    });

    for (final posixDesktopPlatform in {
      TargetPlatform.macOS,
      TargetPlatform.linux
    }) {
      test('returns / on ${posixDesktopPlatform.name}', () {
        debugDefaultTargetPlatformOverride = posixDesktopPlatform;
        expect(pathSeparator, '/');
        debugDefaultTargetPlatformOverride = null;
      });
    }
  });
```

There is no reason to reinvent the wheel, will add the `path` package back.

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
EchoEllet: Fixed (https://github.com/singerdmx/flutter-quill/pull/2403/commits/7ed86d80288237e80613e9b9210ea997e93c96c1).

flutter_quill_extensions/lib/src/common/utils/file_path_utils.dart
EchoEllet: Fixed (https://github.com/singerdmx/flutter-quill/pull/2403/commits/7ed86d80288237e80613e9b9210ea997e93c96c1).

flutter_quill_extensions/lib/src/editor/image/image_menu.dart
EchoEllet: Fixed (https://github.com/singerdmx/flutter-quill/pull/2403/commits/fa09be78b0a9cb6335fa6cb1de73011092ef5a3c).

However, the open file location functionality is replaced with the open file (the saved file) on macOS since opening the directory of saved files requires [macOS entitlements](https://docs.flutter.dev/platform-integration/macos/building#setting-up-entitlements) (platform setup) in case it was saved to the user owned directory (e.g., `Pictures`) and that's a breaking change.

See: https://pub.dev/packages/url_launcher#macos-file-access-configuration

flutter_quill_extensions/lib/src/editor/image/image_save_utils.dart
EchoEllet: The getter and setter are static fields of `ImageSaver`. Doesn't fully solve the issue, though we don't have a specific consistent design in the project.

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,12 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
 
+    <!-- To save images on Android 9 (API 28) and earlier -->
+    <!-- See: https://pub.dev/packages/quill_native_bridge#-saving-images-to-the-gallery -->
+    <uses-permission
+        android:name="android.permission.WRITE_EXTERNAL_STORAGE"
+        android:maxSdkVersion="28" />
+
```
EchoEllet, line 10: Not a breaking change, this was already required previously though the value was `29` when it should `28`.

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -726,6 +734,18 @@
...
+			isa = XCLocalSwiftPackageReference;
+			relativePath = Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage;
+		};
+/* End XCLocalSwiftPackageReference section */
+/* Begin XCSwiftPackageProductDependency section */
+		78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */ = {
+			isa = XCSwiftPackageProductDependency;
+			productName = FlutterGeneratedPluginSwiftPackage;
+		};
+/* End XCSwiftPackageProductDependency section */
```
EchoEllet, line 748: All of these iOS/macOS Xcode project changes to the example to add Swift package manager integration that will be automatically added when running the app Flutter using the `main` channel (previously `master`).

Migrate with minimal steps (without running the app):

```shell
flutter channel main --no-cache-artifacts
flutter upgrade
flutter config --enable-swift-package-manager
flutter build ios --config-only
flutter build macos --config-only
```

However, those changes to the example app's Xcode project expect a `FlutterGeneratedPluginSwiftPackage` which is generated in Flutter 3.24 or higher (empty for forward compatibility, supported only in the `main` channel). 

This means all contributors need to use Flutter 3.24 or higher to run the iOS or macOS example even if CocoaPods is not supported.

See https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-app-developers

example/macos/Podfile.lock
```diff
@@ -1,48 +1,16 @@
...
-  - gal (1.0.0):
-    - Flutter
-    - FlutterMacOS
-  - quill_native_bridge_macos (0.0.1):
-    - FlutterMacOS
-  - url_launcher_macos (0.0.1):
-    - FlutterMacOS
-  - video_player_avfoundation (0.0.1):
-    - Flutter
-    - FlutterMacOS
```
EchoEllet, line 14: Since all of those native dependencies support Swift package manager, they are no longer used in CocoaPods, CocoaPods integration can be removed from the example to improve the project's build time.

However, removing it will require enabling SPM which is only available in the `main` channel. All Flutter plugins already migrated to SPM including the example app (CocoaPods integration not removed).

example/macos/Runner/DebugProfile.entitlements
```diff
@@ -10,7 +10,7 @@
 	<true/>
 	<key>com.apple.security.network.client</key>
 	<true/>
-	<key>com.apple.security.files.user-selected.read-only</key>
+	<key>com.apple.security.files.user-selected.read-write</key>
```
EchoEllet, line 13: Changed since that's required when saving a file to the user directory with [`NSPanel`](https://developer.apple.com/documentation/appkit/nspanel). No runtime permission is needed.

However, attempting to save the image to the user's selected file without this change will crash the app. That's the main reason why `prefersGallerySave` is `true` instead of `false` in `ImageOptionsMenu`.

flutter_quill_extensions/lib/src/common/utils/utils.dart
```diff
@@ -35,71 +28,3 @@ bool isYouTubeUrl(String videoUrl) {
...
-  }
-}
-
-Future<void> _saveImageFromNetwork(Uri imageUrl) async {
-  final response = await http.get(
-    imageUrl,
-  );
-  if (response.statusCode != 200) {
-    throw Exception('Response to $imageUrl is not successful.');
-  }
```
EchoEllet, line 95: Legacy internal code, now the image bytes are loaded from the image provider directly.

lib/src/common/utils/quill_native_provider.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:meta/meta.dart';
+import 'package:quill_native_bridge/quill_native_bridge.dart';
+
+export 'package:quill_native_bridge/quill_native_bridge.dart';
+
+@visibleForTesting
+typedef DefaultQuillNativeBridge = QuillNativeBridge;
+
+abstract final class QuillNativeProvider {
```
EchoEllet, line 9: I don't know if this is confusing but this could also be `QuillNativeBridgeProvider`.

lib/src/editor/widgets/text/text_line.dart
```diff
@@ -6,7 +6,7 @@ import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
 import 'package:flutter/services.dart';
-import 'package:url_launcher/url_launcher_string.dart' show launchUrlString;
```
EchoEllet, line 9: Using `url_launcher_string.dart` is discouraged per https://pub.dev/packages/url_launcher#urls-not-handled-by-uri

EchoEllet (commented):
Ready for review, most of the changes are new keys added to `quill_en.arb`, tests, or changes to the macOS and iOS example app for Swift package manager integration.

singerdmx (approved):

EchoEllet:
Merged as I wasn’t sure if more time was needed for review, and we have other issues to focus on as well.
Flutter 3.27.0 is now stable and supports Swift package manager.

PR #2413, opened 2024-12-11
revert!: magnifier

EchoEllet:
## Description

**Remove the magnifier due to buggy behavior with many regressions.**

See #2406 for a list of issues, reasons, and details.

This reverts the two related PRs using:

```shell
git revert af691e69ed185b99899aa2dd9bb9a57620abbf05  # https://github.com/singerdmx/flutter-quill/pull/2026
git revert 2937dc8c955d02f97cf2c254da1db69799483dfc  # https://github.com/singerdmx/flutter-quill/pull/2128
```

And then removed anything related to the magnifier that was made in related PRs manually.

To try this branch:

```yaml
dependency_overrides:
  flutter_quill:
    git:
      url: https://github.com/singerdmx/flutter-quill.git
      ref: revert/magnifier-feat
```

## Related Issues

- *Fix #2406*
- *Revert #2026*
- *Revert #2128*
- *Fix #2376*
- *Fix #2153*
- *Related #2391*
- *Related #2124*
- *Related #2401*
- *Related #2111*
- *Related #2379*
- *Related #2402*
- *Related #2412*
- *Related #2049*
- *Related #2251*
- *Related #2116*
- #2338 (minor cleanup, see https://github.com/singerdmx/flutter-quill/pull/2338/commits/79a771a342c6a63b677739a658e5785aa7486c3e, https://github.com/singerdmx/flutter-quill/pull/2338/commits/f1a7e57f65b178634de4c1a05444cb8e73b81867 and https://github.com/singerdmx/flutter-quill/pull/2338/commits/b739b700cbae9c3d4427e4966963d97cebf0a852)

## Type of Change

- [ ] ✨ **Feature:** New functionality without breaking existing features.
- [x] 🛠️ **Bug fix:** Resolves an issue without altering current behavior.
- [ ] 🧹 **Refactor:** Code reorganization, no behavior change.
- [x] ❌ **Breaking:** Alters existing functionality and requires updates.
- [ ] 🧪 **Tests:** New or modified tests
- [ ] 📝 **Documentation:** Updates or additions to documentation.
- [x] 🗑️ **Chore:** Routine tasks, or maintenance.
- [ ] ✅ **Build configuration change:** Build/configuration changes.

Issue #2406, opened 2024-12-04
Revert the magnifier feature


EchoEllet:
Revert the magnifier feature (#2026) in v11 (as a breaking change) with all related changes.

It's already marked as experimental in v11.

### Reason

Too many regressions, bugs, and workarounds, made changes to many parts of the API in not very good design, [removed an assert check to dodge issues](https://github.com/singerdmx/flutter-quill/pull/2026#discussion_r1679744497), not implemented cleanly.

I don't have a strong opinion upon the revert since it doesn't affect me but I do have about this: We should focus on improving existing features (as it's done with #2403), improve code coverage, and docs, fix existing issues, bugs, the not very good design we have before we move further forward, we have discussed this many times in different threads (e.g., #2235) with more details and reasons (might create an issue for collecting all such details) and it seems we agreed on the same opinion.

Implementing solutions poorly without understanding the details or doing them properly is strongly discouraged, as it will cause the project to fall apart in the long term.

This is not a justification though, we should test features thoughtfully before releasing them as stable, otherwise, all users including the end-users will be a test site.

Issue #1504 will reopened if reverted.

### Regressions

- #2376
- #2153
- #2391
- #2124
- #2401
- #2111
- #2379
- https://github.com/singerdmx/flutter-quill/issues/2412#issuecomment-2531084869

I found some other bugs while testing and working on different areas (e.g., https://github.com/singerdmx/flutter-quill/issues/2323#issuecomment-2418862462 or https://github.com/singerdmx/flutter-quill/issues/2323#issuecomment-2419086701).

### Related PRs

PRs that made fixes to the issue:

- #2049
- #2402
- #2251
- #2116
- #2338 (minor cleanup, see https://github.com/singerdmx/flutter-quill/pull/2338/commits/79a771a342c6a63b677739a658e5785aa7486c3e, https://github.com/singerdmx/flutter-quill/pull/2338/commits/f1a7e57f65b178634de4c1a05444cb8e73b81867 and https://github.com/singerdmx/flutter-quill/pull/2338/commits/b739b700cbae9c3d4427e4966963d97cebf0a852)
- #2117

PRs to implement the feature:

- #2026
- #2128

singerdmx:
Agree

jonasbernardo:
I agree, I'm having a lot of problems because of this magnifier

EchoEllet:
> I agree, I'm having a lot of problems because of this magnifier

[...]

PR #2427, opened 2025-01-05
Removed ClipboardMonitor from ClipboardAction.paste type button

### LINKED ISSUE #2421: ClipboardMonitor causes background lag when enabling clipboard buttons

### Flutter Quill version

flutter_quill: ^11.0.0-dev.17  &  flutter_quill_extensions: ^11.0.0-dev.7

### Steps to reproduce

1. Add `QuillSimpleToolbar` and pass `QuillSimpleToolbarConfig` with `showClipboardCut`, `showClipboardCopy`, or `showClipboardPaste` set to **true**
2. Since it adds the button with Clipboard Service it causes performance issues:
Source code for reference:
```dart
import 'dart:async';

import 'package:flutter/foundation.dart';

import '../../common/utils/widgets.dart';
import '../../editor_toolbar_controller_shared/clipboard/clipboard_service_provider.dart';
import '../../l10n/extensions/localizations_ext.dart';
import '../base_button/base_value_button.dart';
import '../simple_toolbar.dart';

enum ClipboardAction { cut, copy, paste }

class ClipboardMonitor {
  bool _canPaste = false;
  bool get canPaste => _canPaste;
  Timer? _timer;

  void monitorClipboard(bool add, void Function() listener) {
    if (kIsWeb) return;
    if (add) {
      _timer = Timer.periodic(
          const Duration(seconds: 1), (timer) => _update(listener));
    } else {
      _timer?.cancel();
    }
  }

  Future<void> _update(void Function() listener) async {
    final clipboardService = ClipboardServiceProvider.instance;
    if (await clipboardService.hasClipboardContent) {
      _canPaste = true;
      listener();
    }
  }
}

class QuillToolbarClipboardButton extends QuillToolbarToggleStyleBaseButton {
  QuillToolbarClipboardButton({
    required super.controller,
    required this.clipboardAction,
    super.options = const QuillToolbarToggleStyleButtonOptions(),

    /// Shares common options between all buttons, prefer the [options]
    /// over the [baseOptions].
    super.baseOptions,
    super.key,
  });

  final ClipboardAction clipboardAction;

  final ClipboardMonitor _monitor = ClipboardMonitor();

  @override
  State<StatefulWidget> createState() => QuillToolbarClipboardButtonState();
}

class QuillToolbarClipboardButtonState
    extends QuillToolbarToggleStyleBaseButtonState<
        QuillToolbarClipboardButton> {
  @override
  bool get currentStateValue {
    switch (widget.clipboardAction) {
      case ClipboardAction.cut:
        return !controller.readOnly && !controller.selection.isCollapsed;
      case ClipboardAction.copy:
        return !controller.selection.isCollapsed;
      case ClipboardAction.paste:
        return !controller.readOnly && (kIsWeb || widget._monitor.canPaste);
    }
  }

  void _listenClipboardStatus() => didChangeEditingValue();

  @override
  void addExtraListener() {
    if (widget.clipboardAction == ClipboardAction.paste) {
      widget._monitor.monitorClipboard(true, _listenClipboardStatus);
    }
  }

  @override
  void removeExtraListener(covariant QuillToolbarClipboardButton oldWidget) {
    if (widget.clipboardAction == ClipboardAction.paste) {
      oldWidget._monitor.monitorClipboard(false, _listenClipboardStatus);
    }
  }

  @override
  String get defaultTooltip => switch (widget.clipboardAction) {
        ClipboardAction.cut => context.loc.cut,
        ClipboardAction.copy => context.loc.copy,
        ClipboardAction.paste => context.loc.paste,
      };

  @override
  IconData get defaultIconData => switch (widget.clipboardAction) {
        ClipboardAction.cut => Icons.cut_outlined,
        ClipboardAction.copy => Icons.copy_outlined,
        ClipboardAction.paste => Icons.paste_outlined,
      };

  void _onPressed() {
    switch (widget.clipboardAction) {
      case ClipboardAction.cut:
        controller.clipboardSelection(false);
        break;
      case ClipboardAction.copy:
        controller.clipboardSelection(true);
        break;
      case ClipboardAction.paste:
        controller.clipboardPaste();
        break;
    }
    afterButtonPressed?.call();
  }

  @override
    final childBuilder = this.childBuilder;
    if (childBuilder != null) {
      return childBuilder(
        options,
        QuillToolbarToggleStyleButtonExtraOptions(
          context: context,
          controller: controller,
          onPressed: _onPressed,
          isToggled: currentValue,
        ),
      );
    }

    return UtilityWidgets.maybeTooltip(
        message: tooltip,
        child: QuillToolbarIconButton(
          icon: Icon(
            iconData,
            size: iconSize * iconButtonFactor,
          ),
          isSelected: false,
          onPressed: currentValue ? _onPressed : null,
          afterPressed: afterButtonPressed,
          iconTheme: iconTheme,
        ));
  }
}
```

3. Since the `QuillToolbarClipboardButton` was added to the Widget tree, it starts affecting the app performance & memory on the background, even if you are not actively copying or pasting with clipboard at the moment.
```dart
W/Looper  ( 4641): PerfMonitor longMsg : seq=69 plan=18:41:47.480 late=0ms wall=1570ms h=android.os.Handler c=io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0 procState=-1
W/Looper  ( 4641): PerfMonitor longMsg : seq=71 plan=18:41:48.479 late=572ms wall=1418ms h=android.os.Handler c=io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0 procState=-1
W/Looper  ( 4641): PerfMonitor longMsg : seq=73 plan=18:41:49.479 late=991ms wall=1410ms h=android.os.Handler c=io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0 procState=-1
W/Looper  ( 4641): PerfMonitor longMsg : seq=74 plan=18:41:50.559 late=1323ms wall=1427ms h=android.os.Handler c=io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0 procState=-1
...
...
```

### Expected results

Expected Clipboard to not be as heavy for the application and to not cause performance issues.

### Actual results

ClipboardService causes background lag when used.

Maksim-Nikolaev:
## Description

According the issue [#2421](https://github.com/singerdmx/flutter-quill/issues/2421) ClipboardMonitor was causing lag if "paste" button was enabled in QuillSimpleToolbar. 

That was because of the periodic 1 second check for Clipboard content. In case with large strings or images in clipboard, it would lag out the application and slow down everything drastically. 
```dart
class ClipboardMonitor {
  bool _canPaste = false;
  bool get canPaste => _canPaste;
  Timer? _timer;

  void monitorClipboard(bool add, void Function() listener) {
    if (kIsWeb) return;
    if (add) {
      _timer = Timer.periodic(
          const Duration(seconds: 1), (timer) => _update(listener));
    } else {
      _timer?.cancel();
    }
  }

  Future<void> _update(void Function() listener) async {
    final clipboardService = ClipboardServiceProvider.instance;
    if (await clipboardService.hasClipboardContent) {
      _canPaste = true;
      listener();
    }
  }
}
```

This is not a breaking change, however it will change the behavior of  availability for "paste" button for QuillSimpleToolbar since it always be enabled (except of readonly mode).

## Related Issues

- *Fix #2421*

## Type of Change

- [x] 🛠️ **Bug fix:** Resolves an issue without changing current behavior.


EchoEllet:
Thanks for your contribution.

> This is not a breaking change

A breaking change is not the kind that requires users to update their code to compile, but to also keep the current state without any behavior breaking.

Changing the default of a value is also a breaking change.

What would happen if the user pressed the paste button and there was nothing to paste?

I suggest providing the option to override the `canPaste` state instead, so they don't need to depend on `ClipboardMonitor`.

I'm with the idea of disabling something that's causing issues. However, users should have the option to revert to the old behavior in case they still want `ClipboardMonitor` with a signle boolean parameter.

Maksim-Nikolaev:
> A breaking change is not the kind that requires users to update their code to compile, but to also keep the current state without any behavior breaking. Changing the default of a value is also a breaking change.

Ok I understand

> What would happen if the user pressed the paste button and there was nothing to paste?

Nothing will happen. There are few checks that prevent breaking everything on calling "paste" function. 

I double checked it with Web & Android version and I strongly believe that iOS will not break either (don't have it right now, so I can't check)

QuillController.clipboardPaste handles this quite well, here is the full function:


```dart
  /// Returns whether paste operation was handled here.
  /// [updateEditor] is called if paste operation was successful.
  @experimental
  Future<bool> clipboardPaste({void Function()? updateEditor}) async {
    if (readOnly || !selection.isValid) return true;

    final clipboardConfig = config.clipboardConfig;

    if (await clipboardConfig?.onClipboardPaste?.call() == true) {
      updateEditor?.call();
      return true;
    }

    final pasteInternalImageSuccess = await _pasteInternalImage();
    if (pasteInternalImageSuccess) {
      updateEditor?.call();
      return true;
    }

    const enableExternalRichPasteDefault = true;
    if (clipboardConfig?.enableExternalRichPaste ??
        enableExternalRichPasteDefault) {
      final pasteHtmlSuccess = await pasteHTML();
      if (pasteHtmlSuccess) {
        updateEditor?.call();
        return true;
      }

      final pasteMarkdownSuccess = await pasteMarkdown();
      if (pasteMarkdownSuccess) {
        updateEditor?.call();
        return true;
      }
    }

    final clipboardService = ClipboardServiceProvider.instance;

    final onImagePaste = clipboardConfig?.onImagePaste;
    if (onImagePaste != null) {
      final imageBytes = await clipboardService.getImageFile();

      if (imageBytes != null) {
        final imageUrl = await onImagePaste(imageBytes);
        if (imageUrl != null) {
          replaceText(
            plainTextEditingValue.selection.end,
            0,
            BlockEmbed.image(imageUrl),
            null,
          );
          updateEditor?.call();
          return true;
        }
      }
    }

    final onGifPaste = clipboardConfig?.onGifPaste;
    if (onGifPaste != null) {
      final gifBytes = await clipboardService.getGifFile();
      if (gifBytes != null) {
        final gifUrl = await onGifPaste(gifBytes);
        if (gifUrl != null) {
          replaceText(
            plainTextEditingValue.selection.end,
            0,
            BlockEmbed.image(gifUrl),
            null,
          );
          updateEditor?.call();
          return true;
        }
      }
    }

    // Only process plain text if no image/gif was pasted.
    // Snapshot the input before using `await`.
    // See https://github.com/flutter/flutter/issues/11427
    final plainText = (await Clipboard.getData(Clipboard.kTextPlain))?.text;

    if (plainText != null) {
      final plainTextToPaste = await getTextToPaste(plainText);
      if (pastePlainTextOrDelta(plainTextToPaste,
          pastePlainText: _pastePlainText, pasteDelta: _pasteDelta)) {
        updateEditor?.call();
        return true;
      }
    }

    final onUnprocessedPaste = clipboardConfig?.onUnprocessedPaste;
    if (onUnprocessedPaste != null) {
      if (await onUnprocessedPaste()) {
        updateEditor?.call();
        return true;
      }
    }

    return false;
  }
```



> I suggest providing the option to override the `canPaste` state instead, so they don't need to depend on `ClipboardMonitor`.

I will then provide such option.
As an extra step, if we are keeping ClipboardMonitor, I suggest we optimize it at least to reduce the amount of checks. Can I make the timer to 5 seconds and prevent overlapping checks? (So it won't start checking if it didn't finish the previous one)

Maksim-Nikolaev:
Please check the most recent commit.
I added **bool? enableClipboardPaste** in lib\src\toolbar\config\simple_toolbar_config.dart for QuillSimpleToolbarConfig 

This is passed directly to ClipboardButton with .paste action

As for the ClipboardButton itself and monitor:

- [x] enableClipboardPaste flag can be null or bool:
```dart
  /// If null, uses in-built [ClipboardMonitor]
  /// If true, paste button is enabled (if true & not readonly)
  /// If false, paste button is disabled
  final bool? enableClipboardPaste;
```
- [x] Updated ClipboardMonitor periodic timer from 1 second to 5 seconds
- [x] Update ClipboardMonitor to not overlap with previous task of itself (in case of big clipboard content)
- [x] Changed getter for "currentStateValue" to prioritize the enableClipboardPaste flag:
```dart
        return !controller.readOnly &&
            (kIsWeb || (widget.enableClipboardPaste ?? _monitor.canPaste));
```

EchoEllet:
> Updated ClipboardMonitor periodic timer from 1 second to 5 seconds

This is still a breaking change. 
Could you minimize the changes to only override the can paste state (whether the button is enabled)?

> I added bool? enableClipboardPaste in lib\src\toolbar\config\simple_toolbar_config.dart for QuillSimpleToolbarConfig

This option should be only in the button config, not the toolbar config.

Maksim-Nikolaev:
> This is still a breaking change.

Reverted to 1 second

Maksim-Nikolaev:
> This option should be only in the button config, not the toolbar config.

I agree that it would be the best way to do it. Unfortunately I couldn't find the proper place to add a new value, could you elaborate here?
```dart
  final QuillToolbarToggleStyleButtonOptions clipboardPaste;
```

```dart
import 'package:flutter/foundation.dart' show immutable;
import 'package:meta/meta.dart';

import '../base_button_options.dart';

class QuillToolbarToggleStyleButtonExtraOptions
    extends QuillToolbarBaseButtonExtraOptions
    implements QuillToolbarBaseButtonExtraOptionsIsToggled {
  const QuillToolbarToggleStyleButtonExtraOptions({
    required super.controller,
    required super.context,
    required super.onPressed,
    required this.isToggled,
  });

  @override
  final bool isToggled;
}

@immutable
class QuillToolbarToggleStyleButtonOptions
    extends QuillToolbarBaseButtonOptions<QuillToolbarToggleStyleButtonOptions,
        QuillToolbarToggleStyleButtonExtraOptions> {
  const QuillToolbarToggleStyleButtonOptions({
    super.iconData,
    super.iconSize,
    super.iconButtonFactor,
    super.tooltip,
    super.afterButtonPressed,
    super.iconTheme,
    super.childBuilder,
  });
}
```

Maksim-Nikolaev:
> Could you minimize the changes to only override the can paste state (whether the button is enabled)?

I would love to, but since it's coded in the way that we have little control over ClipboardMonitor it's difficult to update it's state or to stop the timer when the button already exists.

To stop the timer we need to trigger "didUpdateWidget" and add/remove extra listeners but it only happens when controller changes.
I kept the original handler to imitate the same behavior, as well as added some custom code to add/remove listener if "enableClipboardPaste" value changed from parent.

Maksim-Nikolaev:
I think adding "isEnabled" to **QuillToolbarBaseButtonOptions** would be the best option, but this would require adding the functionality of this option to all the buttons. I can do it if you give me a green light.
Otherwise I'm open for recommendations how to handle it for just 1 type of button

Maksim-Nikolaev:
- [x] Moved **enableClipboardPaste** to **QuillToolbarBaseButtonOptions**

On the video you can see that it works for all 3 cases (null, true, false) and the timer will continue as normal if it is set back to "null" 

https://github.com/user-attachments/assets/b076a992-27c8-4a51-8a5f-a6d38d5ea0f6

singerdmx (approved):

singerdmx:
@EchoEllet can we merge this?

EchoEllet:
> Updated ClipboardMonitor periodic timer from 1 second to 5 seconds

This is still a breaking change. 
Could you minimize the changes to only override the can paste state (whether the button is enabled)?

> I added bool? enableClipboardPaste in lib\src\toolbar\config\simple_toolbar_config.dart for QuillSimpleToolbarConfig

This option should be only in the button config, not the toolbar config.
 
> can we merge this?

I will work on it soon. However, it's not a high-priorty issue given that I have disabled the buttons by default (revert to the default behavior).

Maksim-Nikolaev:
> This is still a breaking change.

I adjusted the code changes followed by your requests. Check some comments above.
Timer is reverted back to 1 second
The flag is now in the button config 



I really appreciate your time and work on this!

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -16,27 +16,49 @@ class ClipboardMonitor {
...
   Future<void> _update(void Function() listener) async {
+    // If the clipboard is already being checked, return early.
+    if (_isCheckingClipboard) {
+      return;
+    }
+
+    // Prevent multiple clipboard checks at the same time.
+    _isCheckingClipboard = true;
+
+    // Initialize the clipboard service.
```
EchoEllet, line 43: This is static, and a new instance will not be created. Could you remove comments for code that's already obvious?

We're trying to not over comment, and only comment when there is something confusing, not clear or the code is not self explanatory and descriptive, in other words anything that can't be understood from other developers perspective.

lib/src/toolbar/config/base_button_options.dart
```diff
@@ -65,6 +66,11 @@ class QuillToolbarBaseButtonOptions<T, I> {
   /// Use custom theme
   final QuillIconTheme? iconTheme;
 
+  /// Disables the [ClipboardMonitor] of [QuillToolbarClipboardButton] if not null
+  /// If true or false, it will enable or disable the clipboard paste button
+  /// Only affects the clipboard paste button
+  final bool? enableClipboardPaste;
```
EchoEllet, line 72: I suggest:

```dart
/// Determines if the paste button is enabled. 
///
/// Defaults to [ClipboardMonitor] in case of `null`, which will check if the clipboard has content to paste every second, and only then enable the button, indicating to the user that they can paste something.
///
/// Set to `true` to always enable it even if the clipboard has no content to paste, which will do nothing on a press.
```

However, it's not final and could use improvements. I will review it soon.

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -16,27 +16,49 @@ class ClipboardMonitor {
...
     final clipboardService = ClipboardServiceProvider.instance;
+
+    // Check if the clipboard has content.
     if (await clipboardService.hasClipboardContent) {
+      // If the clipboard has content, set the flag to true.
```
EchoEllet, line 48: It's already obvious what the code does, and over commenting can complicate the code readability.

CHANGELOG.md
```diff
@@ -10,6 +10,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Added
+
+- Added **enableClipboardPaste** flag to ButtonOptions to control clipboard paste button enabled state and ClipboardMonitor [#2421](https://github.com/singerdmx/flutter-quill/issues/2421)
+    - null (default): Original ClipboardMonitor behavior (1 second periodic checks)
+    - false: Disable ClipboardMonitor & button is disabled
+    - true: Disable ClipboardMonitor & button is enabled
+
+### Changed
+
+- Prevented ClipboardMonitor from doing multiple simultaneous clipboard checks to improve performance. [#2421](https://github.com/singerdmx/flutter-quill/issues/2421)
```
EchoEllet, line 22: The quality of the `CHANGELOG.MD` is a priority and will need a few changes, I will work on it soon.

EchoEllet (changes requested):

EchoEllet:
> I really appreciate your time and work on this!

Thanks, I appreciate your contributions, too.
There are a few changes, and I will work on it today.

Maksim-Nikolaev:
I applied some changes as you requested. Obviously feel free to change the wording/comments to your liking! Thanks again

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -16,27 +16,43 @@ class ClipboardMonitor {
...
       listener();
     }
+
+    _isCheckingClipboard = false;
   }
 }
 
 class QuillToolbarClipboardButton extends QuillToolbarToggleStyleBaseButton {
-  QuillToolbarClipboardButton({
+  const QuillToolbarClipboardButton({
```
EchoEllet, line 55: There is almost no use in adding the `const`, the `controller` can never be const so neither `QuillToolbarClipboardButton` can. It would only make the commit history more complex, the code more complex, and slightly confusing and introducing them removing it (if needed in case of the constructor body) would be a breaking change. I was addicted to using the `const` keyword but it doesn't bring much.

image

Maksim-Nikolaev: I see, my bad for not going in-depth with that. In my experience it's the default-ish way in Flutter to mark class constructors with const and even flutter analysis options usually prefers const there

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -66,23 +82,57 @@ class QuillToolbarClipboardButtonState
...
     }
   }
 
   @override
   void removeExtraListener(covariant QuillToolbarClipboardButton oldWidget) {
-    if (widget.clipboardAction == ClipboardAction.paste) {
-      oldWidget._monitor.monitorClipboard(false, _listenClipboardStatus);
+    if (widget.clipboardAction == ClipboardAction.paste &&
+        widget.options.enableClipboardPaste == null) {
+      _monitor.monitorClipboard(false, _listenClipboardStatus);
```
EchoEllet, line 135: Could you add a private getter called `shouldUseClipboardMonitor` that is only `true` when `enableClipboardPaste` is `null` and then update related parts such as those to make use of this to make it easier and faster to understand from other developers's perspective.

lib/src/toolbar/buttons/clipboard_button.dart
EchoEllet: > my bad

No worries, I just wanted to ensure that you're aware of this in case you're not.

> mark class constructors with const

You only need to mark a class, but not the subclasses of that class. I'm uncertain if the analysis has a lint for checking the `const` or `@immutable`.

> flutter analysis options usually prefers const there

That's in case it's possible to instate a constructor that can be `const`, or if the superclass marked with `@immutable` such as `Widget`, the `StatelessWidget` and `StatefulWidget` doesn't need `@immutable` annotation.

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -100,6 +147,11 @@ class QuillToolbarClipboardButtonState
         ClipboardAction.paste => Icons.paste_outlined,
       };
 
+  bool get shouldUseClipboardMonitor {
```
EchoEllet, line 150: Unless there is a need, it's preferable to keep this private (`_shouldUseClipboardMonitor` instead).

lib/src/toolbar/config/base_button_options.dart
```diff
@@ -65,6 +66,13 @@ class QuillToolbarBaseButtonOptions<T, I> {
   /// Use custom theme
   final QuillIconTheme? iconTheme;
 
+  /// Determines if the paste button is enabled. The button is disabled and cannot be clicked if set to `false`.
+  ///
+  /// Defaults to [ClipboardMonitor] in case of `null`, which checks if the clipboard has content to paste every second and only then enables the button, indicating to the user that they can paste something.
+  ///
+  /// Set it to `true` to enable it even if the clipboard has no content to paste, which will do nothing on a press.
+  final bool? enableClipboardPaste;
+
```
EchoEllet, line 75: * I'm confused, why is this change in the `QuillToolbarBaseButtonOptions`? This is the base for all buttons, and any configuration specific to the paste button should be in `QuillToolbarPasteButton` instead.
* It's necessary to document if this is only applicable if the clipboard action is `ClipboardAction.paste`.

Maksim-Nikolaev: Could you please take care of the location where to put the "enableClipboardPaste" flag? 

ClipboardButtons use "QuillToolbarToggleStyleButtonOptions" and there is no "QuillToolbarPasteButton" class. 
I'm not 100% familiar with how to introduce it without changes to be breaking.

```dart
  final QuillToolbarToggleStyleButtonOptions clipboardCut;
  final QuillToolbarToggleStyleButtonOptions clipboardCopy;
  final QuillToolbarToggleStyleButtonOptions clipboardPaste;
```

EchoEllet: ClipboardButtons use "QuillToolbarToggleStyleButtonOptions".

It looks like the `QuillToolbarClipboardButton` doesn't have a custom config class (`QuillToolbarClipboardButtonConfig`), I'm not sure why but it shouldn't even if it doesn't require extra parameters. I will create a config class for this button and you can move this flag there instead.

> here is no "QuillToolbarPasteButton" class.

I updated my message, I meant the config class of `QuillToolbarClipboardButton` which is missing.

EchoEllet: > ntroduce it without changes to be breaking.

You're right, it's not possible without any breaking in case users were already using `QuillToolbarToggleStyleButtonOptions` for those buttons (see #2433).

To be honest I'm not motivated to work on anything related to `QuillSimpleToolbar`, the design is a pain to work with and almost always confusing despite it's not very complex, our long plan is to completely rework the toolbar with a new modern one, better API design and usability while keeping the old toolbar backward compatible.

If #2433 merged, you can access `enableClipboardPaste` inside `QuillToolbarClipboardButtonState`:

```dart
widget._options.enableClipboardPaste;
```

But again, wouldn't it be better to mark this feature as experimental, rework on it later (to avoid breaking)? It's almost not a priority at least IMO. Working on it is not a good time investment which is why I updated `README.md` to include a note about contributing: https://github.com/singerdmx/flutter-quill#-contributing

@CatHood0 Do you think it worth introducing this breaking (changing `options` from `QuillToolbarToggleStyleButtonOptions` to `QuillToolbarClipboardButtonOptions` and slightly inconsistent API).

Maksim-Nikolaev: > But again, wouldn't it be better to mark this feature as experimental, rework on it later (to avoid breaking)? It's almost not a priority at least IMO. Working on it is not a good time investment

I can understand the pain and now I realize more how hard it is to maintain open source project. I discovered that the paste button caused **extreme** lag in the app I'm working on, therefore I thought I'd contribute.

Otherwise in my case the usage of such button can be simplified to calling "controller.clipboardPaste" and can be achieved with custom button. 

Once again, really appreciate your time and patience with this PR and I hope I don't cause too much issues for you

EchoEllet: > I don't cause too much issues for you

This is a project issue and not exclusive to this PR.

>  I realize more how hard it is to maintain open source project.

It's not very hard if you have a good solid foundation from the start, but there was a time when we kept merging PRs without much review and also introducing changes without much consideration or any tests at all, and that made it harder for us to fix those issues or maintain the project.

For example, the magnifier feature was introduced and merged quickly that people started depending on it (see #2406 for a list of regressions and issues), which caused major issues that affected the editor experience, and reverting to the old behavior is no longer trivial (see #2413), introducing breaking changes caused many issues for the app developers. That's why I think adding new features is not a priority, especially those that add more on the surface for more issues, this is not an issue with your code.

EchoEllet: I have merged #2433, please rebase and use:

```dart
widget._options.enableClipboardPaste;
```

While also removing those two lines:

```diff
- // TODO: This field will be used by the PR: https://github.com/singerdmx/flutter-quill/pull/2427
- // ignore: unused_field
final QuillToolbarClipboardButtonOptions? _options;
```

> [!NOTE]
> This is far from an efficient solution, but we will continue with it as we plan on reworking the toolbar at a later point.

Maksim-Nikolaev: I applied the changes, please review 👍

lib/src/toolbar/buttons/clipboard_button.dart
```diff
@@ -143,16 +193,17 @@ class QuillToolbarClipboardButtonState
...
+        icon: Icon(
+          iconData,
+          size: iconSize * iconButtonFactor,
+        ),
+        isSelected: false,
+        onPressed: currentValue ? _onPressed : null,
+        afterPressed: afterButtonPressed,
+        iconTheme: iconTheme,
+      ),
+    );
```
EchoEllet, line 207: If the trailing comma (`,`) is the only change, then please revert, this is a common style inconsistency and it should be fixed in a separate PR (#2442).

This change will make it harder to track the commit history, mainly we want to be able to see the PR and the commit that made a feature.

lib/src/toolbar/config/base_button_options.dart
EchoEllet: > It's necessary to document if this is only applicable if the clipboard action is ClipboardAction.paste. Since the QuillToolbarClipboardButton is the same widget for cut, copy, and paste buttons.

This is still not addressed.

lib/src/toolbar/buttons/clipboard_button.dart
EchoEllet: Fixed.

EchoEllet:
>  info • Constructors in '@immutable' classes should be declared as 'const' • lib/src/toolbar/buttons/clipboard_button.dart:62:3 • prefer_const_constructors_in_immutables
   info • Field overrides a field inherited from 'QuillToolbarBaseButtonOptions' • lib/src/toolbar/config/buttons/clipboard_button_options.dart:28:15 • overridden_fields
   info • The member 'enableClipboardPaste' overrides an inherited member but isn't annotated with '@override' • lib/src/toolbar/config/buttons/clipboard_button_options.dart:28:15 • annotate_overrides

Could you confirm if you have rebased your branch?

```shell
$ git remote add upstream git@github.com:singerdmx/flutter-quill.git
$ git fetch upstream
$ git rebase upstream/master # Fix conflicts if any. Follow the git CLI instructions.
# git push --force # Force push if conflicts were solved or didn't occur.
```

lib/src/toolbar/config/base_button_options.dart
Maksim-Nikolaev: Removed it now from QuillToolbarBaseButtonOptions

lib/src/toolbar/config/buttons/toggle_style_options.dart
```diff
@@ -29,5 +29,6 @@ class QuillToolbarToggleStyleButtonOptions
     super.afterButtonPressed,
     super.iconTheme,
     super.childBuilder,
+    super.enableClipboardPaste,
```
EchoEllet, line 32: Could you revert any changes made to [`toggle_style_options.dart`](https://github.com/singerdmx/flutter-quill/pull/2427/files#diff-9258316c22a048d566a87172e61c3e053607f1dd38273c8fbdfcd57ff3ad06a7)?

EchoEllet:
> Error: Failed to create outline comment: Resource not accessible by integration

Could you [rebase your branch](https://github.com/singerdmx/flutter-quill/pull/2427#issuecomment-2598621102) once again?

EchoEllet:
> Error: Failed to create outline comment: Resource not accessible by integration

Could you [rebase your branch](https://github.com/singerdmx/flutter-quill/pull/2427#issuecomment-2598621102) once again?

Maksim-Nikolaev:
> Could you [rebase your branch](https://github.com/singerdmx/flutter-quill/pull/2427#issuecomment-2598621102) once again?

Rebased again.

> Please fix the analysis issue. https://dart.dev/tools/linter-rules/prefer_const_constructors_in_immutables

Added const to QuillToolbarClipboardButton constructor.

PR #2447, opened 2025-01-19
Fix: customStyleBuilder is not accumulating styles

### LINKED ISSUE #2425: customStyleBuilder Does Not Accumulate Styles for Multiple Attributes

### The question

The `customStyleBuilder` in `flutter_quill` does not accumulate styles when multiple attributes are applied to the same text. Instead, it processes each attribute independently, leading to only the last attribute being applied. This behavior makes it impossible to correctly combine styles like bold, italic, underline, and custom font sizes.

### Steps to Reproduce
Create a QuillEditor with a customStyleBuilder to handle custom styles.
Apply multiple styles (e.g., bold, italic, and size) to the same text.
Observe that only one of the styles is applied (the last one processed).

### Expected Behavior
All styles (bold, italic, underline, size, etc.) should be combined into a single TextStyle and applied simultaneously to the text.

### Actual Behavior
Only one style (the last one processed by customStyleBuilder) is applied, overriding any previously applied styles.

### Example Code
```
  customStyleBuilder: (attribute) {
    // basically settings.fontSize is the size factor of a zoom mechanism already implemented, so i can make the text larger or smaller by pinching the screen, without modifying the quill editor document
    TextStyle style = DefaultTextStyle.of(context).style.copyWith(fontSize: settings.fontSize * 20);

    if (attribute.key == "bold" && attribute.value == true) {
      style = style.copyWith(fontWeight: FontWeight.bold);
    }
    if (attribute.key == "italic" && attribute.value == true) {
      style = style.copyWith(fontStyle: FontStyle.italic);
    }
    if (attribute.key == "underline" && attribute.value == true) {
      style = style.copyWith(decoration: TextDecoration.underline);
    }
    // the thing I need the most, specific size attribute 
    if (attribute.key == "size") {
      final size = double.parse(attribute.value.toString());
      style = style.copyWith(fontSize: settings.fontSize * size);
    }

    return style;
  },
)
```
(read comments)

Can you help me with this? Thank you so much

CatHood0:
## Description

Fix issue where the `customStyleBuilder` does not accumulate styles when multiple attributes are applied to the same text.

## Related Issues

*Fix #2425*

- [x] ❌ **Breaking:** Alters existing functionality and requires updates.

### DIFF
--- a/lib/src/editor/widgets/delegate.dart
+++ b/lib/src/editor/widgets/delegate.dart
@@ -10,7 +10,8 @@ import '../editor.dart';
 import '../raw_editor/raw_editor.dart';
 import 'text/text_selection.dart';
 
-typedef CustomStyleBuilder = TextStyle Function(Attribute attribute);
+typedef CustomStyleBuilder = TextStyle Function(
+    Attribute attribute, TextStyle style);
 
 typedef CustomRecognizerBuilder = GestureRecognizer? Function(
     Attribute attribute, Leaf leaf);
--- a/lib/src/editor/widgets/text/text_line.dart
+++ b/lib/src/editor/widgets/text/text_line.dart
@@ -413,7 +413,7 @@ class _TextLineState extends State<TextLine> {
       final attr = attributes[key];
       if (attr != null) {
         /// Custom Attribute
-        final customAttr = widget.customStyleBuilder!.call(attr);
+        final customAttr = widget.customStyleBuilder!.call(attr, textStyle);
         textStyle = textStyle.merge(customAttr);
       }
     }


CatHood0:
@manudicri can you test if these changes fix your issue? I cannot really reproduce your issue, but i think that, with these changes, it could be fixed.

lib/src/editor/widgets/delegate.dart
```diff
@@ -10,7 +10,8 @@ import '../editor.dart';
 import '../raw_editor/raw_editor.dart';
 import 'text/text_selection.dart';
 
-typedef CustomStyleBuilder = TextStyle Function(Attribute attribute);
+typedef CustomStyleBuilder = TextStyle Function(
+    Attribute attribute, TextStyle style);
```
EchoEllet, line 14: Isn't this a breaking change if this is a public API?

manudicri:
Hi @CatHood0 ,
I think it could be fine, so I can add my custom styles by adding them to the TextStyle (even though I have to use .copyWith, I don't know if that could work).
Anyway, It's not urgent to me anymore since I found another workaround, but I thought it could be a problem to someone else, too.

CatHood0:
Ok, i will remove the current style property

PR #2477, opened 2025-02-17
Focus and open context menu on right click if unfocused

tjarvstrand:
## Description

If the editor is not focused and the user right clicks it, nothing currently happens. With these changes the editor is now focused and the context menu opened. Similar to how `TextField` works

## Type of Change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Removed unicode from `QuillText` element that causes weird caret behavior on empty lines [#2453](https://github.com/singerdmx/flutter-quill/pull/2453).
+* Request focus on secondary click.
 
 ## [11.0.0] - 2025-02-16
 
--- a/lib/src/editor/editor.dart
+++ b/lib/src/editor/editor.dart
@@ -6,6 +6,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/rendering.dart';
+import 'package:flutter/scheduler.dart';
 import 'package:flutter/services.dart';
 
 import '../common/utils/platform.dart';
@@ -554,6 +555,28 @@ class _QuillEditorSelectionGestureDetectorBuilder
       }
     } finally {
       _state._requestKeyboard();
+    }
+  }
+
+  /// onSingleTapUp for mouse right click
+  @override
+  void onSecondarySingleTapUp(TapUpDetails details) {
+    if (delegate.selectionEnabled &&
+        renderEditor != null &&
+        renderEditor!.selection.isCollapsed) {
+      renderEditor!.selectPositionAt(
+        from: details.globalPosition,
+        cause: SelectionChangedCause.longPress,
+      );
+    }
+
+    if (renderEditor?._hasFocus == false) {
+      _state._requestKeyboard();
+      SchedulerBinding.instance.addPostFrameCallback((_) {
+        super.onSecondarySingleTapUp(details);
+      });
+    } else {
+      super.onSecondarySingleTapUp(details);
     }
   }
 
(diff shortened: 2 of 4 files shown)


lib/src/editor/widgets/delegate.dart
```diff
@@ -199,10 +199,12 @@ class EditorTextSelectionGestureDetectorBuilder {
   /// onSingleTapUp for mouse right click
   @protected
   void onSecondarySingleTapUp(TapUpDetails details) {
-    // added to show toolbar by right click
-    if (checkSelectionToolbarShouldShow(isAdditionalAction: true)) {
-      editor!.showToolbar();
-    }
+    SchedulerBinding.instance.addPostFrameCallback((_) {
+      // added to show toolbar by right click
```
CatHood0, line 203: Why do we suddenly use `SchedulerBinding`? We didn't use it here before.

lib/src/editor/editor.dart
```diff
@@ -557,6 +558,28 @@ class _QuillEditorSelectionGestureDetectorBuilder
     }
   }
 
+  /// onSingleTapUp for mouse right click
+  @override
+  void onSecondarySingleTapUp(TapUpDetails details) {
+    if (delegate.selectionEnabled &&
+        renderEditor != null &&
+        renderEditor!.selection.isCollapsed) {
```
CatHood0, line 566: I think we shouldn't check if the selection is collapsed or not, it doesn't really matter much, do you think?

As far as i've checked, if you right click whether you're selecting a portion of text or not, the cursor should update (which means we force a collapsed selection)

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Removed unicode from `QuillText` element that causes weird caret behavior on empty lines [#2453](https://github.com/singerdmx/flutter-quill/pull/2453).
+* Request focus on secondary click.
```
CatHood0, line 16: I think maybe you should be a little more clear with the fix that was made in this PR (you should also add the link to the PR, you can take the other changes as a reference)

I think a better description would be:

```markdown
* Focus and open context menu on right click if unfocused [#2477](https://github.com/singerdmx/flutter-quill/pull/2477)
```

CatHood0 (changes requested):

lib/src/editor/widgets/delegate.dart
tjarvstrand: It was added here because it seems that when right clicking in the editor after it has lost focus, some part of the state (presumably related to regaining focus) has not been fully updated when reaching this part of the code so showing the context menu right away doesn't work.

I think this is an issue, because selected text doesn't get unselected when the editor looses focus, so to the user it still appears like they can right click the selected text.

The idea to use `SchedulerBinding` was taken from `onDoubleTapDown`.

lib/src/editor/editor.dart
tjarvstrand: In that case I believe we instead have to make sure to only collapse the selection when the editor will gain focus. Otherwise it will not be possible to select a range of text and then right click. Do you agree?

CHANGELOG.md
tjarvstrand: Sure, my bad!

EchoEllet: > I think maybe you should be a little more clear with the fix that was made in this PR (you should also add the link to the PR, you can take the other changes as a reference)
> 
> I think a better description would be:
> 
> ```markdown
> * Focus and open context menu on right click if unfocused [#2477](https://github.com/singerdmx/flutter-quill/pull/2477)
> ```

I think it's okay. The important part is to update CHANGELOG so we can update it ourselves before merging the PR in GitHub UI.

lib/src/editor/editor.dart
CatHood0: > In that case I believe we instead have to make sure to only collapse the selection when the editor will gain focus. Otherwise it will not be possible to select a range of text and then right click. Would you agree?

You have a point, but you are only partially right. The problem here is that we have two situations:

1. Right click inside the selection range
2. Right click outside the selection range

I was wrong and did not specify the last aspect. I meant that it would be better if when the right click is done outside the selection, it would collapse, and if not, it would remain the same.

lib/src/editor/widgets/delegate.dart
CatHood0: It seems like some extra work would have to be done to see what might be going on in the editor (this is outside of what PR does).

CatHood0 (approved):
LGTM.

PR #2482, opened 2025-02-18
Chore: added cache for toPlainText in Document to avoid unnecessary text computing

CatHood0:
## Description

I added a variable that allows caching all the text in the document when needed. In this case, when `compose()` is called, this text will be reloaded, which will prevent important changes from NOT being added.

### Why we need this change?

If you look a bit at `raw_editor_state` and `raw_editor_state_selection_delegate_mixin,` you can notice that both in [`_getGlyphHeights`](https://github.com/singerdmx/flutter-quill/blob/2c54d43f8b5cfe4a352b128f3891fc6c82f5ab58/lib/src/editor/raw_editor/raw_editor_state.dart#L267-L294) and in [`textEditingValue`](https://github.com/singerdmx/flutter-quill/blob/2c54d43f8b5cfe4a352b128f3891fc6c82f5ab58/lib/src/editor/raw_editor/raw_editor_state_selection_delegate_mixin.dart#L17-L30) the `toPlainText` method is called. However, this does not happen once time, literally every frame the `toPlainText` method is called (the same thing happens if we also change the position of the caret) rebuilding the entire `Document` text (i suppose this behavior is related with the `CursorCont`).

> [!NOTE]
> We set the cached value to `null` because, in certain scenarios like performing an undo/redo operation, the cached content becomes invalid and needs to be refreshed. 
>
> When does this update happen? When the `toPlainText` method is called, which, by the way, is invoked very frequently. This ensures that the cached value always reflects the most up-to-date content of the editor.
> 
> Additionally, to prevent the cache from falling out of sync with the changes made to the editor, we reset its value to `null` within the `compose()` method whenever a modification operation occurs (such as deleting, inserting, or replacing text). This ensures that, instead of holding onto a potentially outdated value, the cache is regenerated only when needed, maintaining consistency with the editor's current state.

**This is basically the change:**

```diff
/// Returns plain text representation of this document.
String toPlainText([
  Iterable<EmbedBuilder>? embedBuilders,
  EmbedBuilder? unknownEmbedBuilder,
]) {
- return 
-   _root.children
-      .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
-       .join();
+  _cachedPlainText ??=
+   _root.children
+       .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
+        .join();
+    return _cachedPlainText!;
}
```

## Related issues

- *Related #997*

## Type of Change

### DIFF
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -581,6 +581,7 @@ class Document {
         _root.childCount > 1) {
       _root.remove(node);
     }
+    _cachedPlainText = null;
   }
 
   bool isEmpty() {
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -174,7 +174,7 @@ class Document {
     final formatDelta = _rules.apply(RuleType.format, this, index,
         len: len, attribute: attribute);
     if (formatDelta.isNotEmpty) {
-      compose(formatDelta, ChangeSource.local, false);
+      compose(formatDelta, ChangeSource.local);
       delta = delta.compose(formatDelta);
     }
 
@@ -438,7 +438,7 @@ class Document {
   /// of this document.
   ///
   /// In case the [change] is invalid, behavior of this method is unspecified.
-  void compose(Delta delta, ChangeSource changeSource, [bool reloadCacheText = true]) {
+  void compose(Delta delta, ChangeSource changeSource) {
     assert(!documentChangeObserver.isClosed);
     delta.trim();
     assert(delta.isNotEmpty);
@@ -470,9 +470,7 @@ class Document {
       throw StateError('_delta compose failed');
     }
     assert(_delta == _root.toDelta(), 'Compose failed');
-    if(reloadCacheText) {
-      _cachedPlainText = null;
-    }
+    _cachedPlainText = null;
     final change = DocChange(originalDelta, delta, changeSource);
     documentChangeObserver.add(change);
     history.handleDocChange(change);
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -548,12 +548,10 @@ class Document {
   String toPlainText([
     Iterable<EmbedBuilder>? embedBuilders,
     EmbedBuilder? unknownEmbedBuilder,
-  ]) {
-    cachedPlainText ??= _root.children
-        .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
-        .join();
-    return cachedPlainText!;
-  }
+  ]) =>
+      cachedPlainText ??= _root.children
+          .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
+          .join();
 
   @visibleForTesting
   @internal
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -25,23 +25,25 @@ import 'style.dart';
 class Document {
   /// Creates new empty document.
   Document() : _delta = Delta()..insert('\n') {
-    _loadDocument(_delta);
+    loadDocument(_delta);
   }
 
   /// Creates new document from provided JSON `data`.
   Document.fromJson(List data) : _delta = _transform(Delta.fromJson(data)) {
-    _loadDocument(_delta);
+    loadDocument(_delta);
   }
 
   /// Creates new document from provided `delta`.
   Document.fromDelta(Delta delta) : _delta = delta {
-    _loadDocument(delta);
+    loadDocument(delta);
   }
 
   /// Stores the plain text content of the entire document in memory for quick access.
   ///
   /// It acts as a cache to avoid repeatedly extracting or generating the plain text.
-  String? _cachedPlainText;
+  @visibleForTesting
+  @internal
+  String? cachedPlainText;
 
   /// The root node of the document tree
   final Root _root = Root();
@@ -470,19 +472,19 @@ class Document {
       throw StateError('_delta compose failed');
     }
     assert(_delta == _root.toDelta(), 'Compose failed');
-    _cachedPlainText = null;
+    cachedPlainText = null;
     final change = DocChange(originalDelta, delta, changeSource);
     documentChangeObserver.add(change);
     history.handleDocChange(change);
   }
 
   HistoryChanged undo() {
-    _cachedPlainText = null;
+    cachedPlainText = null;
     return history.undo(this);
   }
 
   HistoryChanged redo() {
-    _cachedPlainText = null;
+    cachedPlainText = null;
     return history.redo(this);
   }
 
@@ -547,14 +549,15 @@ class Document {
     Iterable<EmbedBuilder>? embedBuilders,
     EmbedBuilder? unknownEmbedBuilder,
   ]) {
-    _cachedPlainText ??=
-      _root.children
-          .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
-          .join();
-    return _cachedPlainText!;
+    cachedPlainText ??= _root.children
+        .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
+        .join();
+    return cachedPlainText!;
   }
 
-  void _loadDocument(Delta doc) {
+  @visibleForTesting
+  @internal
+  void loadDocument(Delta doc) {
     if (doc.isEmpty) {
       throw ArgumentError.value(
           doc.toString(), 'Document Delta cannot be empty.');
@@ -581,7 +584,7 @@ class Document {
         _root.childCount > 1) {
       _root.remove(node);
     }
-    _cachedPlainText = null;
+    cachedPlainText = null;
   }
 
   bool isEmpty() {
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -172,7 +172,7 @@ class Document {
     final formatDelta = _rules.apply(RuleType.format, this, index,
         len: len, attribute: attribute);
     if (formatDelta.isNotEmpty) {
-      compose(formatDelta, ChangeSource.local);
+      compose(formatDelta, ChangeSource.local, false);
       delta = delta.compose(formatDelta);
     }
 
@@ -436,7 +436,7 @@ class Document {
   /// of this document.
   ///
   /// In case the [change] is invalid, behavior of this method is unspecified.
-  void compose(Delta delta, ChangeSource changeSource) {
+  void compose(Delta delta, ChangeSource changeSource, [bool reloadCacheText = true]) {
     assert(!documentChangeObserver.isClosed);
     delta.trim();
     assert(delta.isNotEmpty);
@@ -468,7 +468,9 @@ class Document {
       throw StateError('_delta compose failed');
     }
     assert(_delta == _root.toDelta(), 'Compose failed');
-    _cachedPlainText = null;
+    if(reloadCacheText) {
+      _cachedPlainText = null;
+    }
     final change = DocChange(originalDelta, delta, changeSource);
     documentChangeObserver.add(change);
     history.handleDocChange(change);
--- a/lib/src/document/document.dart
+++ b/lib/src/document/document.dart
@@ -472,19 +472,19 @@ class Document {
       throw StateError('_delta compose failed');
     }
     assert(_delta == _root.toDelta(), 'Compose failed');
-    cachedPlainText = null;
+    // cachedPlainText = null;
     final change = DocChange(originalDelta, delta, changeSource);
     documentChangeObserver.add(change);
     history.handleDocChange(change);
   }
 
   HistoryChanged undo() {
-    cachedPlainText = null;
+    // cachedPlainText = null;
     return history.undo(this);
   }
 
   HistoryChanged redo() {
-    cachedPlainText = null;
+    // cachedPlainText = null;
     return history.redo(this);
   }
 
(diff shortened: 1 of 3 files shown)


CatHood0:
Should i add also cache plain text param to `Rule` classes? This could avoid us to use `Document.fromDelta(...).toPlainText()`.

lib/src/document/document.dart
```diff
@@ -433,7 +436,7 @@ class Document {
   /// of this document.
   ///
   /// In case the [change] is invalid, behavior of this method is unspecified.
-  void compose(Delta delta, ChangeSource changeSource) {
+  void compose(Delta delta, ChangeSource changeSource, [bool reloadCacheText = true]) {
```
EchoEllet, line 439: I think it makes sense to keep booleans as named arguments. There is even a Dart analysis rule to enforce this.

In this context, it's not clear what the third argument is.

lib/src/document/document.dart
```diff
@@ -38,6 +38,9 @@ class Document {
     _loadDocument(delta);
   }
 
+  /// a var that contains the plain text of the entire document
+  String? _cachedPlainText;
```
EchoEllet, line 42:
```dart
/// The cached plain text of the document.
```

And to be honest, I'm not sure this comment adds anything.

lib/src/document/document.dart
CatHood0: I'll remove it.

EchoEllet (commented):
Seems good, but maybe need more testing to avoid regressions?

CatHood0:
**A small update:** this fix also improves performance when getting plain text from the `Rules` used by default within the editor (one of the ones that was causing a lot of problems was `AutoFormatMultipleLinksRule`, which when using `toPlainText()`, literally made the editor work unnecessarily hard).

EchoEllet:
Is this PR ready for merge?

CatHood0:
@EchoEllet ies, i've already tested both the performance improvement and whether it could cause problems when formatting, inserting, or deleting text. There have not been (as far as I've tested) any problems, and it works exactly the same (just doing the necessary caching)

I even tried adding 1500 paragraphs, and unlike before, when with so many paragraphs, it became extremely slow to select or move the selection to different parts, now it works fluidly even in deletions, insertions or formatting as we would expect from a text editor. Although obviously there are details, but these are more related to how we handle the rendering of the nodes.

This was the page I used to test the performance: https://www.lipsum.com/feed/html using a huge variety of paragraphs

EchoEllet:
Then, LGTM, I will merge it after I publish a new version soon.

EchoEllet:
Published `11.1.0`, we're ready to merge this PR.

lib/src/document/document.dart
```diff
@@ -538,10 +546,13 @@ class Document {
   String toPlainText([
     Iterable<EmbedBuilder>? embedBuilders,
     EmbedBuilder? unknownEmbedBuilder,
-  ]) =>
+  ]) {
+    _cachedPlainText ??=
```
EchoEllet, line 550: Is there a possibility for this cached plain text to become outdated or invalid? I see you're setting it to `null` in some cases.

EchoEllet: Also, the following could work:

```dart
String toPlainText([
    Iterable<EmbedBuilder>? embedBuilders,
    EmbedBuilder? unknownEmbedBuilder,
  ]) =>
      _cachedPlainText ??= _root.children
          .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
          .join();
```

CatHood0: > Is there a possibility for this cached plain text to become outdated or invalid? I see you're setting it to null in some cases.

I set it to null because, for example, if we perform an undo/redo, the cached content is already invalid and needs to be updated. When is it updated? When toPlainText is called, which actually happens a ridiculous number of times, which means we always keep this value updated with the latest content.

And, to prevent it from being out of sync with the changes applied to the editor, what we do is set it to null again in compose(). That is, if we delete, insert, or replace, we always reset its value and regenerate it.

CatHood0: > Also, the following could work.

You have a point. I'll make that change in a moment.

CatHood0:
Now that we think about it, maybe we should set it to null in other methods as well, just to rule out any errors. For now, we shouldn't merge this PR until we're 100% sure about everything.

lib/src/document/document.dart
EchoEllet: Sounds good!

EchoEllet:
This PR needs tests, I already added basic tests, but need permission to push the changes to the branch in your fork.

lib/src/document/document.dart
```diff
@@ -547,14 +549,15 @@ class Document {
...
-    return _cachedPlainText!;
+    cachedPlainText ??= _root.children
+        .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
+        .join();
+    return cachedPlainText!;
   }
 
-  void _loadDocument(Delta doc) {
+  @visibleForTesting
+  @internal
```
EchoEllet, line 559: The internal is needed here since clients can use this API in their tests outside of this package, where the `visibleForTesting` ensures that it can be used in tests only.

test/document/document_test.dart
```diff
@@ -182,4 +182,79 @@ void main() {
           reason: 'start of blank line');
     });
   });
+  group('cachedPlainText', () {
```
EchoEllet, line 185: @CatHood0 These tests were quickly written without much consideration, I'm unsure if we missed anything, however they should only test the changes you made, so if `redo` doesn't set `cachedPlainText` to `null`, a test will fail.

lib/src/document/document.dart
EchoEllet: > I set it to null because, for example, if we perform an undo/redo, the cached content is already invalid and needs to be updated. When is it updated? When `toPlainText` is called, which actually happens a ridiculous number of times, which means we always keep this value updated with the latest content.
> 
> And, to prevent it from being out of sync with the changes applied to the editor, what we do is set it to null again in `compose()`. That is, if we delete, insert, or replace, we always reset its value and regenerate it when required.

Please add this explanation somewhere to the PR description so others can understand this PR more quickly.

test/document/document_test.dart
CatHood0: I've already tried the tests, and they actually cover most cases where `cachedPlainText` could be affected. So far, there don't seem to be any errors. I'll see if I can come up with more cases where this variable could fail.

lib/src/document/document.dart
```diff
@@ -539,11 +549,13 @@ class Document {
     Iterable<EmbedBuilder>? embedBuilders,
     EmbedBuilder? unknownEmbedBuilder,
```
EchoEllet, line 550: Is there a need for adding a parameter in the `toPlainText` that ignores the cached plain text and always returns the result as before? If not, there is no need to add it, it will increase the complexity and require more tests.

CatHood0: Is there a need for adding a parameter in the `toPlainText` that ignores the cached plain text and always returns the result as before? If not, there is no need to add it, it will increase the complexity and require more tests.

CatHood0: Could you explain this a little more? I don't think i'm quite understanding.

CatHood0:
> I haven't experienced any noticeable issues while testing this branch. Do you expect any?

Not really. Like you, I took my time to make sure there were no issues with the editor's behavior (toolbar, keyboard shortcuts, common text operations, copy/cut/paste operations), and I didn't experience any problems. So I'm confident this PR won't cause any problems.

EchoEllet:
I will test it on all platforms soon. We could optionally test this branch on production apps (if they are open source).

lib/src/document/document.dart
EchoEllet: > I set it to null because, for example, if we perform an undo/redo, the cached content is already invalid and needs to be updated. When is it updated? When `toPlainText` is called, which actually happens a ridiculous number of times, which means we always keep this value updated with the latest content.
> 
> And, to prevent it from being out of sync with the changes applied to the editor, what we do is set it to null again in `compose()`. That is, if we delete, insert, or replace, we always reset its value and regenerate it when required.

Maybe in https://github.com/singerdmx/flutter-quill/pull/2482/files#diff-9f4e5dee6df0ec5a66471ba7aa0fc205d25389bffaba2e520cff6a3db1594016R41-R47 too

EchoEllet:
I can confirm that I didn't reproduce any noticeable issues in the editor on all platforms, including iOS and macOS.

EchoEllet:
To test this branch:

```yaml
dependency_overrides:
  flutter_quill:
    git:
      url: https://github.com/CatHood0/flutter-quill.git
      ref: cache_plain_text
```

singerdmx (approved):
LGTM

EchoEllet:
@CatHood0 BTW, this isn't a chore; it's a feature?

CatHood0:
I don't consider it a feature in itself. It's more of an addition that prevents you from having to perform certain processes multiple times unnecessarily. It's more like maintenance (in my opinion).

ZhuJHua:
Hello, thanks for your work! But i hava a question about this:
```dart
String toPlainText([
  Iterable<EmbedBuilder>? embedBuilders,
  EmbedBuilder? unknownEmbedBuilder,
]) =>
    cachedPlainText ??= _root.children
        .map((e) => e.toPlainText(embedBuilders, unknownEmbedBuilder))
        .join();
```
Does this mean that the parameter of embedBuilder is invalid? Because all `toPlainText` called inside quill do not carry these two parameters, the problem I am now facing is that the `toPlainText` of the embedBuilder I defined myself does not take effect.
I know that carrying parameters in toPlainText inside quill may cause wrong cursor position calculations, maybe we should add a parameter that can be used without cache?

ZhuJHua:
> > Does this mean that the parameter of embedBuilder is invalid? Because some toPlainText called inside quill do not carry these two parameters, the problem I am now facing is that the toPlainText of the embedBuilder I defined myself does not take effect.
> 
> Hmm, I don't think that's really the problem. This PR simply adds a variable to cache the plain INTERNAL text used by the editor and that the user doesn't need to see.
> 
> Maybe I'm misunderstanding you, as I don't understand what you mean by saying your embeds aren't having an "effect."
> 
> > I know that carrying parameters in toPlainText inside quill may cause wrong cursor position calculations, maybe we should add a parameter that can be used without cache?
> 
> I understand this can be confusing, as one might think that defining embedBuilder within the configurations will have some adverse effect on the editor, but no, internally, `embedBuilders` are never passed to `toPlainText`, which avoids the problem you're worried about.
> 
> We allow it to define `embedBuilders` in parameters, so that the plain text output is whatever you prefer (i.e., it's purely aesthetic; it will never have any real effect on the editor).

I have some EmbedBuilders for displaying custom components, and I have rewritten the toPlainText method among them
```dart
class AudioEmbedBuilder extends EmbedBuilder {
  final bool isEdit;

  AudioEmbedBuilder({required this.isEdit});

  @override
  String get key => AudioBlockEmbed.embedType;

  @override
  String toPlainText(Embed node) => '';

  @override
  Widget build(BuildContext context, EmbedContext embedContext) {
    final audioEmbed = AudioBlockEmbed(
      embedContext.node.value.data,
    ); 
    final path =
        isEdit
            ? FileUtil.getCachePath(audioEmbed.name)
            : FileUtil.getRealPath('audio', audioEmbed.name);

    return AudioPlayerComponent(path: path); 
  }
}
```
I set "toPlaintext" as an empty string to avoid characters like "OBJ" in the quillControllers `toPlaintext` method.
However, when I called the toPlaintext method, the OBJ character was still returned, and the toPlainText I defined for each embedBuilder did not take effect

ZhuJHua:
My conjecture is that quillController internally called toPlaintext without carrying parameters, resulting in the caching of incorrect results.Or is my usage method wrong? 
```dart
 quillController!.document
            .toPlainText(
              [
                ImageEmbedBuilder(isEdit: true),
                VideoEmbedBuilder(isEdit: true),
                AudioEmbedBuilder(isEdit: true),
                TextIndentEmbedBuilder(isEdit: true),
              ],
            )
            .trim();
```

CatHood0:
> I set "toPlaintext" as an empty string to avoid characters like "OBJ" in the quillControllers toPlaintext method.
However, when I called the toPlaintext method, the OBJ character was still returned, and the toPlainText I defined for each embedBuilder did not take effect

I get it now! I'll fix this as soon as i can. 

Thanks for taking the time to describe the issue 🌟.

IChat123:
> > I set "toPlaintext" as an empty string to avoid characters like "OBJ" in the quillControllers toPlaintext method.我将“toPlaintext”设置为空字符串，以避免在quillControllers toPlaintext方法中使用“OBJ”等字符。
> > However, when I called the toPlaintext method, the OBJ character was still returned, and the toPlainText I defined for each embedBuilder did not take effect但是，当我调用 toPlaintext 方法时，OBJ 字符仍然被返回，并且我为每个 embedBuilder 定义的 toPlainText 没有生效
> 
> I get it now! I'll fix this as soon as i can.我现在明白了！我会尽快解决这个问题。
> 
> Thanks for taking the time to describe the issue 🌟.感谢您花时间描述这个问题🌟。

Hello, may I ask if this issue has been resolved? Because I also encountered a cachedPlainText when rewriting the 'toPlainText' method, which caused my 'toPlainText' to not take effect.

PR #2484, opened 2025-02-20
Expose Rule type so that Document.setCustomRules can be used

tjarvstrand:
## Description

Exposes the `Rule` type through the public API. It's been around for a while and not seen a ton of changes. `Document` exposes the `setCustomRules`-method, but it is not strictly speaking callable through the public API since `Rule` and all its subtypes are only exposed through the internal API.

In our specific use case, we have a custom inline style that we would like to preserve the style of when typing in the middle of it. One option would be to expose the `PreserveInlineStylesRule` and make it more configurable. That would actually be more preferable for us but it seemed a lot more invasive, so I opted to try this route instead.

## Type of Change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Removed unicode from `QuillText` element that causes weird caret behavior on empty lines [#2453](https://github.com/singerdmx/flutter-quill/pull/2453).
 * Update QuillController `length` extension method deprecation message [#2483](https://github.com/singerdmx/flutter-quill/pull/2483).
 
+### Added
+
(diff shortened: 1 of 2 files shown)


CatHood0 (approved):

tjarvstrand:
@CatHood0 Would you be open to also expose the `PreserveInlineStylesRule` and make it parameterizable in terms of which keys it preserves?

EchoEllet (approved):

CatHood0:
> @CatHood0 Would you be open to also exposing the PreserveInlineStylesRule and make it parameterizable in terms of which keys it preserves?

@tjarvstrand The problem of parameterizing a single `Rule`, could perhaps cause all the others to have to be changed.

If you can explain to me a little more what you will do, I will be able to tell you if it will be viable or if you will have to change it.

Issue #2370, opened 2024-11-11
ShortcutEvents are not working on mobile platform.


hjkim-mango:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill version

10.8.5

### Steps to reproduce

1. Write '*' at first and then jot something down and write '*' at last.
2. Write '-' at first and then press the Spacebar.
3. Write '#' at first and then press the Spacebar.

### Expected results

On Step 1, characters should be changed to Italic style.
On Step 2, the block style of this line should be changed to the bulleted list.
On Step 3, the block style of this line should be changed to the numbered list.

### Actual results

Nothing are changed.

### Additional Context

<details>
<summary>Screenshots / Video demonstration</summary>

https://github.com/user-attachments/assets/3601d263-a235-45b5-90b6-e23d29095954

https://github.com/user-attachments/assets/69a5e360-34b3-4d5e-acf8-773d3e92cd96

</details>

<details><summary>Source Codes</summary>

```code
import 'package:flutter/material.dart';
import 'package:flutter_quill/flutter_quill.dart' as quill;

class FlutterQuillWidget extends StatefulWidget {
  const FlutterQuillWidget({super.key});

  @override
  State<FlutterQuillWidget> createState() => _FlutterQuillWidgetState();
}

class _FlutterQuillWidgetState extends State<FlutterQuillWidget> {
  late quill.QuillController _controller;
  final _editorFocusNode = FocusNode();
  final _editorScrollController = ScrollController();
  bool canScroll = false;

  @override
  void initState() {
    super.initState();
    _controller = quill.QuillController(
        document: quill.Document(),
        selection: const TextSelection.collapsed(offset: 0));
  }

  @override
  void dispose() {
    _controller.dispose();
    _editorFocusNode.dispose();
    _editorScrollController.dispose();
    super.dispose();
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
        resizeToAvoidBottomInset: false,
        appBar: AppBar(
          title: const Text('Flutter Quill'),
        ),
        body: Column(
          children: [
            Expanded(
                child: SingleChildScrollView(
                    controller: _editorScrollController,
                    child: Column(children: [
                      quill.QuillToolbar.simple(controller: _controller),
                      quill.QuillEditor(
                        controller: _controller,
                        scrollController: _editorScrollController,
                        focusNode: FocusNode(),
[...]

Issue #2450, opened 2025-01-22
Text highlighting/selection issue with Flutter 3.27.2


bloemy7:
### Have you checked for an existing issue?

- [x] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill Version

10.8.5

### Steps to Reproduce

1. Setup project with flutter-quill when running flutter version 3.27.2
2. Try highlighting text to select it

### Expected results

Expecting to be able to smoothly select text

### Actual results

Almost impossible to select text as it auto-select multiple lines in a very buggy way

### Additional Context

<details>
The issue seems to be around the precision of the cursor when selecting a line. If you put it exactly at the height of a line, it acts as if it was already slightly above the line, and so it automatically selects the full line. This makes the entire experience impossible to properly control.

I have downgraded Flutter to 3.27.1 and 3.27.0, but there when selecting text it doesn't show the resulting highlight as you are dragging, only once you release the drag, which is not a good experience either.

Any ideas?
</details>

EchoEllet:
> 10.8.5

Could you try with `11.0.0-dev.20` (or later) as the magnifier has been removed and related issues have been fixed.

bloemy7:
Hi @EchoEllet, just tried with 11.0.0-dev.21, and it's still happening for me unfortunately. 

https://github.com/user-attachments/assets/92c13c2c-5279-453e-82ed-7f320a73ab16

Any other ideas? Thanks!

jonasbernardo:
I tested it here, on Android the selection with a bluetooth mouse works normally, the problem only happens in the web version, another problem is also when clicking on empty lines, the cursor always appears on the top line and not on the clicked line, but it needs to be between two text when there is an empty space

jonasbernardo:
i use 11.0.0-dev.20

jonasbernardo:
These 2 problems are big, which makes the web version very difficult to use, it should be fixed urgently
@EchoEllet 
@CatHood0

bloemy7:
Yes that is a separate bug that is not related to flutter-quill at all, sorry for the confusion here. But it's a bug that forces us to update Flutter version, except that we cannot because of this bug @EchoEllet .

CatHood0:
I cannot reproduce this issue.

CatHood0:
I cannot reproduce this issue in Flutter 3.27.3 and the `11.0.0-dev.20`

bloemy7:
[...]

Issue #2445, opened 2025-01-18
The inserted text is not synced with the cursor location on Android Emulators


EchoEllet:
### Have you checked for an existing issue?

- [x] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill Version

10.8.5, 11.0.0-dev.19, 11.0.0-dev.20

### Steps to Reproduce

1. Run the Android example app on an emulator.
2. Type any text anywhere.
3. Change the cursor location.
4. Type any text again.

https://github.com/user-attachments/assets/b4e9c6cc-a767-4367-b7e4-da0ac76dd89f

Also, repeat the same steps on a real device and the issue won't be reproduced:

https://github.com/user-attachments/assets/fe19d24e-38e8-466f-8b59-1e1e3026a769

### Expected results

The inserted text should be at the correct location where the cursor is. The editor works as expected on a real device.

### Actual results

The inserted text is inserted at the incorrect location, changing the cursor location and then typing text will insert the text at a previous location. The editor doesn't work as expected on an emulator.

## Related Issues:

- #2413: At first, I assumed this was a regression after reverting the magnifier but [this is not the case](https://github.com/singerdmx/flutter-quill/pull/2413#issuecomment-2599766765).
- #2128 and #2026: Reverted by #2413 and still able to reproduce the issue, I'm not able to reproduce this issue on older versions even on an emulator.

Issue reproduced on the [`10.8.5`](https://pub.dev/packages/flutter_quill/versions/10.8.5), [11.0.0-dev.19](https://pub.dev/packages/flutter_quill/versions/11.0.0-dev.19), and the [PR #2413 while revert the magnifier](https://github.com/singerdmx/flutter-quill/pull/2413).

This is not a regression caused by v11 or #2413 (the magnifier revert).

https://github.com/user-attachments/assets/61be490e-6e29-4bbb-bf37-74d3a2d7dccd

https://github.com/user-attachments/assets/0fac0e48-aa89-4489-8450-f99cf9e64de3

CatHood0:
I'll check this too.

CatHood0:
I'm noticing that the class `TextInputClient`, exactly in `updateEditingValue`:

```dart
@override
void updateEditingValue(TextEditingValue value) {
  // our implementation
}
```

[...]

PR #2510, opened 2025-03-12
Chore(breaking changes): Replace TextInputClient with DeltaTextInputClient

CatHood0:
## Description

The previous implementation of applying changes in the `TextInputClient` worked well and did the job that was expected of it, but it was not very easy to read (not everyone investigates what the `Diff` class is), and sometimes it was somewhat complicated to modify (at least in my case), since it depended a lot on what was obtained from the `Diff` object.

This change is actually quite simple.

> [!WARNING]
> I've marked this change as potentially breaking, as we'll likely need to modify the testing API to now use TextEditingDelta values ​​if this PR is to be merged.

### Using the available `TextEditingDelta` classes

What we did was go from applying the changes directly to dividing the logic into different sections:

- [`ime`](https://github.com/singerdmx/flutter-quill/tree/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime): In this folder, we find the most common methods used to apply the `TextEditingDeltas` we obtained from [`updateEditingValueWithDeltas`](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/diff_services.dart#L5). Here, you can find methods such as [`onReplace`](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime/on_replace_method.dart), [`onInsert`](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime/on_insert.dart), [`onDelete`](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime/on_delete.dart), or [`onNonTextUpdate`](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime/on_non_update_text.dart), which have their own logic for each use case.

### Was this change really necessary?

Yes, since this allows us to obtain more granular changes and avoids having to use a method like getDiff, which, in the long run (or for long documents), ruins the user experience quite a bit when having to compare large changes that occur in the editor.

We went from applying the `Diff` class directly to the `replaceText`, to instead obtaining specific classes for each use case/type of change made, which makes the logic for applying them more independent.

### Some of the advantages of this change are:

- Better code readability.
- Better stability thanks of the granular information that `TextEditingDelta` gives to us.
- It's more maintainable in the long term.
- The logic for applying each change is now specific to each use case and is easier to modify. 
- Allows new contributors to enter the code and modify it without much fear of breaking something.

## Related Issues

* Fix: *#2370*

## Type of Change

- [x] ❌ **Breaking:** Alters existing functionality and requires updates.

## TODO

- [ ] check if works for all platforms (i'll need some help with apple OS)
- [x] fix buggy behavior on deletion operations in web browsers
- [x] fix buggy behavior in android (the caret is not synced with the `TextEditingValue` passed)


CatHood0:
Idk exactly why, but, when i use web browsers, this exceptions are throwed:

```console
DartError: Unsupported operation: Platform._operatingSystem
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 307:3                         throw_
dart-sdk/lib/_internal/js_dev_runtime/private/profile.dart 117:39                                   _operatingSystem

```

CatHood0:
If you're wondering why this PR suddenly has commits from other PRs, it was my mistake when trying to get the latest changes from master, using `git pull` instead of `git mer`ge.

lib/src/editor/raw_editor/input/ime/on_delete.dart
```diff
@@ -0,0 +1,32 @@
...
+
+Future<void> onDelete(
+  TextEditingDeltaDeletion deletion,
+  QuillController controller,
+) async {
+  final selection = controller.selection;
+  if (selection.isCollapsed) {
+    final start = deletion.deletedRange.start;
+    final length = deletion.deletedRange.end - start;
+    controller.replaceText(
```
CatHood0, line 12: Mmm... Probably i will need to check why this part does not work as expected in web

lib/src/editor/raw_editor/input/ime/on_non_update_text.dart
```diff
@@ -0,0 +1,31 @@
...
+  TextEditingDeltaNonTextUpdate nonTextUpdate,
+  QuillController controller,
+) async {
+  // update the selection on Windows
+  //
+  // when typing characters with CJK IME on Windows, a non-text update is sent
+  // with the selection range.
+  if (Platform.isWindows) {
+    if (nonTextUpdate.composing == TextRange.empty &&
+        nonTextUpdate.selection.isCollapsed) {
```
CatHood0, line 15: I'll need to change this since pass the same params to the `updateSelection` method into both conditions.

CatHood0:
Although it solves several problems for desktop and mobile devices, it seems to behave strangely on the web. I wonder what's causing it to behave this way, when in reality, we're doing exactly the same thing as before, but with a little extra code.

For example: 

- If we press `shift+arrow up` and then try to `delete` the text, it will only update the position (note that on the web, it's actually just sending a `TextEditingDeltaNonTextUpdate` instead of a `TextEditingDeltaDeletion`).
- If I try to go right after performing the previous action, it will now completely ignore that event.


https://github.com/user-attachments/assets/7e194447-e3a1-41d7-96e7-9d1965921218

lib/src/editor/raw_editor/input/ime/on_non_update_text.dart
```diff
@@ -0,0 +1,29 @@
...
+import '../../../../../flutter_quill.dart';
+
+Future<void> onNonTextUpdate(
+  TextEditingDeltaNonTextUpdate nonTextUpdate,
+  QuillController controller,
+) async {
+  // when typing characters with CJK IME on Windows, a non-text update is sent
+  // with the selection range.
+  if (Platform.isWindows) {
+    if (nonTextUpdate.composing == TextRange.empty &&
```
CatHood0, line 12: We shouldn't need to check the platform, since the hardware keyboard can also be used on other devices, and this would cause us to completely ignore this event.

lib/src/editor/raw_editor/input/formatters/text_editing_delta_formatters.dart
```diff
@@ -0,0 +1,130 @@
...
+
+const String _whitespace = ' ';
+const int _whitespaceLen = _whitespace.length;
+
+// Extension on TextEditingDelta to provide a generic formatting method.
+// This method checks the type of the TextEditingDelta and calls the appropriate
+// formatting method for the specific delta type (insertion, deletion, replacement, or non-text update).
+// If the delta type is not recognized, it throws an UnimplementedError.
+@internal
+@experimental
```
CatHood0, line 13: I need to improve the documentation of these extensions.

example/pubspec.lock
```diff
@@ -602,18 +602,18 @@ packages:
...
   test_api:
     dependency: transitive
     description:
       name: test_api
-      sha256: fb31f383e2ee25fbbfe06b40fe21e1e458d14080e3c67e7ba0acfde4df4e0bbd
+      sha256: "664d3a9a64782fcdeb83ce9c6b39e78fd2971d4e37827b9b06c3aa1edc5e760c"
       url: "https://pub.dev"
     source: hosted
-    version: "0.7.4"
+    version: "0.7.3"
```
CatHood0, line 616: Idk why this change was included, since i didn't change anything on the example.

CatHood0:
I've been testing the change made in #2505, and it seems that when you press Shift+Down Arrow in a header, it suddenly selects more text than it should. I can't say exactly why, but it seems like the change made in that PR needs to be tested, as it could impact the experience for users who aren't using any Chinese IME.

example/pubspec.lock
```diff
@@ -373,18 +373,18 @@ packages:
...
+    version: "10.0.7"
   leak_tracker_flutter_testing:
     dependency: transitive
     description:
       name: leak_tracker_flutter_testing
-      sha256: f8b613e7e6a13ec79cfdc0e97638fddb3ab848452eff057653abd3edba760573
+      sha256: "9491a714cca3667b60b5c420da8217e6de0d1ba7a5ec322fab01758f6998f379"
       url: "https://pub.dev"
     source: hosted
-    version: "3.0.9"
```
EchoEllet, line 387: Are you using the latest stable Flutter SDK?

lib/src/editor/raw_editor/input/ime/on_replace_method.dart
```diff
@@ -0,0 +1,86 @@
...
+    if (textReplacement.length == 1) {
+      for (final shortcutEvent in characterShortcutEvents) {
+        if (shortcutEvent.character == textReplacement &&
+            shortcutEvent.handler(controller)) {
+          return;
+        }
+      }
+    }
+
+    if (Platform.isIOS) {
```
EchoEllet, line 29: This wouldn't work on web, use `isIOSApp` or `isIOS` (which is `true` when on iOS even on a browser) instead.

EchoEllet:
> Idk exactly why, but, when i use web browsers, this exceptions are throwed:

That's because you're using `dart:io` on the web, which is not supported.

lib/src/editor/raw_editor/input/formatters/text_editing_delta_formatters.dart
```diff
@@ -0,0 +1,130 @@
...
+        selection: selection << _whitespaceLen,
+        composing: composing << _whitespaceLen,
+      );
+}
+
+// Extension on TextEditingDeltaDeletion to format deletion deltas.
+// Adjusts the oldText, deletedRange, selection, and composing properties
+// by shifting them based on a predefined whitespace length.
+@internal
+@experimental
```
EchoEllet, line 48: We should not introduce experimental APIs so often that we have many that are still experimental, and it's usually not a common practice in the Dart ecosystem. 

We should only focus on getting APIs stable and only merge and release them when they are stable.

example/pubspec.lock
EchoEllet: Your IDE is running `flutter pub get` with the example automatically, and I think your SDK is outdated, which will use older versions of some dependencies. I think we should exclude this file from the repo, however, that will cause the CI to use different versions (depending on the SDK version), and it will use the latest compatible versions by default.

Without `pubspec.lock`, we sometimes get unexpected issues (which happened last year with InAppWebview breaking the build), it also ensure all contributor use the same version. We could also exclude the file and lock the versions (without the caret in `pubspec.yaml`).

CatHood0: It's actually strange, since I'm using the latest version of Flutter (as far as i know, it's 3.29.0). However, the first changes made in this PR were in 3.24.

lib/src/editor/raw_editor/input/formatters/text_editing_delta_formatters.dart
CatHood0: I've already removed this API.

To put it into context, i removed it because i did some testing to see if the changes were being applied correctly even without the API, and it seems that they were. As such, this API wasn't really necessary (i can say it even caused unexpected problems in web browsers).

CatHood0:
I'm going to check why `bug_fix_test` is suddenly failing (when I'm actually using the selection that comes from the new `TextEditingValue` which contains the correct values ​​to make the changes in the `QuillController` is `replaceText`.

CatHood0:
> That's because you're using dart:io on the web, which is not supported. We prefer not to import dart:io at all or use conditional imports if necessary (which is error-prone and will need Dart analysis rule) to get a full score on pub.dev.

Thanks for the clarification! I really didn't know this. That explains why some of my projects had problems with the web.

lib/src/editor/raw_editor/input/formatters/text_editing_delta_formatters.dart
EchoEllet: The context is with the usage of experimental usage in general and not this particular API.

I previously stated to introduce APIs as experimental first for a while but then stable, but now I think we need more consideration (a separate discussion) or completely avoid using experimental annotation like most packages in the Dart ecosystem.

I probably was wrong about this.

EchoEllet:
> Better code readability

I think clients prefer stability over readability. 

> Allows new contributors to enter the code and modify it without much fear of breaking something

This should be covered with tests instead. If you want to work on the tests and there are any blockers, please let me know.

Also, we should be responsible for merging the PR and checking it properly, not the contributors.

I'm not very confident about the merge of this PR. That's a lot to cover and check (this and the last few PRs, including the incoming ones) and I'm not able to do much for this period of time, as such, I can't really maintain the project with new PRs so I will leave it to you.

Does this PR impact the functionality or aim to change it?

example/pubspec.lock
CatHood0: I have no idea how, but it fixed itself.

CatHood0:
I'll take some time to create some tests for this PR.

CatHood0:
I don't know exactly why, but in web browsers, when you press `shift+arrow up`, events gradually accumulate unnecessarily. That is, the first time, it selects an entire line, then two full lines, and so on until you reach the beginning of the document just by pressing `shift+arrow up` (the action to reproduce is simple: `shift+arrow up`, then move the cursor to deselect, and repeat the action one more time. With that, you should be able to reproduce this).

I can conclude that, in some way, events are piling up.

https://github.com/user-attachments/assets/63c88b4a-8bfb-41ec-b8b8-000625778e57

CatHood0:
## New major issue that we need to fix

As I've suspected for some time, it seems the cursor out-of-sync issue stems from a Flutter issue. Now that it successfully updated to 3.29.0, this issue has returned on mobile devices. 

For my tests, I used versions prior to 3.27 and 3.29, and it appears this issue was created in one of these versions. I specifically used 3.24, which had worked fine for me up until now.

## Clarification of where is the issue

For some reason, in these two new versions on Android, when Flutter calls `updateEditingValue` on the `TextInputClient`, it returns a value that's not synchronized with the current state, causing the cursor to appear and perform actions we didn't tell it to.

This didn't happen in versions prior to 3.27, so it looks like we'll need to review and report this issue if it hasn't already been done.

CatHood0:
I don't know if it's worth mentioning, but before updating to Flutter 3.29.1, while testing how this PR worked on mobile devices, the cursor out-of-sync issue suddenly returned. No matter what i did, i thought it was back.

Once i updated, i retested the example app, and to my relief, it finally disappeared. It's strange, since when i uploaded this PR, i tested it on mobile devices (on version 3.29.0) and it was working fine, and then suddenly, until a few minutes ago, the issue appeared and disappeared. I wonder what's causing this issue.

I tried looking at the Flutter SDK Release Notes (https://docs.flutter.dev/release/release-notes), but they don't mention fixing anything similar. This is confusing, to be honest.

lib/src/editor/raw_editor/input/diff_services.dart
```diff
@@ -0,0 +1,56 @@
...
+TextEditingDelta getTextEditingDelta(
+  TextEditingValue oldValue,
+  TextEditingValue newValue,
+) {
+  // we need to check why sometimes in android, when we place the caret
+  // at a position, it moves backward unexpectly. By now, i think that we need to use
+  // the removed Debounce class to wait for the android soft-keyboard events
+  // since on android, non-text-update is called more times that we think
+  final currentText = oldValue.text;
+  final diff = getDiff(
```
CatHood0, line 14: I'll remove this comment since it's already fixed.

lib/src/editor/raw_editor/input/diff_services.dart
```diff
@@ -0,0 +1,56 @@
+import 'package:flutter/services.dart';
+import '../../../delta/delta_diff.dart';
+
+/// Return a list of the change type that was do it to the content of the editor
+TextEditingDelta getTextEditingDelta(
+  TextEditingValue oldValue,
+  TextEditingValue newValue,
+) {
+  // we need to check why sometimes in android, when we place the caret
```
CatHood0, line 9: I'll remove this comment since it's already fixed.

EchoEllet:
Is this issue the same as #2445 ?

CatHood0:
Yes, they're closely related since they reproduce in the same way.

The last time I tested using a mobile device, the error didn't appear and it worked as expected.

_If you notice that typing on Android is a bit slow, it's due to a change i made locally for testing purposes. That change isn't included in this PR._


https://github.com/user-attachments/assets/921a5d39-9fa9-4c12-a9c9-777774241109

lib/src/editor/raw_editor/input/ime/on_insert.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:flutter/services.dart';
+
+import '../../../../controller/quill_controller.dart';
+import '../../../raw_editor/config/events/character_shortcuts_events.dart';
+
+void onInsert(
```
CatHood0, line 6: we need to mark these functions as @internal and give them some documentation

lib/src/editor/raw_editor/input/ime/on_delete.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:flutter/services.dart';
+import '../../../../../flutter_quill.dart';
+
+void onDelete(
```
CatHood0, line 4: > we need to mark these functions as @internal and give them some documentation

Here too.

lib/src/editor/raw_editor/input/ime/on_non_update_text.dart
```diff
@@ -0,0 +1,26 @@
+import 'package:flutter/services.dart';
+import '../../../../../flutter_quill.dart';
+import '../../../../../internal.dart';
+
+void onNonTextUpdate(
```
CatHood0, line 5: > we need to mark these functions as @internal and give them some documentation

Here.

lib/src/editor/raw_editor/input/diff_services.dart
```diff
@@ -0,0 +1,56 @@
...
+  // at a position, it moves backward unexpectly. By now, i think that we need to use
+  // the removed Debounce class to wait for the android soft-keyboard events
+  // since on android, non-text-update is called more times that we think
```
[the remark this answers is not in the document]
CatHood0: No, actually, this has already been fixed. It was in a previous phase of the PR where, I don't even know why, the cursor would unexpectedly move to other positions when I wasn't touching the screen (or the editor).

This has since been fixed (I don't know what it was exactly, but some change I made in onNonTextUpdate was what finally fixed it). I've marked it in the comments so I don't forget to remove it.

lib/src/editor/raw_editor/input/ime/on_insert.dart
CatHood0: That would be another solution! I'll keep that in mind.

lib/src/editor/raw_editor/input/diff_services.dart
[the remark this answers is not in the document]
EchoEllet: This is an outdated review comment that I sent now. Thanks for working on it.

CatHood0:
I've been investigating a few things. And one of the most interesting is how [super_editor](https://github.com/superlistapp/super_editor) manages to listen for keyboard events regardless of the platform. And, without needing to use text strings, it's able to apply the changes, since it's based on the same `TextEditingDelta`.

I'm deciding whether we should find another implementation for this aspect, which would eliminate this disadvantage of having to differentiate changes from one string to another, when we could base it on specific positions and specific changes.

For now, I'll continue analyzing how [super_editor](https://github.com/superlistapp/super_editor) manages to do this. It's quite interesting, so for anyone interested in merging this PR, I'm sorry to say it will take even longer.

Possibly related to: #2517

CatHood0:
@EchoEllet do you know that could be happening here?

I'm just using `ctrl+v` (paste text that was copied before from the editor) several times.

```bash
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: PathNotFoundException: Cannot delete file, path = '/tmp/quill_native_bridge_linux/xclip' (OS
 Error: No existe el fichero o el directorio, errno = 2)
#0      _checkForErrorResponse (dart:io/common.dart:58:9)
#1      _File._delete.<anonymous closure> (dart:io/file_impl.dart:352:9)
<asynchronous suspension>
#2      QuillNativeBridgeLinux.getClipboardImage (package:quill_native_bridge_linux/quill_native_bridge_linux.dart:200:7)
<asynchronous suspension>
#3      DefaultClipboardService.getImageFile (package:flutter_quill/src/editor_toolbar_controller_shared/clipboard/default_clipboard_service.dart:28:12)
<asynchronous suspension>
#4      QuillController.clipboardPaste (package:flutter_quill/src/controller/quill_controller.dart:589:26)
<asynchronous suspension>
#5      QuillRawEditorState.pasteText (package:flutter_quill/src/editor/raw_editor/raw_editor_state.dart:145:9)
<asynchronous suspension>
```

CatHood0:
Yeah, i'm noticing that these changes breaks the flutter_quill_test API. I'll fix it.

CatHood0:
This https://github.com/singerdmx/flutter-quill/pull/2510#issuecomment-2719891365 probably is related with #2450?

CatHood0:
> Is this issue the same as https://github.com/singerdmx/flutter-quill/issues/2445 ?

I've managed to reproduce the problem. Now, if you select text, it won't appear. You have to do a little hack.

Steps:

- Copy the entire contents of the example.
- Position yourself anywhere in the editor.
- Paste the text multiple times (even if your phone gets stuck)
- Select anywhere in the editor and quickly type something.

This may not be easy to reproduce. But it's as simple as not giving the editor time to process what's happening.

It stops happening if you first select, wait a moment, and then type.

I suspect this is because the `getDiff` is running in one place, and suddenly, an update from the `DeltaTextInputClient`/`TextInputClient` is also executed. These conflict. The input client runs first, but once the `getDiff` is finished, it doesn't know that it no longer needs to update the selection, and then this happens.

That is, it is an event error that occurs if they are executed at the same time, causing them to collide, since both have their own values, but they do not know which should be executed first and which should not.

EchoEllet:
> @EchoEllet what do you think?

I'm not really related to the topic right now, and will not work on the project for a while.

CatHood0:
I'll start updating these PRs and finishing them. 

I've been a bit busy, and I'll likely have less time in the near future for certain personal reasons. Still, I'll continue to help however I can.

EchoEllet:
> Do you know what could be happening here?

I missed the notification. This is probably a bug. The plugin should remove this file if it hasn't already been removed. Could you share details including the desktop environment and the distro, and whether if you're using any sandboxing to run the app (e.g. Flatpak). Please consider filing an issue to [quill-native-bridge](https://github.com/FlutterQuill/quill-native-bridge/issues) repo if you have the time. Unfortunately, I'm not able to work on the project for this period of time.

lib/src/editor/raw_editor/raw_editor_state.dart
```diff
@@ -486,7 +486,6 @@ class QuillRawEditorState extends EditorState
         child: EditorKeyboardShortcuts(
           actions: _shortcutActionsManager.actions,
           onKeyPressed: widget.config.onKeyPressed,
-          characterEvents: widget.config.characterShortcutEvents,
```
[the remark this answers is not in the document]
CatHood0: Please check out [on_insert](https://github.com/singerdmx/flutter-quill/blob/aad0b56ef97ce01fd07e59fef8ff9b640b6fe372/lib/src/editor/raw_editor/input/ime/on_insert.dart). 

I've moved them to work directly with `DeltaTextInputClient` events. This allows all platforms to work with `CharacterShortcutEvent` without having to create native implementations or widgets that aren't really needed.

CatHood0:
> Could you share details including the desktop environment and the distro, and whether if you're using any sandboxing to run the app (e.g. Flatpak)

I'm using Debian 12 with the Kernel 6.12.12. And I'm just running the app directly using `flutter run -d linux` in the terminal

> Please consider filing an issue to [quill-native-bridge](https://github.com/FlutterQuill/quill-native-bridge/issues) repo if you have the time. Unfortunately, I'm not able to work on the project for this period of time. This [line could be related](https://github.com/FlutterQuill/quill-native-bridge/blob/48b7237873fe1f1f8d492b123b16dfddf28268d2/quill_native_bridge_linux/lib/quill_native_bridge_linux.dart#L98).

Let me reproduce the problem again and add more details to the issue.

And don't worry about the difficulty over time issue. I understand. I'm exactly the same right now.

CatHood0:
> Is there anything I can do to get it fixed? I'm having a lot of bugs due to this problem and I'm available to help

It's not much at the moment.

We're just having issues with how `TextSelectionDelegate` and `TextInputClient` communicate, which (still theoretical) seem to be colliding when certain events occur too quickly, causing the issue with the cursor not updating.

Also, on the Web, there seem to be some issues when moving the cursor up or down using `ctrl+arrow up/down`, which seem to be accumulating (something not from our side).

This last issue is something Flutter is currently having (the same thing is happening with text selection). I was waiting for Flutter to release a new update to check if they had fixed this issue, but nothing has happened yet.

EchoEllet:
> i'll need some help with apple OS

Could you provide some steps on how you're testing this change so I can reproduce on macOS and iOS?

Issue #2374, opened 2024-11-12
Continous jank in text with 1k rows


vargab95:
### Is there an existing issue for this?

- [X] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill version

master branch

### Steps to reproduce

1. Clone the repo
2. Go to the example folder
3. Start the example
4. Open the performance tab of flutter dev tools
5. Delete the example text
6. Paste example + enter 1k times
7. Click in a row and add one character

### Expected results

A document with few thousand rows renders at least with 30 FPS when edited.

### Actual results

I've tested with a linux desktop PC (AMD Ryzen 5 5600G). With 1k rows, it rendered in 30-60ms, with 2k rows in 60-80ms, with 4k rows in 100-120ms, with 8k it took 200-250ms. It is more or less exactly O(n). Which means that even when a user edits a document with 1k rows, it will generate a jank frame.

### Additional Context

I think, having documents with 1k rows are not so rare in case of a rich text editor.

I've searched through the issues and I've found that in https://github.com/singerdmx/flutter-quill/issues/997 it was already described. However, on master I cannot see the rasterization charts, but jank frames are still problematic.

EchoEllet:
> master branch

Have you tried older/other major versions of the project? If yes, can you mention them to ensure if this is a regression.

vargab95:
I've started a pixel 5 emulator and using that even editing the original example text generates jank frames. When I hit a key on the virtual keyboard, it results in a jank frame with 40-70ms duration.

EchoEllet:
> some older versions

Can you also test the latest pre-release? If the [migration](https://github.com/singerdmx/flutter-quill/blob/master/doc/migration/10_to_11.md) is too much to test, maybe with the example of the `flutter_quill`?

vargab95:
Sure, with the flutter_quill example it's still the same 30-60ms with 1k rows when using v11.0.0-dev.12.

vargab95:
I've enabled tracking widget builds in all code. The result is that a single render of QuillRawEditorMultiChildRenderObject takes 59.5ms.

![60ms_render](https://github.com/user-attachments/assets/6f73be07-496a-45a4-8ee8-6f37ad0482fc)

And another screenshot, where you can see one of the 1k TextLine objects.

![60ms_render_zoomed_in](https://github.com/user-attachments/assets/a967bafb-65e5-41eb-93b3-572864333c86)

[...]

PR #2517, opened 2025-03-18
Chore: improve getDiff perfomance for the editor

CatHood0:
## Description

Improved the performance of the `getDiff` method, which calculates the differences between two texts (`oldText` and `newText`). The main focus is to optimize the code to handle extremely long strings (e.g., 24,000 words or more) without affecting the editor's responsiveness.

> [!NOTE]
> The videos for preview how this PR will be added as soon as i can fix a little part of code that degrades the perfomance.

## Related Issues

* Related: *#2523*
* Partially fix: *#2374*
* Partially fix: *#997*

## Type of Change


CatHood0:
I don't know whether to actually close this PR or not. It improves getDiff's performance, but only slightly. With large strings (I tested with 20,000 words), the performance difference is usually very small (less than 200 ms).

EchoEllet:
> I don't know whether to actually close this PR or not.

Let's focus on the previous PRs to discuss the testing situation issues.

CatHood0:
This PR is ready to be tested for anyone. Please, if you detect that something is not working, report it here.

test/editor/get_diff_test.dart
```diff
@@ -0,0 +1,122 @@
...
+void main() {
+  group('Performance Tests', () {
+    late Stopwatch stopwatch;
+    const loremIpsum =
+        'Lorem ipsum dolor sit amet, consectetur adipiscing elit. '
+        'Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. ';
+
+    setUp(() => stopwatch = Stopwatch());
+
+    test('Simple insert (should complete <10ms)', () {
```
EchoEllet: Usually, test names in Dart start with a lowercase character for common conventions, [example](https://github.com/flutter/packages/blob/main/packages/image_picker/image_picker/test/image_picker_test.dart#L38). I will close this as solved since this is not a priority, but I usually maintain consistency as long as it doesn't take and block development.

CatHood0: Thanks for tell me this. I'll change it!

EchoEllet: It's not a priority at all. I hope we agree on the same thing on future PRs.

EchoEllet:
> This PR is ready to be tested for anyone.

Please provide them with minimal steps to test your branch:

```yaml
dependency_overrides:
  flutter_quill:
    git:
      url: https://github.com/CatHood0/flutter-quill.git
      ref: improve_diff
```

CatHood0:
> BTW, I remember you telling this doesn't make a performance different. is it different now?

Yes, that was because it only improved the previous implementation, but didn't change its foundation. The problem with the previous one is that it didn't limit the ranges for searching for changes. That is, with the previous getDiff, with a text of 50,000 words, if we inserted only one character at any position, it would unnecessarily go through the ENTIRE STRING, only to realize that only one character had been inserted.

Unlike that, the new `getDiff` uses the previous selection (the one in the `QuillController`, which hasn't been updated) and the new one (which comes from the `TextEditingValue` provided by `updateEditingValue`) as range limiters. By default (and it's better), the range for searching for changes is only limited to 20 characters from the left and right. However, when the selection isn't collapsed, we use its values ​​to correctly calculate the ranges we need to check without losing what we're looking for: limiting the search for changes to places where there aren't any.

By limiting the range, performance improves significantly, since it doesn't have to loop all the way to the end. Of course, always keep in mind that all this is relative to the cursor position.

EchoEllet:
The video has been removed. Could you please consider re-adding it with before vs after in the PR description?

CatHood0:
Since it's likely going to take me longer to improve, I'll close this PR. I'll reopen it once I'm 100% certain this significantly improves performance.

There is strange: i tested this change before, and the diff in perfomance was significant, but suddenly there's no difference at all. I sound schizophrenic.

lib/src/delta/delta_diff.dart
```diff
@@ -31,30 +57,222 @@ class Diff {
...
+  // 2. Adjust bounds for length variations
+  final oldLen = oldStr.length;
+  final newLen = newStr.length;
+  final lengthDiff = newLen - oldLen;
+
+  // 3. Forward search from range start
+  var hasForwardChange = false;
+
+  while (start < end && start < oldStr.length && start < newStr.length) {
+    if (oldStr[start] != newStr[start]) {
```
CatHood0, line 91: This operation is too heavy. We need to replace with a more smooth and efficent alternative.

lib/src/delta/delta_diff.dart
```diff
@@ -31,30 +57,222 @@ class Diff {
...
+  }
+
+  // 4. Backward search from range end
+  var oldEnd = math.min(end, oldLen);
+  var newEnd = math.min(end + lengthDiff, newLen);
+
+  var hasBackwardChange = false;
+
+  while (oldEnd > start && newEnd > start) {
+    if (oldStr[oldEnd - 1] != newStr[newEnd - 1]) {
```
CatHood0, line 109: > This operation is too heavy. We need to replace with a more smooth and efficent alternative.

This too.

CatHood0:
@EchoEllet check this:

https://github.com/user-attachments/assets/42abe50b-8dde-405f-8eb4-f69d85d98920

CatHood0:
Since #2510 is working on avoid use of `getDiff`, i thinking on close this.

PR #2488, opened 2025-02-25
fix unpredicatble endless loop of '_handleFocusChanged' in the hase of page route changingwhen editor is set to readonly

chaosue:
## Description
If the property `readOnly` is set to true, typically in the read mode of contents, the UI tread may be stuck by endless loop when the route push back to the page with quill.  The direct cause comes from the `_handleFocusChanged` listener, the `focusNode` will keep losing focus and requesting focus again instantly. 
The root cause hasn't been digged out,  removing the unnesessary listeners in `readOnly` mode can resolve the issue at the moment.

## Type of Change
- [ ] ✨ **Feature:** New functionality without breaking existing features.
- [x ] 🛠️ **Bug fix:** Resolves an issue without altering current behavior.
- [ ] 🧹 **Refactor:** Code reorganization, no behavior change.
- [ ] ❌ **Breaking:** Alters existing functionality and requires updates.
- [ ] 🧪 **Tests:** New or modified tests
- [ ] 📝 **Documentation:** Updates or additions to documentation.
- [ ] 🗑️ **Chore:** Routine tasks, or maintenance.
- [ ] ✅ **Build configuration change:** Build/configuration changes.

PR #2529, opened 2025-04-02
Fix selection in readOnly mode, Add magnifier via RawMagnifier widget

### LINKED ISSUE #1504: Text magnification feature for the Quill Editor
I'm making a note-taking app using the `flutter_quill` package. While using the editor, I noticed that dragging the cursor cannot magnify the text.

Since I’m aware that the `TextMagnifierConfiguration` class is available for the `TextField` widget and text magnification is enabled by default, I thought there might be something similar in the `QuillEditorConfigurations` constructor but could not find any.

Is this feature already available or is there a workaround to make it work? It’d be more convenient for users to edit their notes by moving the cursor using the magnifier.

### LINKED ISSUE #2518: Can't select text when readOnly to true
### Have you checked for an existing issue?

- [x] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill Version

11.0.1

### Steps to Reproduce

Create a readonly QuillEditor like so:
```dart
class _ShowTextState extends State<ShowText> {
  late final QuillController _controller = QuillController(
      document: Document.fromDelta(widget.doc),
      readOnly: true,
      selection: const TextSelection.collapsed(offset: 0));

  @override
    return Container(
      constraints: const BoxConstraints(minHeight: 200),
      decoration: BoxDecoration(
        border: Border.all(color: context.theme.colorScheme.outlineVariant),
        borderRadius: BorderRadius.circular(32),
      ),
      width: double.maxFinite,
      child: QuillEditor(
        scrollController: ScrollController(),
        config: const QuillEditorConfig(
          showCursor: false,
          autoFocus: false,
        ),
        focusNode: FocusNode(),
        controller: _controller,
      ),
    );
  }
}
```

### Expected results

I expect the text to be selectable even if it's on readOnly mode.

### Actual results

https://github.com/user-attachments/assets/c1f91781-d332-4068-8bdd-a444a8d03fbe

### Additional Context

<details>
<summary>
Not working on web and Android, not tested in IOS.

The text is not selectable when readOnly is set to true.
</summary>

https://github.com/user-attachments/assets/467721b9-3f44-4780-9600-ebb76bd8674a

</details>

<details>

</details>

mtallenca:
## Description

1) Allow selection in read only mode
2) Displays magnifier when dragging on iOS/Android

## Related Issues

- *Fix #2518*
- *Fix #1504*

## Type of Change

### DIFF
--- a/lib/src/editor/raw_editor/raw_editor.dart
+++ b/lib/src/editor/raw_editor/raw_editor.dart
@@ -26,6 +26,7 @@ class QuillRawEditor extends StatefulWidget {
   final QuillController controller;
   final QuillRawEditorConfig config;
 
+  /// {@template drag_offset_notifier}
   /// dragOffsetNotifier - Only used on iOS and Android
   ///
   /// [QuillRawEditor] contains a gesture detector [EditorTextSelectionGestureDetector]
@@ -44,6 +45,7 @@ class QuillRawEditor extends StatefulWidget {
   /// the correct location (or hide the magnifier if null). [EditorTextSelectionOverlay] will
   /// use the value of the dragOffsetNotifier to hide the context menu when the magnifier is
   /// displayed and show the context menu when dragging is complete.
+  /// {@endtemplate}
   final ValueNotifier<Offset?>? dragOffsetNotifier;
 
   @override
--- a/lib/src/editor/raw_editor/raw_editor.dart
+++ b/lib/src/editor/raw_editor/raw_editor.dart
@@ -25,6 +25,25 @@ class QuillRawEditor extends StatefulWidget {
 
   final QuillController controller;
   final QuillRawEditorConfig config;
+
+  /// dragOffsetNotifier - Only used on iOS and Android
+  ///
+  /// [QuillRawEditor] contains a gesture detector [EditorTextSelectionGestureDetector]
+  /// within it's widget tree that includes a [RawMagnifier]. The RawMagnifier needs
+  /// the current position of selection drag events in order to display the magnifier
+  /// in the correct location. Setting the position to null will hide the magnifier.
+  ///
+  /// Initial selection events are posted by [EditorTextSelectionGestureDetector]. Once
+  /// a selection has been created, dragging the selection handles happens in
+  /// [EditorTextSelectionOverlay].
+  ///
+  /// Both [EditorTextSelectionGestureDetector] and [EditorTextSelectionOverlay] will update
+  /// the value of the dragOffsetNotifier.
+  ///
+  /// The [EditorTextSelectionGestureDetector] will use the value to display the magnifier in
+  /// the correct location (or hide the magnifier if null). [EditorTextSelectionOverlay] will
+  /// use the value of the dragOffsetNotifier to hide the context menu when the magnifier is
+  /// displayed and show the context menu when dragging is complete.
   final ValueNotifier<Offset?>? dragOffsetNotifier;
 
   @override
--- a/lib/src/editor/raw_editor/raw_editor.dart
+++ b/lib/src/editor/raw_editor/raw_editor.dart
@@ -11,6 +11,7 @@ class QuillRawEditor extends StatefulWidget {
   QuillRawEditor({
     required this.config,
     required this.controller,
+    this.dragOffsetNotifier,
     super.key,
   })  : assert(config.maxHeight == null || config.maxHeight! > 0,
             'maxHeight cannot be null'),
@@ -24,6 +25,7 @@ class QuillRawEditor extends StatefulWidget {
 
   final QuillController controller;
   final QuillRawEditorConfig config;
+  ValueNotifier<Offset?>? dragOffsetNotifier;
 
   @override
   State<StatefulWidget> createState() => QuillRawEditorState();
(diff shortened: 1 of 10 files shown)


CatHood0:
From what I see, this restores the magnifier behavior, right?

We've had issues with this implementation before, and it led to some problems during rollbacks.

For the reasons mentioned above, it's best to test this PR until we know it won't cause us problems, like it did last time.

In any case, thank you for taking the time and effort to make these changes.

mtallenca:
This does restore magnifier behavior.  I did not create the previous magnifier implementation, but I did fix some of it's bugs that had been merged. However, this is a completely different implementation that is using Flutter's [RawMagnifier](https://api.flutter.dev/flutter/widgets/RawMagnifier-class.html) widget in the Widget tree (see text_selection.dart) instead of changing events / config / etc throughout the editor.  The other dart changes in this PR are just passing a ValueNotifier around, and that ValueNotifier is only used in text_selection.dart. 

(This PR also has some selection fixes when in read only mode)

I agree needs to be tested.  I have an internal team of 8, 1.2 million app users creating just over 10K notes a day. They have been running the 10.8.5 version that had my previous magnifier fixes.  Prior to the fixes, we received several support tickets about the magnifier bugs, since the fixes no reports. 

After we finish internal testing, we will do a rollout to our users and the feature will be used there.  We have finished updating to flutter_quill 11.x and require the selection and magnification changes.

lib/src/editor/raw_editor/raw_editor.dart
```diff
@@ -11,6 +11,7 @@ class QuillRawEditor extends StatefulWidget {
   QuillRawEditor({
     required this.config,
     required this.controller,
+    this.dragOffsetNotifier,
```
EchoEllet, line 14: I think this needs documentation, it's not clear this code is related to the magnifier feature.

lib/src/editor/widgets/text/text_selection.dart
```diff
@@ -984,7 +1055,42 @@ class _EditorTextSelectionGestureDetectorState
...
+                    ),
+                    size: const Size(100, 45),
+                    focalPointOffset: const Offset(5, 55),
+                    magnificationScale: 1.3,
+                  ),
+                ),
+              );
+            }),
+        ],
+      ),
```
EchoEllet, line 1093: This code is hardcoded; is there a need for the users to customize it?

lib/src/editor/raw_editor/raw_editor.dart
mtallenca: I added some comments in raw_editor.dart. Is that sufficient?

lib/src/editor/widgets/text/text_selection.dart
mtallenca: Good point - however changing the translate numbers, impacts focalPointOffset. Changing the size will impact both translate and focalPointOffset.  Making the entire builder a part of configuration would be appropriate.  In testing this looks good for light / dark modes on different font accessibility settings.  

This hard coded config also requires the scaffoldBackgroundColor to be reasonably set.

Should I update the code so that this is a builder passed in through configuration?

lib/src/editor/raw_editor/raw_editor.dart
EchoEllet: Thanks for the fix!

lib/src/editor/widgets/text/text_selection.dart
EchoEllet: Sounds ideal. I was about to suggest that we probably should allow overriding the implementation rather than a few properties (e.g., `magnificationScale`) to give them full flexibility. LGTM.

EchoEllet:
As a minor note, we should update the v11 migration guide to inform clients that the magnifier has been restored in version X after [it was removed](https://github.com/singerdmx/flutter-quill/blob/master/doc/migration/10_to_11.md#5-removal-of-the-magnifier-feature). This is not part of this PR since it should be done after releasing the new version.

EchoEllet:
Sorry for the delay, is the fix of selection in readOnly mode related to the PR #2488?

lib/src/editor/raw_editor/raw_editor_state.dart
```diff
@@ -853,8 +853,9 @@ class QuillRawEditorState extends EditorState
       });
     }
 
+    controller.addListener(_didChangeTextEditingValueListener);
```
EchoEllet, line 856: It's looks like this PR [removes the lines added by #2488](https://github.com/singerdmx/flutter-quill/pull/2488/files#diff-5f87ac57652e1ab8fa6a4062411e9cd842ab7a8163fe0b74a1f82fa7eb8f494cL817-L818) but keet some of the other changes, does this PR handle any related issues or conflicts with that PR? If no, then we need to revert that PR first, then merge this but revert the selection fix changes.

mtallenca:
Read only mode needs the listener for selection changes.  Only changes I see in my branch is  

controller.addListener(_didChangeTextEditingValueListener); and     controller.removeListener(_didChangeTextEditingValueListener);

are called in readonly mode.

#2488 moves the handle focus change listener inside if (!widget.config.readOnly) {}  - Could that have been the fix for focus changes bug?

mtallenca:
FYI - we've been internal testing with 20+ daily users for a couple of weeks with no issues.

EchoEllet:
Thanks for your contribution.

md-rifatkhan:
The default Android apps magnifier stays fixed to the text selection line, but this one moves freely with the finger's position.
and if the magnifier is above of the editor body, then text doesn't show anymore in magnifier.

Issue #2553, opened 2025-04-29
Issue with Grid Line Alignment in Flutter-Quill Editor


shanshen6:
### Have you checked for an existing issue?

- [x] I have searched the [existing issues](https://github.com/singerdmx/flutter-quill/issues)

### Flutter Quill Version

Flutter-Quill Version: 11.2.0 (or “flutter_quill: ^11.0.0-dev.4”)

### Steps to Reproduce

Dear Flutter-Quill Developers,

I hope this message finds you well. I am reaching out to report an issue I encountered while attempting to draw grid lines in the Flutter-Quill editor.

I noticed that the alignment of each line within the editor is not accurate, which prevents me from correctly drawing the grid lines. As a result, the grid lines are misaligned, leading to an unsatisfactory layout. I have set up custom styles for the editor, but despite my efforts, the grid lines do not align as expected.

Here are the details of my environment:
	•	Flutter-Quill Version: 11.2.0 (or “flutter_quill: ^11.0.0-dev.4”)
	•	Flutter Version:
Flutter 3.27.1 • channel stable
Framework • revision 17025dd882 (4 months ago)
Engine • revision cb4b5fff73
Tools • Dart 3.6.0 • DevTools 2.40.2

	Custom Styles: I am using a custom style setup for the editor, but the issue persists.

Could you please advise on how to resolve this issue or provide any guidance on how to achieve accurate line alignment when drawing grid lines in the editor?

I would greatly appreciate any insights or suggestions.

Thank you for your time and support.

this is my code: 

  @override
  Widget build(BuildContext context) {
    final style =
        YwDefaultStyles.getYwInstance(context, widget.controller.ywEditorConfig.editorType);
    widget.controller.ywEditorConfig.defaultStyles = style;
    return QuillEditor(
      focusNode: _editorFocusNode,
      scrollController: _editorScrollController,
      controller: _controller,
      config: QuillEditorConfig(
        editorKey: _controller.editorKey,
        scrollable: widget.controller.ywEditorConfig.editorType != EditorType.chapter,
        expands: true,
        showCursor: !widget.controller.ywEditorConfig.readOnly,
        placeholder: widget.controller.ywEditorConfig.placeholder,
        contextMenuBuilder:
            YwQuillRawEditorConfig.getContextMenuBuilder(widget.controller.ywEditorConfig),
        customStyleBuilder: (Attribute<dynamic> attribute) {
          return CustomAttrRender.render(attribute, context);
        },
[...]

PR #2555, opened 2025-05-01
Add Fields to Control Uniform Line Height for Span in flutter-quill

wangjinshan:
This pull request introduces two additional fields to control the uniform line height for the RichText widget in the flutter-quill package. The goal is to enhance the text rendering by enabling a fixed line height using the StrutStyle properties, which can be controlled via forceStrutHeight and leadingOffset. This update ensures that the line height is consistent across multiple TextSpan widgets, providing better visual alignment when rendering text.

Modifications:
	1.	Added the strutStyle: A StrutStyle is created from the textSpan.style (or a default TextStyle) to apply a fixed line height.
	•	forceStrutHeight: Ensures the strut height is enforced, even if it would normally be overridden.
	•	leading: This controls the leading offset (the space between lines of text).
	2.	RichText Widget Update: The strutStyle is applied to the RichText widget to manage text rendering with the specified line height and spacing.

Code Changes:

in commit

Rationale:
	•	Control over Line Height: The primary reason for this change is to allow better control over the line height for text spans in flutter-quill. By adding these fields, we ensure that the line height is consistent and can be adjusted per use case.
	•	Improved Text Layout: This change makes it easier to handle cases where consistent vertical spacing is required between lines of text, especially for rich text editing use cases.

Potential Impact:

This modification is non-breaking, as it only introduces additional properties that can be customized. It ensures better control over text appearance without altering the core functionality of the RichText widget.

Suggested Reviewers:
	•	@flutter-quill team
	•	@maintainers

### DIFF
--- a/lib/src/editor/widgets/default_styles.dart
+++ b/lib/src/editor/widgets/default_styles.dart
@@ -248,6 +248,7 @@ class DefaultStyles {
     this.sizeHuge,
     this.palette,
     this.forceStrutHeight,
+    this.leadingOffset,
   });
 
   final DefaultTextBlockStyle? h1;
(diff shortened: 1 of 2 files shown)


wangjinshan:
see 
https://github.com/singerdmx/flutter-quill/issues/2553

wangjinshan:
image

PR #2640, opened 2025-08-15
feat(iOS): use system context menu in iOS(for secure paste)

kexuejin:
Based on the PR template you provided, here's the Pull Request content prepared for your system context menu feature modification:

## Description
This PR modifies the Flutter Quill editor's context menu building logic to prioritize QuillSystemContextMenu when config.useSystemContextMenuItems is set to true . This change enables iOS secure paste functionality and improves user experience with native system context menus.

New behavior: When useSystemContextMenuItems is true and the platform supports it, the editor uses QuillSystemContextMenu.quillEditor() to display native system menus; if not supported, it falls back to the original menu building logic.

Motivation: To fully utilize iOS system-provided secure paste functionality and native user experience.

## Related Issues
- Related to iOS secure paste functionality improvements
## Type of Change
- ✨ Feature: New functionality without breaking existing features.
- 🛠️ Bug fix: Resolves an issue without altering current behavior.
- 🧹 Refactor: Code reorganization, no behavior change.
- ❌ Breaking: Alters existing functionality and requires updates.
- 🧪 Tests: New or modified tests
- 📝 Documentation: Updates or additions to documentation.
- 🗑️ Chore: Routine tasks, or maintenance.
- ✅ Build configuration change: Build/configuration changes.

### DIFF
--- a/lib/src/editor/editor.dart
+++ b/lib/src/editor/editor.dart
@@ -18,6 +18,7 @@ import '../document/nodes/leaf.dart';
 import 'config/editor_config.dart';
 import 'embed/embed_editor_builder.dart';
 import 'raw_editor/config/raw_editor_config.dart';
+import 'raw_editor/quill_system_context_menu.dart';
 import 'raw_editor/raw_editor.dart';
 import 'widgets/box.dart';
 import 'widgets/cursor.dart';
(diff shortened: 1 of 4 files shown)


kexuejin:
Since Flutter's SystemContextMenu only supports text in the clipboard, if there is an image in the clipboard, SystemContextMenu will not provide the Paste Menu.

lib/src/editor/editor.dart
```diff
@@ -281,8 +282,19 @@ class QuillEditorState extends State<QuillEditor>
...
+                      return QuillSystemContextMenu.quillEditor(
+                        quillEditorState: state,
+                      );
+                    }
+                    return (config.contextMenuBuilder ??
+                            QuillRawEditorConfig.defaultContextMenuBuilder)(
+                        context, state);
+                  }
+                : (config.contextMenuBuilder ??
+                    QuillRawEditorConfig.defaultContextMenuBuilder))
```
EchoEllet, line 297: This is a lot of nested code; I would prefer to clean it up, but also, the current code is not tested, so it's hard to guarantee that we won't introduce regressions.

I do get that this is a fork for your apps, but the way we should address this is by first testing the current code and verifying the tests are useful (since we don't follow TDD), and then adding the new feature with its tests.

senri1101:
I wanted to follow up on the status of this PR.
We believe this change is critical for enhancing the UX, so we are very keen to see it merged.
I am happy to assist with any remaining tasks if needed.

PR #2736, opened 2026-06-02
fix: make _selectedRects growable to avoid 'Cannot add to a fixed-length list'

Mounix99:
## Description

`RenderEditableTextLine.paint` appends a `TextBox` to `_selectedRects` when a selection contains an **empty line**:

https://github.com/singerdmx/flutter-quill/blob/master/lib/src/editor/widgets/text/text_line.dart#L1436

`_selectedRects` is assigned from `RenderBox.getBoxesForSelection(...)`, which can return a **fixed-length / unmodifiable** list. The subsequent `.add(...)` then throws:

```
UnsupportedError: Unsupported operation: Cannot add to a fixed-length list
  at FixedLengthListMixin.add (list.dart)
  at RenderEditableTextLine.paint (text_line.dart:1436)
```

This is a `thrown during paint()` crash, so it takes down the frame. We observed it in production (Flutter 3.41.2 / Dart 3.11, both Android and iOS) on a screen with a Quill editor. Repro: place the caret selection so that it spans an empty line within the document.

## Fix

Convert the result of `getBoxesForSelection(...)` to a growable list with `.toList()` before any `.add(...)`. One-line change; no behavioral difference other than the list now being mutable as the code already assumes.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed `UnsupportedError: Cannot add to a fixed-length list` thrown during `paint()` in `RenderEditableTextLine` when a selection contains an empty line. `_selectedRects` is now a growable list.
+
 ## [11.5.1] - 2026-05-20
 
 ### Added
--- a/lib/src/editor/widgets/text/text_line.dart
+++ b/lib/src/editor/widgets/text/text_line.dart
@@ -1419,9 +1419,14 @@ class RenderEditableTextLine extends RenderEditableBox {
           line.documentOffset <= textSelection.end &&
           textSelection.start <= line.documentOffset + line.length - 1) {
         final local = localSelection(line, textSelection, false);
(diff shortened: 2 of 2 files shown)


Alspb:
`.toList()` fixes the crash, but not its cause. The small extra rectangle for empty lines is still added outside the `??=` check, so it's appended on every redraw. Previously that crashed, now it just grows silently. Could you also move that add inside the check, something like below?

```dart
final local = localSelection(line, textSelection, false);

if (_selectedRects == null) {
  final rects = [..._body!.getBoxesForSelection(local)];

  // Paint a small rect at the start of empty lines that are contained by the selection.
  if (line.isEmpty &&
      textSelection.baseOffset <= line.offset &&
      textSelection.extentOffset > line.offset) {
    final lineHeight = preferredLineHeight(
      TextPosition(offset: line.offset),
    );
    rects.add(
      TextBox.fromLTRBD(0, 0, 3, lineHeight, textDirection),
    );
  }

  _selectedRects = rects;
}
```

Alspb:
Then could you merge the current `master` into this branch, following the steps under "Updating a PR opened before this change" in #2744? Merging it directly conflicts on almost every line.

PR #2765, opened 2026-09-09
Migrate to the standalone material_ui / cupertino_ui packages

putnokiabel:
## Description

Migrates the packages off `package:flutter/material.dart` / `package:flutter/cupertino.dart` and onto the new standalone **`material_ui`** / **`cupertino_ui`** packages, following the Flutter breaking change:
https://docs.flutter.dev/release/breaking-changes/material-ui-and-cupertino-ui

### How it was done

Following the documented procedure:

```bash
dart fix --apply --code=migrate_design_widgets   # rewrite imports
flutter pub add material_ui cupertino_ui         # add the new deps (per package)
dart fix --apply --code=directives_ordering      # re-sort imports
```

Applied across all packages: `flutter_quill`, `flutter_quill_extensions`, `flutter_quill_test` and `example`.

### Changes
- **97 Dart files**: imports rewritten from `package:flutter/material.dart` → `package:material_ui/material_ui.dart` and `package:flutter/cupertino.dart` → `package:cupertino_ui/cupertino_ui.dart`.
- **Dependencies**: added `material_ui: ^1.2.0` (all packages) and `cupertino_ui: ^1.0.2` (packages that use Cupertino widgets).
- **`example/lib/main.dart`**: `hide GlobalMaterialLocalizations` on the `material_ui` import — it now also exports that symbol, which was ambiguous against `flutter_localizations`.
- **`flutter_quill_extensions/pubspec_overrides.yaml`**: path override so the package builds against the in-repo (migrated) `flutter_quill` rather than the published release (which still imports `package:flutter/material.dart`, breaking the `ThemeData.isCupertino` extension). The override is ignored by consumers and excluded from `pub publish`, so it only affects local/CI resolution during the transition.

### Verification
- `dart analyze` — **no issues** in all four packages (lib + tests).
- `dart format --set-exit-if-changed .` — clean.

### Note for maintainers
CI (`.github/workflows/checks.yml`) pins `flutter-version: 3.44.1`. The `material_ui` / `cupertino_ui` packages require a newer Flutter (verified locally on **3.47.2**), so the pinned CI version will need bumping alongside this migration.

### DIFF
--- a/example/integration_test/flutter_quill_integration_test.dart
+++ b/example/integration_test/flutter_quill_integration_test.dart
@@ -1,8 +1,8 @@
-import 'package:flutter/material.dart';
 import 'package:flutter_quill/flutter_quill.dart';
-import 'package:flutter_test/flutter_test.dart';
 import 'package:flutter_quill_test/flutter_quill_test.dart';
+import 'package:flutter_test/flutter_test.dart';
 import 'package:integration_test/integration_test.dart';
+import 'package:material_ui/material_ui.dart';
 
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
--- a/example/lib/custom_toolbar.dart
+++ b/example/lib/custom_toolbar.dart
@@ -1,6 +1,6 @@
-import 'package:flutter/material.dart';
 import 'package:flutter_quill/flutter_quill.dart';
 import 'package:flutter_quill_extensions/flutter_quill_extensions.dart';
+import 'package:material_ui/material_ui.dart';
 
 /// Custom toolbar that uses the buttons of [`flutter_quill`](https://pub.dev/packages/flutter_quill).
 ///
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -2,11 +2,11 @@ import 'dart:convert';
 import 'dart:io' as io show Directory, File;
 
 import 'package:flutter/foundation.dart';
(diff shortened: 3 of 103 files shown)


EchoEllet:
Thank you for the PR. The migration documents suggest that migrating to the standalone `material_ui` and `cupertino_ui` packages introduces a breaking change for consumers who have not yet updated their dependencies: https://docs.flutter.dev/release/breaking-changes/material-ui-and-cupertino-ui#guidance-for-package-authors

So it might be more appropriate to release this as part of v12.0.0, and while we do it, we can introduce other breaking changes (including removing deprecated classes and methods).

putnokiabel:
> Thank you for the PR. The migration documents suggest that migrating to the standalone `material_ui` and `cupertino_ui` packages introduces a breaking change for consumers who have not yet updated their dependencies: https://docs.flutter.dev/release/breaking-changes/material-ui-and-cupertino-ui#guidance-for-package-authors
> 
> So it might be more appropriate to release this as part of v12.0.0, and while we do it, we can introduce other breaking changes (including removing deprecated classes and methods).

It's indeed a breaking change, it'd be great to get it in a major release soon! 🙏 
Most other major UI packages (that we are using) have upgraded (using a major release), users who are not ready for the migration yet can just stay on the previous major release (though the sdk built-in material and cupertino libraries will be deprecated in the next Flutter release so everyone will have to migrate soon).

